PR #2, opened 2015-07-02
AddStatement 1st version & stuff

wiztigers:
Hi Laurent,

Here are some changes that could use some revewing and pushing on your master branch :)

Regards


wiztigers:
6a8d49a should be refactored soon
- when identifiers and literals will be done
- to allow detection of 2nd operand-should-not-be-literal-because-there-s-no-GIVING error and such

laurentprudhon:
Thank you for your contribution, but I can't merge all the commits right now.
Please see the comments I attached to each one of the 4 commits, and to some significant source code lines.

laurentprudhon:
Recommended video :
[Using Git in Visual Studio 2015](https://channel9.msdn.com/Events/Build/2015/3-746)

wiztigers:
AddStatement is not yet refactored into Expression class, but I made some progress in making grammar rules more explicit as long as unit testing this feature.

Besides, this made me wondering about some grammar shenanigans I encountered while digging into .g4 files. I'm a bit unsure about some intents and inner workings of ANTLR.

In short: the goal of this commit is not to close the request, but to show progress.

wiztigers:
That's just a first proposal for class Expression and children.
There must be a better way to represent identifiers and litterals. I'll study this next week as long as the rules for these.

See you next week ! ᶘ ᵒᴥᵒᶅ

wiztigers:
The 3 ADD formats are now implemented, and I added more unit testing (compare the built token tree vs Reversed Polish Notation).
I now have to figure out how to manage syntax errors. However, I feel this pull request is mergeable as is, isn't it ?

laurentprudhon:
There are still TODOs, notably :
- handle ZERO figurative constant
- replace literal by numericLiteral if it makes sense
- full parsing of identifier rule
- correct expressions trees

But I feel this is explicit enough in the code and we can merge this pull request.

wiztigers:
Thanks for the merge !

Good point!
I've created the 3 first issues in my repo https://github.com/wiztigers/TypeCobol/issues
However I'm not sure about the meaning of the 4th one.

PR #5, opened 2015-07-09
ZERO/ZEROS/ZEROES

wiztigers:
Seen in [ef784cb] (https://github.com/wiztigers/TypeCobol/commit/ef784cbb11d0026bb5a091df689aa78bd00181b0)
Traced in https://github.com/wiztigers/TypeCobol/issues/5

Manage ZERO/ZEROS/ZEROES Tokens as (long) SyntaxNumber for ADD Statements.

### DIFF
--- a/TypeCobol.Test/Compiler/CodeElements/ArithmeticStatementTester.cs
+++ b/TypeCobol.Test/Compiler/CodeElements/ArithmeticStatementTester.cs
@@ -27,7 +27,7 @@ namespace TypeCobol.Test.Compiler.CodeElements
                 StringBuilder s = new StringBuilder();
                 Dump(s, add);
                 string dump = s.ToString();
-                if (c >= rpn.Length) errors.AppendFormat("RPN number {} not provided.", c);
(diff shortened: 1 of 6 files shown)


laurentprudhon:
ZERO(ES)(S) will be interpreted differently in different grammar rules.
Here I propose to simply create a SyntaxInteger from the token.
No need to create a new class, but you can adapt the SyntaxInteger Value property getter like this :

else if(Token.TokenType == TokenType.ZERO || Token.TokenType == TokenType.ZEROES || Token.TokenType == TokenType.ZEROS)
{
   return 0;
}

We don't try to convert between data types at the parser level : if we find a integer literal and then a floating point literal in an expression, we parse them without conversion.

wiztigers:
3e0c06f3bd97536fe47661060bb9a020d170619b okay dokey !
I'd think this begins to be a tad too costly for a getter, but ... heck, we'll see. No premature optimization !

wiztigers:
yay! thanks!

Issue #50, opened 2015-08-07
Identifiers beginning with digits


wiztigers:
**BLOCKS ISSUE #23**

Seen in real-life code: `PERFORM 0000-INIT-STD`, as the first line of a PROCEDURE DIVISION.
Are identifier names allowed to begin with digits ?
If yes, in which context(s) ? Always ?

osmedile:
You can write identifiers and paragraphs that start with digits.

Page 9 of Cobol V5.1 reference:
> A COBOL word is a character-string that forms a user-defined word, a
> system-name, or a reserved word. The maximum size of a COBOL user-defined
> word is 30 bytes. The number of characters that can be specified depends on the
> code page indicated by the compile-time locale.
> Except for arithmetic operators and relation characters, each character of a COBOL
> word is selected from the following set:
> * Latin uppercase letters A through Z
> * Latin lowercase letters a through z
> * digits 0 through 9
> * - (hyphen)
> * _ (underscore)
>
> The hyphen cannot appear as the first or last character in such words. The
> underscore cannot appear as the first character in such words. Most user-defined
> words (all except section-names, paragraph-names, priority-numbers, and
> level-numbers) must contain at least one alphabetic character. Priority numbers and
> level numbers need not be unique; a given specification of a priority-number or
> level-number can be identical to any other priority-number or level-number.
> In COBOL words (but not in the content of alphanumeric, DBCS, and national
> literals), each lowercase single-byte alphabetic letter is considered to be equivalent
> to its corresponding single-byte uppercase alphabetic letter.*

smedilol:
As stated by Cobol reference:
The mantissa of a floating point always contains a decimal point (char '.' by default or char ',' if DECIMAL-POINT IS COMMA is specified).
So there's no confusion between a floating-point and a variable name.


Picture of a floating-point use almot the same syntax, except that char 'V' can replace the decimal point. The following picture clause are equivalent:
```Cobol    
       01 floatingPoint  pic -9v9(9)E-99.
       01 floatingPoint2 pic -9.9(9)E-99.
```
I can then write the following paragraph:
```Cobol
       testFloatingPoint.
           compute floatingPoint = 2.3E-13 + -3.444E-1.
```

laurentprudhon:
In the current version, the Scanner has been implemented with an important goal in mind : to help the developers correct their spelling mistakes by giving them meaningful error messages.
[...]

PR #55, opened 2015-08-10
Fix for the issue #50 : support for user defined words starting with a digit

laurentprudhon:
First attempt to implement the ambiguous spec for user defined words starting with a digit :

PROBLEMS :   
123 is a valid user defined word (for section & paragraph names) AND a valid numeric literal.  
123E-4 is a valid user defined word (for any type of name) AND a valid numeric literal (floating point format).  
123-456 is a valid user defined word (for section & paragraph names), AND it could be interpreted as two numeric literals. NB: it is NOT a valid subtraction (minus must be preceded and followedf by space) but we would like to display a nice error message informing the user that these spaces are mandatory, because a subtraction was most likely intended in this case.  
000010-000050 should be interpreted as a range of numbers (indicated by separating the two bounding numbers of the range by a hyphen) in sequence-number-fields of compiler directive statements (ex: DELETE).


osmedile:
Pull request manually merged on Aug 11, 2015

PR #115, opened 2015-10-19
Refactoring for a safe multithread incremental compilation process

laurentprudhon:
There is stil work to do : first we will have to test that everything works as expected in incremental mode (no new unit tests are available yet), then we need to implement the incremental code elements parsing step, and finally we must finalize and test the implementation in TypeCobolStudio.
But a lot of work has been done in a separate branch, all existing unit tests are passing, so I think this is the right time to merge these changes in the master branch please.
I will synchronize my fork with the master branch as soon as possible, and switch back to small and frequent pull requests to make progress on the three points listed above.


wiztigers:
This breaks sample testing using `GrammarTest`, but ... -hey, it's okay, I'll fix it.
Thanks for all this work, @laurentprudhon  ( ＾◡＾)っ❤

Issue #124, opened 2015-11-05
Multi-pass parsing


wiztigers:
I'm attempting to write a test asserting that you can parse a document, then modify it and reparse it a given number of times, and that the code elements are always consistent with the modifications.
I saw that `TestCollection.CheckPipeline` tests are totaly empty, so I tried to roll my own in d9f4fafda5639c70a65ff4deaf728a4718ca7f9d.

I'm trying to wrap my head around how this should be done, and it seems either I went the wrong direction, or there is something off with the current implementation.

Problem is: wether I update or remove a given line, or even clear the whole document, the resulting code elements are not as expected.
For example, clearing the following document ...
```cobol
IDENTIFICATION DIVISION.
  PROGRAM-ID. Empty.
```
... gives a parsing result corresponding to this document:
```cobol
IDENTIFICATION DIVISION.
IDENTIFICATION DIVISION.
```

How I attempt to do the test can be seen in d9f4fafda5639c70a65ff4deaf728a4718ca7f9d, but to summarize:
```csharp
...
// ... document is Sample/Programs/Empty.cbl ...
var compiler = new FileCompiler(null, filename, project.SourceFileProvider, 
                                project, format.ColumnsLayout, options, false);
compiler.CompileOnce();
// ... compare CodeElements with ResultFiles/Programs/Empty.0.txt ...

// ... create an event TextChangeType.LineInserted, line=2, and text "END PROGRAM Empty."
compiler.CompilationResultsForProgram.UpdateTextLines(e);
compiler.CompileOnce();
// ... compare CodeElements with ResultFiles/Programs/Empty.1.txt ...

// ... create an event TextChangeType.LineUpdated, line=1, and text "PROGRAM-ID. Emptier."
compiler.CompilationResultsForProgram.UpdateTextLines(e);
compiler.CompileOnce();
// ... compare CodeElements with ResultFiles/Programs/Empty.2.txt ...

// ... create an event TextChangeType.DocumentCleared
compiler.CompilationResultsForProgram.UpdateTextLines(e);
compiler.CompileOnce();
// ... compare CodeElements with ResultFiles/Programs/Empty.3.txt
```

wiztigers:
@laurentprudhon @prudholu  , any ideas ?
Rather than using `UpdateTextLines`, shouldn't I provide a custom implementation of `TextDocument`, mirroring the client's line modifications ?
But then, how can I poke the parser so it re-parses only what has changed ?

Issue #125, opened 2015-11-05
ProgramClass phase: (?) ANTLR receives empty stream


wiztigers:
Okay, the analysis in the issue title may be totally wrong, but it seems to be what happens.

First parsing phase (with CodeElements grammar, converting text Tokens into a CodeElements list) happens nicely as always.
Second parsing phase (with ProgramClass grammar, convertin CodeElements into more higher level CodeElements) cannot seem to happen.

Whatever the input file is, ANTLR complains with the following error:
```
[1,0]<27,Error,Syntax>Syntax error : no viable alternative at input '<EOF>' 
(RuleStack=cobolCompilationUnit, OffendingSymbol=[1,0:]<EndOfFile> on line 0)
```
... even tough `CodeElementsLinesTokenSource` seems to provide the appropriate tokens.

Even if I change ProgramClass grammar starting rule `cobolCompilationUnit` to be something like ...
```
cobolCompilationUnit: ContinueStatement EOF;
```
... and I only provide the word `CONTINUE`, ANTLR complains with the following error:
```
[1,0]<27,Error,Syntax>Syntax error : missing ContinueStatement at '' (RuleStack=cobolCompilationUnit, OffendingSymbol=[1,0:]<EndOfFile> on line 0)
```

This leads me to think that there may be a problem of token stream resetting between the two parsing phases (like [described here](http://stackoverflow.com/questions/7335847/all-antlr-grammars-produce-error-no-viable-alternative-at-input-eof)) or something. According to @laurentprudhon , there can be ANTLR caching management issues, too.

PR #129, opened 2015-11-10
Fixes issues #124 and #125 for incremental parsing

laurentprudhon:
The test TestPaser.Check_BeforeAfterInsertion() now runs correctly.
Main fixes for issue 124 :
- copy/paste error : choosed the wrong list of lines as a starting point when initializing the CodeElements parsing step
- the incremental parsing of CodeElements is not implement yet, so we must reset all CodeElements each time before starting to parse again

Main fixes for issue 125 : 
- bad channel number returned by CodeElement when used as IToken
- CodeElementType values were starting at 0 while the enumeration of tokens in an Antlr grammar is always transformed in a list of values starting at 1


wiztigers:
Works like a charm.
Thanks, mate!  ٩(•̤̀ᵕ•̤́๑)

Issue #149, opened 2015-12-01
Compiler directing statements


wiztigers:
Compiler directing statements seem processed by the preprocessor (in `CompilerDirectiveBuilder`).
However, whereas these statements don't impair CodeElements grammar parsing at all, they raise syntax errors in ProgramClass grammar.

wiztigers:
Doesn't work:
```cobol
CONTINUE
SKIP3.
```

Work:
```cobol
SKIP3.
```

```cobol
CONTINUE.
SKIP3.
```

```cobol
CONTINUE
SKIP3
CONTINUE.
```

Replacing `CONTINUE` by `PERFORM procedurename` yields the same results.

wiztigers:
@laurentprudhon , any ideas about why some compiler directing statements like `SKIP` make the ProgramClass grammar complain, whereas CodeElements grammar doesn't scoff at them ?
What happens can be tested with what I committed in 56c82f191919a6ff94bc9eb5f0e1c39a8c5b68f5 : the test should pass but does not.

laurentprudhon:
Ok, I will fix this this weekend : my guess is that I didn't correctly filter the token types in the token source for the last phase of parsing

laurentprudhon:
Your test program is invalid, the compiler behavior is correct in this case.

You write for example :
PROCEDURE DIVISION.
SKIP-01. => paragraphName
  CONTINUE => continueStatement
  SKIP3. => compiler directing statement, this whole line is invisible to the ProgramClass paser
<< here a separator period is missing to end the sentence >>
SKIP-02. => paragraphName 
<< you can not start a new paragraph here, because the last sentence was not finished >>

Reference text for SKIP statement p545 : 
SKIP1, SKIP2, or SKIP3 can be written anywhere in either Area A or Area B, and can be terminated with a separator period. It must be the only statement on the line.

Reference text for paragraph syntax p253 : 
Paragraph 
A paragraph-name followed by a separator period, optionally followed by one or more sentences.
!! Paragraphs must be preceded by a period because paragraphs always follow either the IDENTIFICATION DIVISION header, a section, or another paragraph, all of which must end with a period. !!
Sentence 
One or more statements terminated by a separator period.

wiztigers:
Thanks to you @laurentprudhon, I have a better understanding now.

However, I'm not sure the test is invalid : it mimicks real production code.

[...]

PR #158, opened 2015-12-28
Incremental parsing for CodeElements  / Fix for bug #149 : EJECT, SKIP, TITLE Period

laurentprudhon:
Fix for bug #149 : EJECT, SKIP and TITLE compiler directives should consume the following period separator only if it is on the same line.

Update Antlr version to 4.5-alpha003.

Add missing reference PDF documents from IBM.


laurentprudhon:
Implementation of incremental parsing in CodeElementsParserStep.
NB : CodeElementsLine lifecyle is still not completely OK for a multithread environment (CanStillBeUpdatedBy test always returns true) + we need to write a lots of tests, but the main algorithm is there.

Performance statistics for all compilation steps + performance test of the complete compilation pipeline in after and incremental update to a large program (20000 lines).
NB : the BigBatch.cbl file was not saved in the public Github because it is a real production file (IPPBATCH).

PR #165, opened 2016-01-22
Implementation of the legacy non-standard "*REMARKS COPY=()" EI syntax 

laurentprudhon:
*REMARKS is a legacy Euro Info proprietary syntax, roughly equivalent to COPY REPLACING.

All the non-standard code sections have been wrapped inside conditional compilation directives : 

#if EUROINFO_LEGACY_REPLACING_SYNTAX

Just undefine this compilation symbol in the project build properties, and all traces of this non standard syntax will instantly disappear.

### DIFF
--- a/TypeCobol.Test/Compiler/Preprocessor/DirectiveResultFiles/Copy.txt
+++ b/TypeCobol.Test/Compiler/Preprocessor/DirectiveResultFiles/Copy.txt
@@ -118,7 +118,7 @@
 -- Line 33 --
 
 -- Line 34 --
-[8,72:REMARKS. COPY=(YIDCHB1                                           ]<CommentLine>
+*** DIRECTIVE REMARKS YIDCHB1 ([8,72:REMARKS. COPY=(YIDCHB1                                           ]<CommentLine>) ***
 
 -- Line 35 --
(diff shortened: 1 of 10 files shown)


wiztigers:
Thanks for your work.

Before merging this PR however, I'm concerned about the best way to integrate such "proprietary" features in our parser.
Good point for wrapping this with a specific flag, but -please see my mail for details.

wiztigers:
Merged as 93df167dbd157c061572b6784270e606003ba328.

wiztigers:
The current implementation doesn't support all formats of `*REMARKS`.
This is an acceptable syntax:
```cobol
*REMARKS.       COPY=(
*                     COPY1   COPY2
*                     COPY3
                      COPY4
*                     ).
```
So, we just have to catch comment or _non-comment_ lines until the next dot, isn't it?
Well, not exactly, because this is an acceptable syntax, too:
```cobol
*REMARKS.
*         COPY=(COPY1, COPY2, COPY3, COPY4, COPY5)
*         COPY=(COPY6, COPY7)
*         COPY=COPY8
```

wiztigers:
Okay, now it seems good.
However, if `*REMARKS` allow empty lines outside of a `COPY=( .. )` we'll be screwed.

PR #178, opened 2016-02-02
Optimization round 1 : total compilation time reduced by 66%

laurentprudhon:
…and memory consumption reduced by 30%.

TO DO : due to changes in the structure of the logical and arithmetic expressions in the grammar, a lot of lines were commented in the related CodeElementBuilders => they must now be adapated to the optimized structure, see Chapter 14 "Removing Direct Left Recursion" p247 in the Antlr 4 reference book.


wiztigers:
I fixed some stuff so unit tests pass and vastly reimplemented arithmetic expressions.

There are still __19__ files in error instead of __1__ when you `CheckGrammarCorrectness` on our production sample, but I merged the PR nevertheless: 2'45'' instead of 4'26'' is just too yummy ◕◡◕

Thanks for the directions, @laurentprudhon !

PR #182, opened 2016-02-04
Implementation of a Visual Studio Code language server for TypeCobol 

laurentprudhon:
The same language server could be used for the Eclipse editor, because it uses an simple, open, and robust json protocol to optimize the communication between editor and compiler.
This is a proof of concept, more work will be needed in the following weeks to make it really useful in Visual Studio Code, but the technical foundations are there.


wiztigers:
I merged pull request today.
The little configuration files fix 2fe30cd5fc0562d424ca68cdeeda15a2956e4550 was necessary ; I don't like increasing required tooling versions without valid reason.

As a027643b019d948cdcae27541f2e81c89aa87afd was part of both pull requests #182(this one) and #178, both were merged at once.
No merge commit this time, as I wanted to test the `--ff` flag.

PR #196, opened 2016-03-13
Symbol definitions and references - Language model and stats

laurentprudhon:
Exhaustive study of symbol definitions and references in Cobol grammar : see TypeCobol/Documentation/Studies/Symbols and User defined words.xlsx.

Antlr grammar cleanup to enable a more precise symbol management : 
- split symbolDefinition and symbolReference in separate rules
- merge rules together when the CodeElements parser can not distinguish between several candidate symbol types

CodeElementsBuilder now uses the newly defined grammar rules to create precise SymbolInformation for each symbol Token, information stored in a new Dictionary called SymbolInformationForTokens in CodeElements class.

Updates to the existing xxxBuilder classes to support the new grammar rules names without breaking anything upstream. Updates to the test result files to take into account the new rules names.

New tool to compute stats and generate a language model - using all the new and reliable SymbolInformation (will be useful for auto completion in IDE) : 
TypeCobol/Tools/CommandLine/TypeCobolStats.cs


laurentprudhon:
WARNING : this a preview of a work in progress, to avoid duplicate work on your side.

While I think that could safely merge this pull request now, you must be aware that I still need a few days of work. Please look at it to understand what is going on, and wait a later comment giving the green light when the work is done.

Current status : 

* Symbols DONE
ProgramName
SectionName
ParagraphName
ClassName
ExternalClassName
MethodName
MnemonicForEnvironmentName
MnemonicForUPSISwitchName
ConditionForUPSISwitchName
SymbolicCharacter
AlphabetName
CharacterClassName
DataName
ConditionName
XmlSchemaName

* Symbols TO DO
IndexName
FileName

* External names DONE
EnvironmentName
UPSISwitchName

* External names TO DO
TextName
LibraryName 
AssignmentName
FunctionName 
ExecTranslatorName

wiztigers:
@laurentprudhon, your pull request is now integrated. Thanks for your work.

As our `master` has continued advancing since your branch `ei-cobol-stats2` was created, there were conflicts preventing an easy merge.
These conflicts were mainly due to the new .g4 file for TypeCobol created for #201, which overrides all name qualification rules. It was rather easy before your PR, as there was only one rule for data and another for conditions ; but now that we have, like, seven different rules, which cannot be handled the same way in C#, creating the need for MANY factory methods for data, conditions, dataorconditionorfile and so on ...
In short, it was a PITA to update all these methods that are the-same-but-not-quite due to strongly typed contexts objects and lack of efficient duck coding in C#.

Anyway, it's done, now.
As you can see however, there are some unit tests that could not be merged "as is" : these are `XML/GENERATE`, `MERGE.ko`, `SET` and `SORT.ko`. The line indexes of all diagnostics are the same, but some messages differ. This bothers me a tad, so feel free to contact me if it bothers you too, as it may be the sign of an hidden integration problem.

Issue #235, opened 2016-06-27
COPY Procedure may produce parsing error


smedilol:
** DUPLICATE OF #312 **
In certain cases, COPY parsing can produce errors:
Giving this COPY:

``` cobol
move c to d
```

The following program parse just fine:

``` cobol
IDENTIFICATION DIVISION.
PROGRAM-ID. TypeCobol.
PROCEDURE DIVISION.
copy CPY
.

IDENTIFICATION DIVISION.
PROGRAM-ID. TypeCobol.
PROCEDURE DIVISION.
copy CPY.
.

IDENTIFICATION DIVISION.
PROGRAM-ID. TypeCobol.
PROCEDURE DIVISION.
move a to b
copy CPY.
.
```

But the following program ends in error:

``` cobol
IDENTIFICATION DIVISION.
PROGRAM-ID. TypeCobol.
PROCEDURE DIVISION.
move a to b
copy CPY
.
```

[1,21:<missing SentenceEnd>]<HexadecimalNationalLiteral>{<missing SentenceEnd>}

You can see tests in:
`TypeCobol.Test/NewTestsToFix/Parser/Samples/HighPriority/CopyProcedure*.cbl`

_edited to add more info on the tests_

smedilol:
When entering ProgramClassParserStep.ParseProgramOrClass method, codeElementsLines List contains one CodeElement for the copy statement. But the CodeElements for the statements inside the copy are absent.

smedilol:
@laurentprudhon  can you look after this issue?

laurentprudhon:
The following two lines in CodeElementsParserStep.ParseProcessedTokensLinesChanges() are incorrect :

> // Get the first line that was parsed
> CodeElementsLine codeElementsLine = ((CodeElementsLine)((Token)codeElementParseTree.Start).TokensLine);
> ...
> // Add code element to the list
> codeElementsLine.AddCodeElement(codeElement);

1. When the first Token of the current CodeElement (codeElementParseTree.Start) was found in an imported (COPY) file, codeElementsLine is a line of the imported (COPY) document, not a line of the mail (PROGRAM) document.
2. In this case, the codeElement object is added to a line of the COPY document, instead of a line of the PROGRAM document (this is the BUG !)
3. The next parsing step expects all CodeElements to be added to the main PROGRAM documents lines : it makes no sense to associate a CodeElement with a COPY document, because a CodeElement could perfectly span Tokens both inside and outside the COPY document.

[...]

PR #312, opened 2016-10-09
Fix: Copy directive without sentenceEnd are not parsed properly.

osmedile:
See #235 for others examples.

Following syntax is accepted by IBM Compiler V4 on Z/OS:

``` cobol
working-storage section.
COPY MyCopy
01 MyData pic X.
local storage section.
```

This PR need to be tested against a large real Cobol 85 samples to ensure there is no regression.

(this PR also include commits from my other PR (PR#311 on TypeCobolTeam/TypeCobol repo).


laurentprudhon:
@smedilol I think you should not fix this bug like this : 
- cf Cobol reference doc p531 : final periodSeparator is mandatory for COPY compiler statement
- if you write : "dataDescriptionEntry sentendEnd?" in the grammar, it is expanded as : "xxx periodSeparator periodSeparator?" : two period separators at the end of a code element can't be right
- this will only work if COPY is used anywhere else than aligned with a dataDescriptionEntry in the grammar (and it needs to work anywhere)
- after this pull request, writing a COPY with a final periodSeparator (as per Cobol reference) in the procedure division could lead to severe regressions, as inserting the period in the middle of a sentence can change the scope of an if condition for example

I would interpret the result of your test as follows : like for other compiler statements, you can terminate COPY either with a final periodSeparator, or with the end of line.
I already fixed a couple of similar bugs for other compiler statements like this :
ejectCompilerStatement: ({IsNextTokenOnTheSameLine()}? EJECT PeriodSeparator?) | (EJECT);

IsNextTokenOnTheSameLine is a property defined in TypeCobol.Compiler.AntlrUtils.LineAwareParser.

We need to find a similar solution for COPY, but is a little bit harder because COPY often are multiline statements. We need to do more tests with the IBM compiler to understand how it behaves regarding with end of lines and the usage of COPY in the data division AND the procedure division.

laurentprudhon:
@smedilol an additional question please : are the following programs working correctly with the IBM Cobol compiler ?

working-storage section.
COPY MyCopy1
local storage section.

working-storage section.
COPY 
MyCopy1
local storage section.

working-storage section.
COPY MyCopy1 REPLACING A BY B
local storage section.

working-storage section.
COPY MyCopy1 
REPLACING A BY B
local storage section.

working-storage section.
COPY 
MyCopy1
01 MyData pic X.
local storage section.

working-storage section.
01 MyData pic X.
COPY 
MyCopy1
local storage section.

working-storage section.
COPY MyCopy1
COPY MyCopy2
local storage section.

working-storage section.
COPY MyCopy1 COPY MyCopy2
local storage section.

working-storage section.
COPY MyCopy1 REPLACING A BY B COPY MyCopy2
local storage section.

working-storage section.
COPY MyCopy1
01 MyData pic X.
COPY MyCopy2
local storage section.

Thanks

smedilol:
I abandon this PR, we'll wait for the new preprocessor for Replace and COPY to fix this eventually.

Issue #288, opened 2016-09-21
Increase parser robustness


wiztigers:
On some input file which contain COBOL errors (for example simple things like forgotten dots, forgotten statement end keywords, and so on) we encounter harsh runtime exceptions like null pointer errrors or cast errors. This breaks the program, we are unable to continue parsing untill end of file, unable to dump any diagnostics and the user (which we can name for example @reydelpa ヾ(-‿-; )) is left clueless with no indication about what is happening.

How could we protect our program by making it able to restart on this kind of error ?
At best, we need something like protecting each `EnterXXX(XXXContext...)` method with its own `try ... catch ...`, create at this place an object like `CodeElementInError` or `NodeInError` which would contain the source code interval causing problem, and a diagnostic message containing the runtime exception.

However, wouldn't this systematic and generic `try ... catch ...` heavily impact performance ?
@laurentprudhon @osmedile @mayanje  @KOBISCAL , any opinions ?

smedilol:
According to:
http://stackoverflow.com/questions/52312/what-is-the-real-overhead-of-try-catch-in-c
http://stackoverflow.com/questions/1308432/do-try-catch-blocks-hurt-performance-when-exceptions-are-not-thrown
The real cost is to throw an exception. having try/catch in your code doesn't seems to have a real impact in performance.

As Antlr already throw an exception, I'd say that you can catch it in order to improve robustness of the parser.

wiztigers:
Code parsed by our parser is intrinsically wrong the majority of the time.
A lifetime wouldn't suffice to test _all_ that can be wrong in _all_ TypeCobol or COBOL programs. Altough our unit tests try to reach this goal, they won't cover _all_ edge cases until faaar in the future.
Until know, our `CodeElement`/`Node` factory methods can theorically break anytime because of a thrown runtime exception (null pointer access error, cast error and so on).

That's why, after discussing with @smedilol and @laurentprudhon , we deemed a `try ... catch ...` to ultimately be necessary, even if we work at increasing our behaviour consistency through efforts like the one discribed in #294.

wiztigers:
This [sample file](https://raw.githubusercontent.com/TypeCobolTeam/TypeCobol/cb8f870fb6e884a95492d5c6fd7ac94afa0f0439/TypeCobol.Test/NewTestsToFix/Parser/Samples/HighPriority/RuntimeExceptions.cbl) provided by @mayanje makes our Parser run in some (NPE) runtime errors.
[...]

PR #394, opened 2016-12-20
#288 Improve error recovery - increase parser robustness

laurentprudhon:
This pull request improves the parser ERROR RECOVERY strategy :

Documented the Antlr parsing and error recovery strategies in the following documents :
- Studies/Antlr 4 parsing algorithm.pdf
- Studies/Antlr 4 error recovery strategy.txt

New CodeElements parsing strategy :
- updated CodeElementsParserStep to parse all codeElements at once instead of one at a time : this gives Antlr the opportunity to implement better error recovery out of the box, because in this context it knows how to resynchronize with the start of the next code element
- updated ParseSection / currentParseSection management in support to the new CodeElements parsing strategy
- updated TokensLinesTokenStream to be able to limit the stream scope to a specific ParseSection : seek to the first token of the ParseSection, wait for the last token of the parse section, and return a replaced Token of type EOF instead
- updated CobolCodeElementBuilder.ExitCodeElement to compute the list of tokens consumed while matching the CodeElement : this is now done by a traversal of the Antlr tree rather than a synchronization with the underlying tokens iterator, with the advantage of being compatible with the new parsing strategy, and to include the missing tokens generated by the improved error recovery strategy

Bugs fixed in CodeElements incremental parsing :
- updated CodeElementsParserStep.CheckIfAdjacentLinesNeedRefresh  to fix a few bugs detected in the code elements incremental parsing algorithm

Improved error recovery strategy :
- updated CodeElementErrorStrategy and CompilerDirectiveErrorStrategy to be much less intrusive in Antlr native error recovery strategy
- the idea is that we only want to avoid Antlr "eating away" the token starting the next CodeElement when trying to resynchronize after an error
- the way to do this is simply to override the SingleTokenDeletion and the ConsumeUntil methods

Improved error reporting :
- created 5 new virtual methods ErrorMessageForXxx in CobolErrorStrategy : you just need to replace or overwrite the code in these five methods to customize the error messages produces by Antlr to your needs
- upgraded Nuget development dependency to to Antlr 4.5.3, necessary to benefit from the brand new contextSuperClass grammar option
- updated all 3 master CobolXxx grammar files with contextSuperClass=TypeCobol.Compiler.AntlrUtils.ParserRuleContextWithDiagnostics : all nodes of the Antlr parse tree will now be able to collect information about all problems encountered during the parsing of the associated rule
- updated ParserDiagnosticErrorListener to attach all collected error messages to the current parse tree node (derived from ParserRuleContextWithDiagnostics)
- updated CobolCodeElementBuilder.ExitCodeElement to compute the list of errors encountered while matching the CodeElement : this is now done by a traversal of the Antlr tree, with the advantage of being compatible with the new parsing strategy, and to prepare with a more precise decomposition of the syntax tree if necessary in a later version

Improved missing tokens generation :
- new properties added on TokensLinesIterators : CurrentLine and LastLine (used to improve the missing token factories)
- updated TokensLinesTokenSource : CobolTokenFactory.Create now returns a new Token-derive class called MissingToken, which does not store the text of the token in the underlying TokensLine, but provides correct values for the position of the virtual Token in the line (with a 0 length)
- updated CodeElementsLinesTokenSource : CobolTokenFactory.Create now instantiates a real CodeElement-derived class depending on the expected TokenType

Improved Scanner and token types for contextual tokens :
- completely refreshed the study in Studies/Code elements starting tokens.xlsx (see v2 tabs)
- refreshed the tokens types and families, code generated from the file Studies/CobolLexer.tokens.xlsx in TokenType.cs and TokenUtils.cs
- removed the state machine in MultilineScanState : replaced by a simple set of properties, reimplemented contextual scanning in a more simple and direct way
- removed all token types specifically created to distinguish code element starting tokens : now using token family for this
- created three new token types to avoid as much ambiguity as possible : LevelNumber, SectionParagraphName, QualifiedNameSeparator
- implemented all the mechanisms needed to transfer the scanner state, especially to scan a COPY in the exact context of its position in the middle of a line

Improved error messages :
- provided human readable names for all token types and token families
- replaced lists of expected token types with more than 10 elements with simplified lists of token families

Tests updates :
- updated RuleStackBuilder.GetRuleStack and DiagnosticSyntaxErrorListener.SyntaxError to hide new root rule "coboCodeElements", as this would have an impact on every single test already written 
- updated TestTokenSource to account for for the improved missing token factory
- added new test CheckParserRobustness to control the improvements of the new error recovery strategy when parsing multiple CodeElements [TO DO finalize this test]
- merged performStatement and performProcedureStatement int the grammar because they can't be distinguished at runtime by the parser (and updated CobolCodeElementBuilder and CobolStatementsBuilder accordingly)
- fixed all tests results according to the new error recovery strategy, which changes the error messages displayed and the list of code elements detected when a syntax error is encountered


smedilol:
After some tests, everything seems to run fine on our extended private sample. 
I notice a small performance gain.

PR #423, opened 2017-02-14
Optimize parser startup time

parandba:
"Optimize parser startup time" issue related code changes.


CLI/src/CLI.csproj
```diff
@@ -76,7 +76,13 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="*.cs" />
+    <Compile Include="CLI.cs" />
```
smedilol: VS try to modify csproj file, but we used to let wildcard *.cs everywhere.
For next Pull request, can you avoid to commit this kind of change, please.

CLI/src/Daemon.cs
```diff
@@ -50,188 +43,194 @@ class Config {
...
-			errmsg += "Try "+PROGNAME+" --help for usage information.";
-			Console.WriteLine(errmsg);
-			return code;
-		}
-
-	}
+            return 0;
+        }
+
+        private static Compiler.DocumentFormat CreateFormat(string encoding, ref Config config)
```
smedilol: The code of this method is duplicated, can you refactor this?

CLI/src/Serialization.cs
```diff
@@ -6,111 +6,230 @@
...
+            foreach (MsgPack item in msgpack.ForcePathObject("CopyFolders"))
+            {
+                CopyFolders.Add(item.ForcePathObject("CopyFolder").AsString);
+            }
+            config.CopyFolders = CopyFolders;
+
+            List<string> Copies = new List<string>();
+            foreach (MsgPack item in msgpack.ForcePathObject("Copies"))
+            {
+                Copies.Add(item.ForcePathObject("Copies").AsString);
```
smedilol: It seems there is bug here, as in Encode method, it was "copy"

CLI/src/Command.cs
```diff
@@ -1,83 +1,162 @@
...
+        }
+    }
+
+    /// <summary>
+    /// RunCommandLine class implementing AbstractCommand members.  
+    /// This is called when client passes value "68".
+    /// </summary>
+    internal class RunCommandLine : AbstractCommand
+    {
+        private StringSerializer Deserializer = new StringSerializer();
```
smedilol: Field is never used, can you remove it?

CLI/src/Serialization.cs
```diff
@@ -6,111 +6,230 @@
...
+                {
+                    var child = item.AddArrayChild();
+                    child.ForcePathObject("Copy").AsString = copy;
+                }
+            }
+        }
+        internal static Config Decode(MsgPack msgpack)
+        {
+            Config config = new Config();
+            config.ErrorFile = (string)msgpack.ForcePathObject("ErrorFile").AsString;
```
smedilol: Cast to string is useless here. Can you remove it?

CLI/src/Serialization.cs
```diff
@@ -6,111 +6,230 @@
...
+            config.skeletonPath = (string)msgpack.ForcePathObject("skeletonPath").AsString;
+            config.Codegen = ((string)msgpack.ForcePathObject("Codegen").AsString).Equals("True") ? true : false;
```
smedilol: End of expression: "? true : false" is useless here. Can you remove it?

CLI/src/Daemon.cs
```diff
@@ -50,188 +43,194 @@ class Config {
...
+                        {
+                            namedPipeClient.Connect(100);
+                        }
+                        catch (TimeoutException tEx)
+                        {
+                            System.Diagnostics.Process process = new System.Diagnostics.Process();
+                            System.Diagnostics.ProcessStartInfo startInfo = new System.Diagnostics.ProcessStartInfo();
+                            startInfo.WindowStyle = System.Diagnostics.ProcessWindowStyle.Hidden;
+                            startInfo.FileName = "cmd.exe";
+                            startInfo.Arguments = @"/c TypeCobol.CLI.exe";
```
smedilol: One thing I didn't thought about is that TypeCobol.CLI is not necessarily in the path.
So I see 2 possible solutions:
- Can you retrieve the current folder of the TypeCobol.CLI.exe and then use it to launch this command
- If it doesn't works, then the path to TypeCobol.CLI.exe need to be passed as an argument (with option -K)

Do you see another solution?

CLI/src/CLI.csproj
parandba: ok

CLI/src/Daemon.cs
parandba: Done.

CLI/src/Serialization.cs
parandba: Done.

CLI/src/Serialization.cs
parandba: Config.Codegen accepts bool value.  Hence, if "True" is returned as string, converting it to true as boolean, else false.

CLI/src/Serialization.cs
parandba: Done.

CLI/src/Command.cs
parandba: Done.

smedilol:
PR merged manually

PR #426, opened 2017-02-15
Diffplex implemented to compare strings in TestUtils.cs

parandba:
Diffplex implemented to compare strings in TestUtils.cs


smedilol:
PR too old and never used.

Issue #442, opened 2017-02-21
Performance measure


smedilol:
** CHILD OF #270 **

### The need
Our parser is currently too slow.
The first step to optimize is to have good metrics of the CPU time spent during execution.

### Solution
PerformanceCounter class seems to way to have that:
https://msdn.microsoft.com/fr-fr/library/system.diagnostics.performancecounter(v=vs.110).aspx
http://stackoverflow.com/questions/3411805/using-performancecounter-to-track-memory-and-cpu-usage-per-process
http://stackoverflow.com/questions/3896685/simplest-possible-performance-counter-example
http://stackoverflow.com/questions/23366831/c-sharp-performancecounter-list-of-possible-parameters

### To do
In NistTest, implement a CPU time measurement with PerformanceCounter (or any equivalent).

collarbe:
After some works on it : 
- PerformanceCounter is a bit heavy to use and need an administrator access level to run
- Using CPU Cycle is not enough safe. It completly depends on the CPU frequency of the machine. 
- StopWatch could be a good solution but it's also dependent of external factor. We may have to found a solution to accept a certain range of time elapsed for a test.. something like 30% difference or more.

smedilol:
@cobarzanc will try **pUnit** : 
http://www.bradoncode.com/blog/2012/02/benchmarking-methods-in-c-made-easier.html

2 new tests will be written to test this tool:
 - incremental
 - full parsing + code generation

# Incremental
Duplicate test TestIncrementalPipeline.CheckPerformance, 

Part from
`// Append one line in the middle of the program`
To
```
// Execute a second (incremental) compilation
compiler.CompileOnce();
```

Must be executed a lot of times..


On top of metrics provided by pUnit, we can use our own metrics:
```
compiler.CompilationResultsForProgram.PerfStatsForText.LastRefreshTime
compiler.CompilationResultsForProgram.PerfStatsForScanner.LastRefreshTime
compiler.CompilationResultsForProgram.PerfStatsForPreprocessor.LastRefreshTime
compiler.CompilationResultsForProgram.PerfStatsForCodeElementsParser.LastRefreshTime
compiler.CompilationResultsForProgram.PerfStatsForTemporarySemantic.LastRefreshTime 
compiler.CompilationResultsForProgram.PerfStatsForProgramCrossCheck.LastRefreshTime
```
These metrics will allow us to identify which phase is taking time.


# Full parsing + code generation
[...]

PR #479, opened 2017-03-23
Cpucycle implemented

parandba1:
Cpucycle implemented


smedilol:
After test, this technique isn't more reliable than an Watch that count the elapsed time.
See #442 to check our solution.

PR #560, opened 2017-05-22
Stacked Program + SourceFile Node

collarbe:


TypeCobol.Test/Compiler/Parser/ResultFiles/Programs/EmptyPGM.txt
```diff
@@ -1,3 +1 @@
-﻿--- Diagnostics ---
-Line 0[1,0] <27, Error, Syntax> - Syntax error : no viable alternative at input '<EOF>' RuleStack=cobolCompilationUnit,  OffendingSymbol=[1,0:]<EndOfFile>
-
+﻿
```
smedilol: It could be a thing to have at least a warning if a file is empty

TypeCobol.Test/Compiler/Parser/ParserUtils.cs
```diff
@@ -134,19 +134,25 @@ public static string DumpCodeElements(CodeElementsDocument result)
...
+        internal static string DumpResult(IEnumerable<Program> programs, IEnumerable<TypeCobol.Compiler.Nodes.Class> classes, IList<Diagnostic> diagnostics)
         {
             StringBuilder builder = new StringBuilder();
             if (diagnostics != null && diagnostics.Count > 0) {
                 builder.AppendLine(DiagnosticsToString(diagnostics));
             }
-            if (program != null) {
-                builder.AppendLine("--- Program ---");
-                Dump(builder, program);
+            if (programs != null && programs.Count() > 0) {
```
smedilol: Better use .Any().
See https://www.jetbrains.com/help/resharper/2017.1/UseMethodAny.0.html
or you can just check for null

TypeCobol.Test/Compiler/Parser/ParserUtils.cs
```diff
@@ -134,19 +134,25 @@ public static string DumpCodeElements(CodeElementsDocument result)
...
+                foreach (var program in programs)
+                {
+                    builder.AppendLine("--- Program ---");
+                    Dump(builder, program);
+                }
             }
-            if (cls != null) {
-                builder.AppendLine("--- Class ---");
-                Dump(builder, cls);
+            if (classes != null && classes.Count() > 0) {
```
smedilol: Better use .Any().
See https://www.jetbrains.com/help/resharper/2017.1/UseMethodAny.0.html
or you can just check for null

TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -562,6 +563,33 @@ public override bool VisitNode(IASTVisitor astVisitor)
...
+
+
+        public IEnumerable<Program> Programs {
+            get
+            {
+                return this.children.Where(c => c is Program && !((Program)c).IsNested).Select(c => c as Program);
+            }
+        }
+
+        public IEnumerable<Class> Clasees
```
smedilol: Typo: classes

TypeCobol.Test/Compiler/Parser/ResultFiles/Programs/EmptyPGM.txt
collarbe: Because of the new grammar `(cobolProgram | cobolClass)* EOF;` ANTLR does not generate an error on empty file. Also because it's an empty file, we can't add a diagnostic. Diagnostic are always associated to a line or at least a token, however, we don't have any line or token... 

I suggest to throw an exeption in FileCompiler, if TextDocument.Lines is null. It means that the empty test needs to be moved as a CLI test in order to get the exception as a diagnostic in error file. 

Do you agree ?

PR #653, opened 2017-08-29
Incremental 597

collarbe:
PR For incremental update. 

Wait for @reydelpa feedback before merging.


TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
```diff
@@ -92,6 +93,18 @@ internal static IList<DocumentChange<ICodeElementsLine>> ParseProcessedTokensLin
...
+                var minParseSection = refreshParseSections.OrderBy(p => p.StartLineIndex).First();
+                var maxParseSection = refreshParseSections.OrderByDescending(p => p.StopLineIndex).First();
+                largestRefreshParseSection = new ParseSection(minParseSection.StartLineIndex,
+                    minParseSection.StartToken, maxParseSection.StopLineIndex, maxParseSection.StopToken,
+                    maxParseSection.StopTokenIsFirstTokenOfTheLine);
+
+                refreshParseSections.Clear();
+                refreshParseSections.Add(largestRefreshParseSection);
+            }
+
```
smedilol: After reading the code, I understand that refreshParseSections can only contains 1 element.
In that case, I'd prefer to only keep `largestRefreshParseSection` and the `refreshParseSections` be removed.

@collarbe what do you think?

PR #656, opened 2017-09-02
Issue #503, You can now call a public proc from a proc.

osmedile:
Public proc called within the program that declare it are called as private procedure.
Also removed var IsLoaded in codegen of caller because it's not used anymore.


smedilol:
Oups, I pushed commit unrelated to that PR. I will create a new clean branch to merge this PR

smedilol:
Good commits are on branch CallPublicFromProc2

PR #657, opened 2017-09-04
Issue #503, You can now call a public proc from a proc.

smedilol:
Public proc called within the program that declare it are called as private procedure.
Also removed var IsLoaded in codegen of caller because it's not used anymore.


smedilol:
Merged manually into master

Issue #611, opened 2017-07-18
Diagnostics on CodeElement are misplaced


collarbe:
It seems (sometimes) that diagnostics on CodeElement are misplaced. 

For instance, a CodeElement error placed on line index 57, will generate a ParserDiagnostics on line 58 with a text message giving the line 59 in error on the document.. 

From line 57 (original error) to line 59... That is not correct 👎

PR #681, opened 2017-10-13
#611, big update on diagnostics (create diag on nodes + move Checker …

collarbe:
…diag to the node diag + manage diag incremental updates)

### DIFF
--- a/TypeCobol.Test/Compiler/Parser/ResultFiles/Programs/Cobol2002/Typedef-Nodes.txt
+++ b/TypeCobol.Test/Compiler/Parser/ResultFiles/Programs/Cobol2002/Typedef-Nodes.txt
@@ -8,29 +8,29 @@ Line 97[33,40] <27, Error, Syntax> - Syntax error : External is only allowed for
 Line 101[27,32] <27, Error, Syntax> - Syntax error : Elementary TYPEDEF cannot be STRONG
 Line 113[29,33] <27, Error, Syntax> - Syntax error : STRONG TYPEDEF cannot contain VALUE clause:
 Line 36[1,31] <30, Error, Semantics> - Semantic error: TYPE 'TypeNotDefined' is not referenced
-Line 45[14,20] <27, Error, Syntax> - Syntax error : TYPEDEF clause can only be specified for level 01 entries
-Line 50[29,35] <27, Error, Syntax> - Syntax error : extraneous input 'TYPEDEF' expecting {separator, statement starting keyword, keyword}
 Line 51[1,41] <30, Error, Semantics> - Semantic error: TYPE 'ClausesOrderKO' is not referenced
 Line 57[1,32] <30, Error, Semantics> - Semantic error: TYPEDEF 'NoDescription' has no description.
-Line 63[17,25] <27, Error, Syntax> - Syntax error : REDEFINES clause cannot be specified with TYPEDEF clause
 Line 64[1,37] <30, Error, Semantics> - Semantic error: TYPE 'NoRedefines1' is not referenced
-Line 68[17,25] <27, Error, Syntax> - Syntax error : REDEFINES clause cannot be specified with TYPEDEF clause
 Line 70[1,37] <30, Error, Semantics> - Semantic error: TYPE 'NoRedefines2' is not referenced
-Line 73[17,25] <27, Error, Syntax> - Syntax error : REDEFINES clause cannot be specified with TYPEDEF clause
 Line 73[1,49] <30, Error, Semantics> - Semantic error: Illegal REDEFINES: Symbol 'TypeNotDefined' is not referenced
 Line 75[1,37] <30, Error, Semantics> - Semantic error: TYPE 'NoRedefines3' is not referenced
 Line 82[5,57] <30, Error, Semantics> - Semantic error: Illegal REDEFINES as part of a TYPEDEF
 Line 90[5,62] <30, Error, Semantics> - Semantic error: Illegal REDEFINES: Symbol 'varNoRedefines5-a' is not referenced
-Line 93[26,33] <27, Error, Syntax> - Syntax error : EXTERNAL clause cannot be specified with TYPEDEF clause
-Line 97[33,40] <27, Error, Syntax> - Syntax error : External is only allowed for level 01
-Line 101[27,32] <27, Error, Syntax> - Syntax error : Elementary TYPEDEF cannot be STRONG
-Line 113[29,33] <27, Error, Syntax> - Syntax error : STRONG TYPEDEF cannot contain VALUE clause:
 Line 114[5,35] <30, Error, Semantics> - Semantic error: Illegal VALUE clause for subordinate 'TYPE-GRP-1' of STRONG TYPEDEF 'TYPE-GRP'
 Line 115[5,34] <30, Error, Semantics> - Semantic error: Illegal VALUE clause for subordinate 'TYPE-GRP-2' of STRONG TYPEDEF 'TYPE-GRP'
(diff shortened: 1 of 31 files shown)


TypeCobol/Compiler/Parser/CodeElementsLine.cs
```diff
@@ -68,34 +69,74 @@ internal void AddCodeElement(CodeElement codeElement)
...
+                }
+            }
+        }
+
+        internal void UpdateDiagnositcsLine()
+        {
+            //Update ParserDiag lines index
+            if (_ParserDiagnostics != null)
+            {
+                _ParserDiagnostics = _ParserDiagnostics.Select(d =>
```
smedilol: Why assign the result to _ParserDiagnostics ?

TypeCobol/Compiler/Parser/CodeElementsLine.cs
```diff
@@ -68,34 +69,74 @@ internal void AddCodeElement(CodeElement codeElement)
...
+                {
+                    d.Line = LineIndex+1;
```
smedilol: Why LineIndex + 1?

TypeCobol/Compiler/Parser/CodeElementsLine.cs
```diff
@@ -68,34 +69,74 @@ internal void AddCodeElement(CodeElement codeElement)
...
+            }
+
+            //Update CodeElements Diags lines index
+            if (CodeElements != null)
+            {
+                foreach (var codeElement in CodeElements)
+                {
+                    if (codeElement.Diagnostics != null)
+                    {
+                        codeElement.Diagnostics = codeElement.Diagnostics.Select(d =>
```
smedilol: Same questions here

TypeCobol/Compiler/Scanner/TokensLine.cs
```diff
@@ -151,9 +158,9 @@ internal void RemoveDiagnosticsForToken(Token token)
         {
             if (token != null)
             {
-                foreach (Diagnostic diag in GetDiagnosticsForToken(token).ToArray())
+                foreach (Diagnostic diag in _ScannerDiagnostics.Where(diag => diag is TokenDiagnostic && ((TokenDiagnostic)diag).Token == token).ToArray())
```
smedilol: Why not keep GetDiagnosticsForToken(token).ToArray() ?

TypeCobol/Compiler/Diagnostics/Cobol85Checkers.cs
```diff
@@ -17,6 +17,12 @@ namespace TypeCobol.Compiler.Diagnostics {
 
     public class Cobol85CompleteASTChecker : AbstractAstVisitor
     {
+        public Cobol85CompleteASTChecker(List<Diagnostic> diagnostics)
```
smedilol: Why add this property ?
A Checker doesn't need to know about diagnostics. They are attached to Token, Line, CodeElement, Node, ...
As I understand it, you use this property so ProgramClassParserStep can retrieve it and then these diagnostics are merged with diagnostics of ProgramClassDocument.

I think a proper solution (but perhaps slower) would be to read all Diagnostics on Node from the CompilationUnit.AllDiagnostics() method.


There will be less code in this Checker and the possible others Checkers.

TypeCobol.Test/Compiler/Parser/ResultFiles/Programs/Cobol2002/Typedef-Nodes.txt
```diff
@@ -8,29 +8,29 @@ Line 97[33,40] <27, Error, Syntax> - Syntax error : External is only allowed for
...
-Line 97[33,40] <27, Error, Syntax> - Syntax error : External is only allowed for level 01
-Line 101[27,32] <27, Error, Syntax> - Syntax error : Elementary TYPEDEF cannot be STRONG
-Line 113[29,33] <27, Error, Syntax> - Syntax error : STRONG TYPEDEF cannot contain VALUE clause:
 Line 114[5,35] <30, Error, Semantics> - Semantic error: Illegal VALUE clause for subordinate 'TYPE-GRP-1' of STRONG TYPEDEF 'TYPE-GRP'
 Line 115[5,34] <30, Error, Semantics> - Semantic error: Illegal VALUE clause for subordinate 'TYPE-GRP-2' of STRONG TYPEDEF 'TYPE-GRP'
 Line 132[5,62] <30, Error, Semantics> - Semantic error: Illegal REDEFINES: 'USE-TYPE-GRP-STRONG' is strongly-typed
 Line 145[6,66] <30, Error, Semantics> - Semantic error: Illegal REDEFINES: Symbol 'TYPE-GRP-STRONG-PART3' is not referenced
 Line 160[1,68] <30, Error, Semantics> - Semantic error: Illegal RENAMES: 'strongType1-PART1 IN useStrongType1' is strongly-typed
 Line 160[1,68] <30, Error, Semantics> - Semantic error: Illegal RENAMES: 'strongType1-PART2 IN useStrongType1' is strongly-typed
+Line 45[14,20] <27, Error, Syntax> - Syntax error : TYPEDEF clause can only be specified for level 01 entries
```
smedilol: Why these diagnostics now appear after semantic error ?

collarbe: Because those diagnostics are now created by the checker and assigned to ProgramClassDocument Diagnostics property. 

Also in method AllDiagnostics in compilation unit, ProgramClassDocumentSnapshot's diagnostics are return after codeElement diagnostics.

TypeCobol/Compiler/Parser/CodeElementsLine.cs
collarbe: ParserDiagnostics are now managed properly with a private filed. It allows to add, remove update and list the diagnostis attached to the line. It now allows us to do whatever we wants with the ParserDiagnostics (Like adding +1 to the diagnostics line for displaying correctly in the editor)

TypeCobol/Compiler/Parser/CodeElementsLine.cs
collarbe: Just a displaying purpose. Because ANTLR is 0 based, we need to return a +1 to diagnostics line in order to display it at the right place

TypeCobol/Compiler/Parser/CodeElementsLine.cs
collarbe: Same anwswer bu now it's affecting CodeElement Diagnostics and not parser diagnostics

TypeCobol/Compiler/Scanner/TokensLine.cs
collarbe: Yes, you're right, I'll change that.

TypeCobol/Compiler/Parser/ProgramClassParserStep.cs
```diff
@@ -85,21 +86,25 @@ public static void ParseProgramOrClass(TextSourceInfo textSourceInfo, ISearchabl
 
             //Create link between datas
             programClassBuilder.SyntaxTree.Root.AcceptASTVisitor(new TypeCobolLinker());
+            //Complete some information on Node and run checker that need a full AST
+            programClassBuilder.SyntaxTree.Root.AcceptASTVisitor(new Cobol85CompleteASTChecker());
+            programClassBuilder.SyntaxTree.Root.AcceptASTVisitor(new DiagnosticsChecker(diagnostics));
```
smedilol: As discussed today, you can move this DiagnosticsChecker to method AllDiagnotics of CompilationDocument.

PR #684, opened 2017-10-20
Cobol keywords update

collarbe:


TypeCobol/Compiler/Scanner/TokenType.cs
```diff
@@ -61,60 +63,46 @@ public enum TokenFamily
     public enum TokenType
     {
```
smedilol: Comment should be kept here

TypeCobol.Grammar/Grammars/Cobol/CobolWords.g4
```diff
@@ -727,8 +773,7 @@ superObjectIdentifier: SUPER;
 // EBCDIC
 // Specifies the EBCDIC character set.
 
-standardCollatingSequence: STANDARD_1 | STANDARD_2 | NATIVE | EBCDIC;
```
smedilol: Are you sure that all grammar rules that reference `standardCollatingSequence` don't need EBCDIC ?

Issue #813, opened 2018-01-29
Create a new step : cross-check between source files


smedilol:
**CHILD OF #811**

Currently step semantic, create the AST and run all checkers.
To allow to load dependencies dynamically we must separate these 2 steps:
- semantic + some simple checkers only on the same source file
- detect dependencies
- load dependencies (from filesystem or cache) and ignore dependencies that are also the target of the FileCompiler
- run cross-check between source file

This last step must create the snapshot ProgramClass

smedilol:
One thing missing:
GrammarTest should separate the time spent in Semantic and CrossCheck steps.

smedilol:
This can be done only in v1.0.11

Issue #811, opened 2018-01-29
Load dependencies in any order


smedilol:
**FOLLOWING #810**
We need to load dependency without any order

smedilol:
**FOLLOWING PR #818**, this issue should be resolved.

PR #818, opened 2018-02-01
#811 #813, implement cross check + allow to load dependencies in any …

collarbe:
…order


TypeCobol/Compiler/CompilationUnit.cs
```diff
@@ -161,29 +161,23 @@ public void RefreshProgramClassDocumentSnapshot()
...
                 {
-                    // Start perf measurement
-                    var perfStatsForParserInvocation = PerfStatsForProgramClassParser.OnStartRefreshParsingStep();
-
                     // Program and Class parsing is not incremental : the objects are rebuilt each time this method is called
-                    SourceFile root;
-                    List<Diagnostic> newDiagnostics;
-                    Dictionary<CodeElement, Node> nodeCodeElementLinkers = new Dictionary<CodeElement, Node>();
+                    SourceFile root = temporarySnapshot.Root;
+                    List<Diagnostic> diagnostics = temporarySnapshot.Diagnostics ?? new List<Diagnostic>();
```
smedilol: A snapshot only contains diagnostics related to its own phase.
It shall not include diagnostics from previous snapshot.

Issue #759, opened 2017-12-29
Update Intrinsic Type completion + make intrisic type more visible


collarbe:

collarbe:
@reydelpa Could you test the functionnality ? Intrinsic type will be marked by "*" symbol. Tell me if you would prefer something else.

collarbe:
Seems to be non-visible. Maybe the LSP client will have to implement a custom Logo to show the intrinsic types

collarbe:
Need to add a new symbol type in SymbolKind enumeration  `IntrinsicType = 19`

collarbe:
Wait for the client to be updated to support this functionnality. Then merge #759_Intrinsic-LSP branch.

PR #827, opened 2018-02-06
#759, add custom Kind for intrinsic type.

mayanje:
The Client now support Intrinsic Type Icon


smedilol:
Ok for me.
But in order to our languageServer with other client, I think it would be better to return this new intrinsic type only if the client say that it can handle it.

mayanje:
Your right these are TypeCobol custom rules, so we must use the" initialize" message and its ClientCapabilities parameters, with the experimental client capabilities field to set them. But our current server implémentations does not support ClientCapabilities, we must implement this feature in our TypeCobol.LanguageServer, So I am going to create an issue for that.

Issue #862, opened 2018-03-08
Create intrinsic type Currency


smedilol:
Create type currency like this:
```cobol
01 Currency typedef strict pic X(03).

01 var1 type Currency.
```
The problem is `currency` is a keyword.
This type must be created directly in the grammar like type `Date`

collarbe:
Not sure it's a good idea to create type like this inside the TypeCobol Parser code. 

I rather think that if `Currency` is a keyword, the developper should  use an other name for his type. Something like `CurrencyType`

smedilol:
@mayanje reminds me that we could use verbatim syntax like C# to declare Currency type:
```cobol
01 @Currency typedef strict pic X(3).
```

smedilol:
As discussed:
- We only need the currency type, not the amount. The amount type can be added with our intrinsic mechanism.
- No need for exchange rate

PR #914, opened 2018-04-24
Issue #862 : Create CURRENCY Type

Brochato:


Codegen/test/resources/input/TypeCobol/TypeCurrency.tcbl
```diff
@@ -0,0 +1,34 @@
+﻿IDENTIFICATION DIVISION.
```
smedilol: I don't see the output file in your commit.
Can you add it?

Issue #901, opened 2018-04-16
Dependency loading with duplicated program doest not work in case program name is less than 8 char


collarbe:
See Helper.cs from Line 95.

collarbe:
With a prgram name like this MYPGMTES345678965357687.tcbl and a dependency file like this MYPGMTES.tcbl. The code will not work properly and the program will be duplicated. 
We need to get the program name until the first dot or only the 8 first chars of the program name to compare. @Brochato can you do this fix ?

PR #923, opened 2018-04-26
issue #901 : add better name identification

Brochato:


Brochato:
It is an enhancement of an existing feature so the Unit Test is already done.

PR #936, opened 2018-05-14
#863 adding test files and clarifying error message

cobarzanc:
Clarifying the error message


TypeCobol.Test/Parser/Programs/TypeCobol/LibraryFunctionDeclarativeCheck.rdzPGM.txt
```diff
@@ -0,0 +1,20 @@
+﻿--- Diagnostics ---
+Line 30[8,26] <27, Error, Syntax> - Syntax error : Inside a library only function declaration or declaratives are allowed  / sentence-1
```
smedilol: This error message should be at line 38, because it's the `exit` statement that is not allowed.

How to fix this?
The problem here, is the test:
`DiagnosticUtils.AddError(child.CodeElement == null ? procedureDivision : child,`

`child.CodeElement` is null because a fake Node `Sentence` (named `sentence-1`) is inserted.
You should then check if the Node is a Sentence. If so then use the first child or iterate under this Sentence Node.

PR #943, opened 2018-05-17
940 set addr of to addr of

collarbe:


TypeCobol.Test/Parser/Programs/Cobol85/SetAddrsOf.rdz.cbl
```diff
@@ -0,0 +1,51 @@
...
+
+       linkage SECTION.
+       01 myDate TYPE DATE.
+       01 myDate2 TYPE DATE.
+       01 PicVar PIC X(10).
+       01 MyBool TYPE BOOL.
+
+
+       PROCEDURE DIVISION.
+
```
smedilol: Cobol doesn't allow to set an alphanumeric var or a typed var to something else.

These 3 lines are KO.

But perhaps it should be the subject of another issue.

PR #955, opened 2018-05-23
951 rule

cobarzanc:
Code analysis enabled at build for every build configuration. 
Ruleset is an example subject to change and updated depending on needs.


smedilol:
Good!
Where do you find the list of RuleSet ?

cobarzanc:
You can open the ruleset file with Visual Studio and it will display several rule categories. If you change the default behaviour of a rule, it will find itself in the ruleset file with the new behaviour.

smedilol:
@mayanje and @collarbe is it ok for you to merge this PR?

Issue #953, opened 2018-05-22
Implement SET TO with UNSAFE


collarbe:
Add support of UNSAFE keywork and rules for SET TO statement.

Issue #950, opened 2018-05-22
NullReferenceException with Renames


smedilol:
The following renames is not valid in Cobol (renames on 01 is not allowed), but it should not throw an exception:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.

       01 var1 pic X(10).
       66 var2 renames var1.
       END PROGRAM DVZF0OSM.
```

collarbe:
This issue is refering to this exception : 
```
System.NullReferenceException:
   at TypeCobol.Compiler.Diagnostics.RenamesChecker.Check (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Diagnostics.CrossCompleteChecker.BeginNode (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
   at TypeCobol.Compiler.CompilationUnit.RefreshProgramClassDocumentSnapshot (TypeCobol, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null)
[...]

Issue #941, opened 2018-05-15
CodeGen: Crash when accessing a depending on variable inside an unnamed group.


mayanje:
The following declaration and usage lead to a null reference exception, when the cnt depending on variable is accessed by the Code Generatror.

```cobol
       01 CallingStack typedef strict private.
         05 cnt  pic S9(5) comp-5.
         05 moduleList.
          10 occurs 40
           depending on cnt.
           15 module pic X(8).

       01 callingStack type CallingStack.
```

Issue #940, opened 2018-05-14
Set adrs of to adrs of with same variable type does not work 


collarbe:
The following code create an error but it should not : 

```COBOL
WORKING-STORAGE SECTION.
01 myDate TYPE DATE.
01 myDate2 TYPE DATE.
01 PicVar PIC X(10).

PROCEDURE DIVISION.
   * Create error 
       SET ADDRESS OF mydate2 to ADDRESS OF mydate.
   * Create error
       SET ADDRESS OF PmyDate TO ADDRESS OF PicVar
    * No error 
        SET ADDRESS OF PicVar TO ADDRESS OF mydate.

```

smedilol:
The target variable of `set address of` must be a level 01 in linkage.

The following code create an error but it should not : 

```COBOL
WORKING-STORAGE SECTION.
01 W-myDate TYPE DATE.
01 W-myDate2 TYPE DATE.
01 W-PicVar PIC X(10).

linkage SECTION.
01 myDate TYPE DATE.
01 myDate2 TYPE DATE.
01 PicVar PIC X(10).

PROCEDURE DIVISION.
* Create error because it's in working-storage
       SET ADDRESS OF W-mydate2 to ADDRESS OF W-mydate.
* Create error because it's in working-storage
       SET ADDRESS OF W-myDate TO ADDRESS OF W-PicVar
* Create error because it's in working-storage
        SET ADDRESS OF W-PicVar TO ADDRESS OF W-mydate.



* Ok, because the left part of SET use a linkage variable
       SET ADDRESS OF mydate2 to ADDRESS OF W-mydate.
* Ok, because the left part of SET use a linkage variable
       SET ADDRESS OF myDate TO ADDRESS OF W-PicVar
* Ok, because the left part of SET use a linkage variable
        SET ADDRESS OF PicVar TO ADDRESS OF W-mydate.


* Should not create error 
       SET ADDRESS OF mydate2 to ADDRESS OF mydate.
* Should not create error 
       SET ADDRESS OF myDate TO ADDRESS OF PicVar
* No error 
        SET ADDRESS OF PicVar TO ADDRESS OF mydate.

```

PR #956, opened 2018-05-23
Merge v1.1.3 to master

collarbe:
## What's New 

- We've had a new instrinsic type which is `Currency` #862
- Add support of LENGTH OF (by content) with procedure call #868 
- Add support for completion for ZERO/SPACE keywords #845 
- Add variable check for WHEN statement #921 
- Add support of UNSAFE keyword in SET ... TO ... statement #953 
- Add support of completion on OF statement #946 

## What's fixed 

- Ambigous symbol diagnostics contains mor information about the founded symbols #716 
- Fix some issue when trying to parse a program with the same name as a dependency #901 
- Typo for PIC statement is now verified  #896 
- Fix several issues linked to MOVE CORR statement #911 #912 #913 #906
- Fix potential NullReferenceException with completion of QualifiedName #905 
- Fix NullReferenceException for a non-named level 07 variable #844 
- Fix completion of variable in Nested program (Nested program name was part of the completion) #902 
- Fix code generation with a continuation line inside a type definition #909 
- Fix code generation line exceed column 73 #928 
- Fix code generation with Global keyword #924 
- Clarify error message for "Illegal non-function or paragraph item in library" #863 
- Fix Global clause for children of a global group #927 
- Fix NullReferenceException on SUBTRACT CORR #935 
- Code generation crashed when accessing a depending on variable inside an unnamed group #941 
- Fix Set ADDRESS OF ... TO ADDRESS OF ... of a same type variable #940 
- Fix NullReferenceException on renames + Add rules for renames with a level 01 variable #950 
- Fix completion support of Global variable inside nested program #949


smedilol:
After second thought/test:
The problem with #948 is already present with version 1.1.2 so it's ok to merge.
We just have to warn developers that `move corr` will check children type in the future

Issue #960, opened 2018-05-25
Forbird SYNC clause with Strict Type


smedilol:
It seems that Cobol 2002 type comparison is more complicated that what we implemented:

> Two typed items are of the same type when:
> - The items are described with TYPE clauses that reference equivalent type declarations; or
> - The items are described as subordinate items in equivalent type declarations, starting at the same relative byte position and having the same length in bytes.
> 
> 
> Two type declarations are considered equivalent when they have the same type-name, and for each 
> elementary item in one type declaration there is a corresponding elementary item in the other type 
> declaration, starting at the same relative byte position and having the same length in bytes. 
> Each pair of corresponding elementary items shall have the same BLANK WHEN ZERO, DYNAMIC LENGTH, JUSTIFIED, PICTURE, SIGN, SYNCHRONIZED, and USAGE clauses, with the following exceptions:
> 1) Currency symbols match if and only if the corresponding currency strings are the same.
> 2) Period picture symbols match if and only if the DECIMAL-POINT IS COMMA clause is in effect for both or for
> neither of these type declarations. Comma picture symbols match if and only if the DECIMAL-POINT IS
> COMMA clause is in effect for both or for neither of these type declarations.
> 
> Additionally, locale specifications in the PICTURE clauses match if and only if:
> - both specify the same SIZE phrase in the LOCALE phrase of the PICTURE clause, and
> - both specify the LOCALE phrase without a locale-name or both specify the LOCALE phrase with the same
> external identification, where the external identification is the external-locale-name or literal value
> associated with a locale-name in the LOCALE clause of the SPECIAL-NAMES paragraph.
> 

> NOTE:  When two typed items are defined in the same source element, the above rules mean that either both items are described with the same TYPE clause or both items are described as the same subordinate item in the same type declaration.

As we only check type name:
For Scrict type which is our own definition we need to forbid Sync clause to ensure type really match at runtime.

- [ ] Forbid SYNC in a Typedef
   -Already done at the typedef level (d1686a4) but not for data under the typedef.
- [ ] A strictly typed variable cannot be SYNC
  - SYNC can be define at a parent level.

smedilol:
Also forbid SYNC for data under the typedef

smedilol:
See commit eec9eb99b6717f88d471a691943e56828dad9b7a

Issue #948, opened 2018-05-18
Add Group children check for MOVE CORR


collarbe:
MOVE CORR is a statement that allows to move a group of varaible to an another via correspondance between variable's name in both group. 

```COBOL
01 MyGroup1.
  05  Data1 PIC 10.
  05  Data2 TYPE DATE.

01 MyGroup2.
   05 Data1 PIC 10.
   05 Data2 TYPE BOOL.

MOVE CORR MyGroup1 TO MyGroup2.
```

The MOVE CORR statement in this example has to check if all the corresponding childrens are compatible. In this case Data2 from MyGroup1 and MyGroup2 are not compatible. It has to generate an error to tell the user about it.

Issue #945, opened 2018-05-17
Crash with completion on a group that contains a filler


smedilol:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 Group1.
          05 var1                type date.
          05 filler pic X.

       01 var2                      PIC X(4).

       PROCEDURE DIVISION.
      *Crash if completion is invoked after ::
            move Group1:: to var2

       END PROGRAM MyPgm.
```

If you remove the filler under Group1 or name, it works fine.

Issue #942, opened 2018-05-17
Check call with ADDRESS OF


smedilol:
`ADDRESS OF` used in Cobol call after `using` clause or in a procedure in `input`, `in-out`, `output` must follow these rules :
- `ADDRESS OF` with a linkage variable always works
- `ADDRESS OF` with a variable not in linkage (working-storage for example), only works if the `ADDRESS OF` is `BY CONTENT` or `BY VALUE`.

PR #975, opened 2018-06-07
PR for v1.1.4

collarbe:
## What's New 

- Calls with ADDRESS OF now check if variable comes from LINKAGE SECTION, or if it's given BY CONTENT/BY VALUE #942 
- You can increment/decrement Pointer with `SET MyPointer UP BY x / DOWN BY X` #780 (Work in progress)
- Add MOVE CORR check for corresponding typed children #948 
- Our concatenated file format with generated code and TypeCobol commented now supports compiler options #932
- Forbid SYNC clause with STRICT Type #960

## What's fixed 

- Module name are now generated in uppercase #886 
- Fix linkage variable wrong export while doing PublicSignature export #784 
- Fix bug with completion used with a group containing a filler #945 
- Fix completion bug if the procedure name case does not match the declared procedure name case #918 
- Fix TC-Call variable definition in generated code to match our Bool type generation #919 
- Fix code generation for boolean used with LENGTH OF / ADDRESS OF #860 
- Fix code generation for LENGTH OF / ADDRESS OF inside standard COBOL Call #861 
- Fix duplicated program openning with LSP Server #583 
- Fix potential Null Reference Exception with the use of LENGTH OF in procedure call #868


collarbe:
@smedilol  is the release description ok ?

smedilol:
Yes it's ok, I've just updated 3 lines.

PR #988, opened 2018-06-13
Add mixed transform output with CLI

collarbe:
Can you check this PR and tell if it's ok for you


Codegen/src/Generators/MixedTransformGenerator.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using TypeCobol.Codegen.Skeletons;
+using TypeCobol.Compiler;
+using TypeCobol.Compiler.Nodes;
+using TypeCobol.Compiler.Text;
+
+namespace TypeCobol.Codegen.Generators
+{
+    public class MixedTransformGenerator : DefaultGenerator
```
smedilol: I'd rather use a GeneratorWrapper instead to inherit from DefaultGenerator.
This will allow to combine the MixedTransformGenerator with any other Generator.

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -135,6 +135,7 @@ public enum OutputFormat {
         Cobol85,
         PublicSignatures,
         ExpandingCopy,
+        MixedTransform
```
smedilol: Can you name it "Cobol85Mixed" ?

Issue #996, opened 2018-06-19
Remove unuseful code in CrossChecker ?


smedilol:
Issue only apply if #977 is merged into develop.

Method 
`public static DataDefinition CheckVariable(Node node, StorageArea storageArea, bool isReadStorageArea)` in CrossChecker, search for existing data definitinon before constructing one.

But as CheckVariable is only called once per StorageArea, this code shouldn't be useful.

Try to see if we can remove this code:
```C#
//search for existing data definitinon before constructing one
            Tuple<string,DataDefinition> searchExistingDataDefinition;
            if (isReadStorageArea)
            {
                node.StorageAreaReadsDataDefinition
                    .TryGetValue(storageArea, out searchExistingDataDefinition);
            }
            else
            {
                node.StorageAreaWritesDataDefinition
                    .TryGetValue(storageArea, out searchExistingDataDefinition);
            }
            if (searchExistingDataDefinition!=null)
            {
                IndexAndFlagDataDefiniton(searchExistingDataDefinition.Item1, searchExistingDataDefinition.Item2,node,area,storageArea);
                return searchExistingDataDefinition.Item2;
            }
```

cobarzanc:
After analysis, the highlighted code is not useful anymore. CheckVariable method is used only to build the dictionaries, so we should not make any search in the dictionaries.

PR #1025, opened 2018-07-03
Issue #996 - Removed unused code in CheckVariable

cobarzanc:
Updated Visit(Move move) overriden method

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
@@ -102,17 +102,9 @@ namespace TypeCobol.Compiler.Diagnostics
                 DataDefinition fromVariable = null;
                 DataDefinition toVariable = null;
                 //For MoveCorrespondingStatement check children compatibility
-                if (move.StorageAreaReadsDataDefinition!=null && move.StorageAreaReadsDataDefinition.TryGetValue(moveCorresponding.FromGroupItem,
-                    out searchedDataDefintion))
-                {
-                    fromVariable = searchedDataDefintion.Item2;
-                }
-
-                if (move.StorageAreaWritesDataDefinition!=null && move.StorageAreaWritesDataDefinition.TryGetValue(moveCorresponding.ToGroupItem,
-                    out searchedDataDefintion))
-                {
-                    toVariable = searchedDataDefintion.Item2;
-                }
+                fromVariable = move.GetDataDefinitionFromStorageAreaDictionary(moveCorresponding.FromGroupItem, true);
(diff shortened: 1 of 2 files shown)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -102,13 +102,13 @@ public override bool Visit(Move move)
                 DataDefinition fromVariable = null;
                 DataDefinition toVariable = null;
                 //For MoveCorrespondingStatement check children compatibility
-                if (move.StorageAreaReadsDataDefinition.TryGetValue(moveCorresponding.FromGroupItem,
+                if (move.StorageAreaReadsDataDefinition!=null && move.StorageAreaReadsDataDefinition.TryGetValue(moveCorresponding.FromGroupItem,
```
smedilol: Why check for null here?
StorageAreaReadsDataDefinition must always be populated by CheckVariable before we enter this code.

cobarzanc: I changed the place where the dictionaries are initialized. In CheckVariable only when a single DataDefintion is found the dictionaries are initialized, if they are null. 
In the case where a variable is not defined (does not have a DataDefintion) in CheckVariable the dictionary will not be initialized anymore, so a check for null in the Visit(Move mode) method is needed.
I do think that this is a better choice, as dictionaries will be initialized only where needed, so memory consumption is lowered.

smedilol: Then can you force the use of method Node.GetDataDefinitionFromStorageAreaDictionary instead of direct access to StorageAreaReadsDataDefinition and StorageAreaWritesDataDefinition.

This will avoid to remember to check for null everywhere in the code.

cobarzanc: Not quite. Node.GetDataDefinitionFromStorageAreaDictionary will search in both StorageAreaReadsDataDefinition and StorageAreaWritesDataDefinition dictionaries. 
In Visit(Move mode) method we need to extract the to and from members which come from either StorageAreaReadsDataDefinition (from member) or StorageAreaWritesDataDefinition (to memeber). Thus we cannot use GetDataDefinitionFromStorageAreaDictionary  because we do not know from what dictionary the retuned DataDefinition comes.

For now, in Visit(Move mode) is the only place where a null check was needed before searching a value in the StorageArea dictionaries. 

If you want to avoid null checks in future, I can move the dictionary initialization where it was before in CheckVariable method. However, in my opinion, this will degrade more the performance than making a null check where is needed, as it will allocate memory for 2 dictionaries (and they may not be used).

Issue #1053, opened 2018-07-13
Opening an empty document create an exception in LS


smedilol:
![exception](https://user-images.githubusercontent.com/13764482/42682665-872de6fe-868b-11e8-8faa-5c9892c615e7.png)

PR #1058, opened 2018-07-20
Issue #1053, Added condition to code executing on empty cobolTextLine

Tenember:


Tenember:
The user still need to save the file before being able to use the autocompletion ; so this may juste be a workaround and not a definitive solution. (i.e the issue may be more complex)

PR #1059, opened 2018-07-23
Improve symbol table

cobarzanc:
The changes enhance the usage of DataTypeDefinition dictionary


TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -621,72 +613,61 @@ private void AddAllReference(IList<DataDefinition> found, DataDefinition heaData
...
+        /// <summary>
+        /// Search the qualified name in the DataTypeEntries dictionary
+        /// </summary>
+        /// <param name="name">QualifiedName</param>
+        /// <param name="table">SymbolTable that can have the searched DataDefinition</param>
+        /// <returns>List of found DataDefinitions that correspond to the passed name</returns>
+        private static IEnumerable<DataDefinition> GetCustomTypesFromSymbolTable(QualifiedName name, SymbolTable table)
+        {
+            //match only last part of the qualified name - the most specific one regardless of the case
+            return table.DataTypeEntries.Where(elem => elem.Key.ToArray().Last().Equals(name.ToArray().Last(),StringComparison.OrdinalIgnoreCase)).SelectMany(def => def.Value);
```
smedilol: As the key and the parameter of this method would be string (see previous remarks), then you use TryGet instead of a Where.

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -824,6 +805,26 @@ public void AddType(TypeDefinition type)
...
+        public void AddTypeDataDefinition(TypeDefinition data)
+        {
+            //Add type if it doesn't have children
+            if (data.Children.Count == 0)
+            {
+                Add(DataTypeEntries, data);
+            }
+
+            //If type has children, add only the children
+            foreach (var elem in data.Children)
```
smedilol: Why not also copy children of children recursively ?

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -72,7 +72,15 @@ public SymbolTable CustomSymbols
                 }
                 // TODO#249: use a COPY for these
                 foreach (var type in DataType.BuiltInCustomTypes)
-                    TableOfIntrisic.AddType(DataType.CreateBuiltIn(type)); //Add default TypeCobol types BOOLEAN and DATE
+                {
+                    var createdType = DataType.CreateBuiltIn(type);
+                    TableOfIntrisic.AddType(createdType); //Add default TypeCobol types BOOLEAN and DATE
+                    //Add children to dictionary in Intrinsic symbol table
+                    if (createdType.Children.Count != 0)
```
smedilol: Why count children here?
If a Type has no children, method AddTypeDataDefinition handle this case + it also have a behavior when there is no children.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -462,19 +469,15 @@ public virtual void StartDataConditionEntry(DataConditionEntry entry)
...
+                node.SymbolTable.AddVariable(node);
         }
 
         public virtual void StartTypeDefinitionEntry(DataTypeDescriptionEntry typedef)
         {
             SetCurrentNodeToTopLevelItem(typedef.LevelNumber);
             var node = new TypeDefinition(typedef);
             Enter(node);
-            SymbolTable table;
-            if (node.CodeElement().IsGlobal) // TCTYPE_GLOBAL_TYPEDEF
```
smedilol: Comment `TCTYPE_GLOBAL_TYPEDEF` refers to a rule into our wiki, please add it back

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -122,8 +121,8 @@ public override bool Visit(Move move)
                         true; //Do not continue, the variables hasn't been found. An error will be raised later by CheckVariable()
                 }
 
-                var fromVariableChildren = fromVariable.Children.Where(c => c?.Name != null);
-                var toVariableChildren = toVariable.Children.Where(c => c?.Name != null);
+                var fromVariableChildren = fromVariable.Children.Where(c => c?.Name != null).AsQueryable();
```
smedilol: Why use AsQueryable here?

TypeCobol/Compiler/CodeModel/SymbolTable.cs
cobarzanc: They will be copied, as the method is called at each level of a TypeDef

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
cobarzanc: AsQueryable() -  query that returns a sequence of values do not consume the target data until the query object is enumerated - also known as deferred execution.

For more references:
https://msdn.microsoft.com/en-us/library/bb507003(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.linq.enumerable(v=vs.110).aspx

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -1184,7 +1181,12 @@ public List<Program> GetPrograms()
         /// <param name="symbol"></param>
         private void Add<T>([NotNull] IDictionary<string, List<T>> table, [NotNull] T symbol) where T : Node
         {
-            string key = symbol.QualifiedName.Head;
+            //QualifiedName of symbol can be null - if we have a filler in the type definition
+            if (symbol.QualifiedName == null)
+            {
+                return;
+            }
+            string key = symbol.QualifiedName.Head.ToLowerInvariant();
```
smedilol: Dictionaries in SymbolTable use StringComparer.OrdinalIgnoreCase for comparison.
Maybe it's faster to put the key in lower case and convert the searched value into lower case.
But with your code, the job is half done. You convert the key to lower case, but the Dictionary still use OrdinalIgnoreCase and the value searched is not converted in lower case.


Conversion to lower case is outside the scope of this issue, so please remove it from this PR.

TypeCobol/Compiler/CodeModel/SymbolTable.cs
smedilol: It's not the case.
All children of children of an intrinsic type are not recognized anymore.

cobarzanc: OK, I will add recursively the children's children

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -148,62 +148,51 @@ public SymbolTable(SymbolTable enclosing, Scope current)
...
+
         internal void AddVariable([NotNull] DataDefinition symbol)
         {
             // TODO: generate a name for FILLERs and anonymous data to be referenced by in the symbol table
             if (symbol.Name == null)
             {
                 return;
             }
-            Add(DataEntries, symbol);
+            //If symbol is a part of a type add to DataTypeEntries
```
smedilol: The comment is not ok.

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -148,62 +148,51 @@ public SymbolTable(SymbolTable enclosing, Scope current)
...
+        void AddTypeVariableFromDataDefinition([NotNull] DataDefinition data)
         {
-            string key = name;
-            List<DataDefinition> found;
-            bool present = DataEntries.TryGetValue(key, out found);
-            if (!present)
+            //Types are declared in the Declarations SymbolTable
+            var table = GetTableFromScope(Scope.Declarations);
+
+            if (table == null)
```
smedilol: If table is null we want to know it.
So assume it's not null so we can get the exception.

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -818,6 +795,43 @@ public void AddType(TypeDefinition type)
...
 
+        public void AddTypeDataDefinition(TypeDefinition data)
+        {
+            //Add type if it doesn't have children
+            if (data.Children.Count == 0)
+            {
+                Add(DataTypeEntries, data);
+            }
+
+            //If type has children, add the children
```
smedilol: This code is easier to read:
```c#
           //Add the children
            AddTypeChildDataDefinition(data);
            
        }

        public void AddTypeChildDataDefinition(DataDefinition data)
        {
            
            foreach (var dataChild in data.Children)
            {
                var childDataDefinition = dataChild as DataDefinition;
                if (childDataDefinition == null)
                {
                    continue;
                }
                Add(DataTypeEntries, childDataDefinition);
                //Continue to add further children
                AddTypeChildDataDefinition(childDataDefinition);
            }
        }
```

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -148,62 +148,51 @@ public SymbolTable(SymbolTable enclosing, Scope current)
...
+            var table = GetTableFromScope(Scope.Declarations);
+
+            if (table == null)
+            {
+                return;
+            }
+            //Add symbol to the dictionary
+            Add(table.DataTypeEntries, data);
+            //Add indexes as they need special treatment
+            foreach (var child in data.GetChildren<IndexDefinition>())
```
smedilol: IndexDefinition are added to SymbolTable directly by ProgramClassBuilder.
However, current ProgramClassBuilder doesn't add them to SymbolTable if it's under a TypeDef.


So remove "//Add indexes as they need special treatment" and make ProgramClassBuilder add Index into SymbolTable (check for comment "//If index is inside a Typedef do not add to symboltable" in ProgramClassBuilder.).

TypeCobol/Compiler/CodeModel/SymbolTable.cs
cobarzanc: If I do this, the types with no children will not be added to the DataTypeEntries dictionary (the foreach will not be played if children.count==0).
At the end of the recursivity, the entry will not have any more children, so it will be added to the dictionary.

smedilol:
I reworked commits : merge almost all of them into one single commit.
This will ease further code review.
I pushed new commits directly to develop.

Issue #1039, opened 2018-07-10
TypeCobolLinker : don't visit Node that cannot contains type usage


smedilol:
The following Node cannot contains children which use a type, so we need to stop the TypeCobolLinker 
 visitor we enter them:
- Paragraph
- Section
- IdentificationDivision
- EnvironmentDivision
- ProcedureDivision inside Procedure
   - ProcedureDivision inside a program (or NestedProgram or StackedProgram) must be visited
- Declaratives

Edit : clarify that it only concern TypeCobolLinker  visitor.
The behavior of others visitors must remain unchanged.

PR #1060, opened 2018-07-23
WI #1039 - Nodes no longer visited

cobarzanc:
Stop visiting nodes that have:
-Paragraph
-Section
-IdentificationDivision
-EnvironmentDivision
-ProcedureDivision inside Procedure
--ProcedureDivision inside a program (or NestedProgram or StackedProgram) must be visited
-Declaratives

### DIFF
--- a/Codegen/src/Nodes/ParagraphGen.cs
+++ b/Codegen/src/Nodes/ParagraphGen.cs
@@ -31,11 +31,11 @@ namespace TypeCobol.Codegen.Nodes
             }
         }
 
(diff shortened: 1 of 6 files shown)


smedilol:
Issue #1039 was only about TypeCobolLinker visitor
The title of the issue is "TypeCobolLinker : don't visit Node that cannot contains type usage" but I didn't repeat that it only concern TypeCobolLinker  in the issue. Hence the confusion.

The behavior of others visitors must remain unchanged.


I propose to create a new PR for this issue.

Issue #1051, opened 2018-07-13
Don't always make a copy of Children Node when visiting a Node


smedilol:
Currently Node class copy its children into an array before calling the visitor.
This is to protect against children modification during the visit of the Node.

This array duplication can be time consuming.

But only the codegen visitor modify children.
So create a property in the Visitor `CanModifyChildrenNode` so the Node know if it must copy its children or not.
This property will be false by default in the AbstractVisitor.
Only codegen visitor will set it to true.

PR #1061, opened 2018-07-23
WI #1051 Optimize visitor on children nodes

cobarzanc:
- only CodeGen visitor will make a copy of Node's children before visit, as it can modify them

### DIFF
--- a/TypeCobol/Compiler/Nodes/Node.cs
+++ b/TypeCobol/Compiler/Nodes/Node.cs
@@ -628,7 +628,7 @@ namespace TypeCobol.Compiler.Nodes {
                 }
                 else
                 {
-                    continueVisit = VisitChildrens(astVisitor, true, children.ToArray());
+                    continueVisit = VisitChildrens(astVisitor, true, children);
                 }
                 
             }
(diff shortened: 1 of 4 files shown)


TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -612,29 +612,46 @@ public Node Get(string uri, int startIndex)
...
-                        break;
-                    }
-                    continueVisit = child.AcceptASTVisitor(astVisitor);                    
+                    //Make copy of  children array if the visitor can modify the children
+                    Node[] childrenNodes = children.ToArray();
+                    continueVisit = VisitChildrens(astVisitor, true, childrenNodes);
+                }
+                else
+                {
+                    continueVisit = VisitChildrens(astVisitor, true, children.ToArray());
```
smedilol: I think the `ToArray()` should be removed.

cobarzanc: `ToArray()` can be removed if the VisitChildren signature is modified to `VisitChildrens(IASTVisitor astVisitor, bool continueVisit, IEnumerable<Node> childrenNodes)` from `VisitChildrens(IASTVisitor astVisitor, bool continueVisit, Node[] childrenNodes)`

Issue #828, opened 2018-02-06
Better diagnostic when there is a small error in procedure or program name


smedilol:
if you declare a Pgm `TCOZDATE` with a procedure `convertToDate`.

If a caller use: `TCOSDATE::convertToDate` (`S` instead of `Z`)
we should:
- say that the program doesn't exists
- try to find if a similar program name exists and propose it

If a caller use: `TCOZDATE::convert` (`convert` instead of `convertToDate`)
we should:
- say that the procedure doesn't exists (already done)
- try to find if a similar procedure name exists and propose it

Tenember:
Feature to find a similar program or function on hold because of performance issue risk

smedilol:
Yes maybe, we should add a new "slow-check" phase for this kind of rules.

PR #1075, opened 2018-07-31
Issue #828, Diagnostic on program name incorrect  

Tenember:


smedilol:
PR too old and never used.

Issue #1071, opened 2018-07-30
Pointers redefindes are not generated in case of function parameter


Brochato:
If a pointer is incremented inside a function, the generated propram does not contain the pointer redefines.

```cobol
         declare procedure BugPointer PRIVATE
         input ptr pointer
         .
         data division.
         working-storage section.
         procedure division.
             set ptr up by 1
             goback
             .
         end-declare.
```

will generate

```cobol
       declare procedure BugPointer PRIVATE
       input ptr pointer
       .
       IDENTIFICATION DIVISION.
       PROGRAM-ID. b233702fBugPointer.

       data division.
       working-storage section.
       LINKAGE SECTION.
       01 ptr pointer.
       PROCEDURE DIVISION
             USINGBY REFERENCE ptr
           .
      *    set ptr up by 1
           ADD 1 to ptrdbef1f3d

           goback
           .
       END PROGRAM b233702fBugPointer.


```

PR #1076, opened 2018-07-31
Issue  #1071: redefines generation inside function

Brochato:
Add redefines generation for pointer inside function + Unit Test


Codegen/src/Nodes/ParameterEntry.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System.Linq;
+using System.Security.Permissions;
```
collarbe: Why adding this using ?

Issue #978, opened 2018-06-11
Procedure overload with POINTER does not work


collarbe:
The code below creates an error on MyProc, only when POINTER is used. 

```COBOL
       DECLARE PROCEDURE MyProc PRIVATE
           INPUT    myPtn POINTER.
        END-DECLARE.

       DECLARE PROCEDURE MyProc PRIVATE
           INPUT    NumericInput PIC S9(9) COMP-5.
        END-DECLARE.
```

smedilol:
To fix all cases regarding overload, can you also add DataType for:
- procedure-pointer
- object reference
- function-pointer
- Index

PR #1080, opened 2018-08-02
Issue #978, fix POINTER handling with procedure

collarbe:
WARNING : Potential breaking changes in this pull request. 

I had a new DataType for Pointer. This DataType is now used for parameter of type pointer. The hash generation for procedures using a pointer is changed. (See unit test update) 

If you merge it to develop like this, you need to merge PR #1079 first.


smedilol:
TypeCobol programs in our company use pointer parameter but they are not in production yet.
So we can make this change.

I'll ask them directly if they agree to re-generate/compile all of their TypeCobol programs.

smedilol:
PR too old and never used.

PR #1085, opened 2018-08-08
Global Storage Section (parser rules only, no generation)

collarbe:


TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -816,38 +859,106 @@ public static void CheckStatement(Node node)
...
+        {
+            var dataDefinition = node as DataDefinition;
+            if (dataDefinition == null) return;
+
+            //Check variable LevelNumber Rule - GLOBALSS_LIMIT48 
+            var data = dataDefinition.CodeElement as DataDefinitionEntry;
+            if (data?.LevelNumber != null && data.LevelNumber.Value == 77)
+                DiagnosticUtils.AddError(node,
+                    "Data must be declared between level 01 to 49, or equals to 66 or 88.");
+            if (!(dataDefinition.Parent is DataDefinition) && data?.LevelNumber != null &&
```
smedilol: This check must be done by a Checker common to all DataDefinition. It can be removed.

TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -816,38 +859,106 @@ public static void CheckStatement(Node node)
...
+        private static void CheckGlobalStorageChildren(Node node)
+        {
+            var dataDefinition = node as DataDefinition;
+            if (dataDefinition == null) return;
+
+            //Check variable LevelNumber Rule - GLOBALSS_LIMIT48 
+            var data = dataDefinition.CodeElement as DataDefinitionEntry;
+            if (data?.LevelNumber != null && data.LevelNumber.Value == 77)
+                DiagnosticUtils.AddError(node,
+                    "Data must be declared between level 01 to 49, or equals to 66 or 88.");
```
smedilol: The error message would be more clear with:
"Level 77 is forbidden in global-storage section."

TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -816,38 +859,106 @@ public static void CheckStatement(Node node)
...
+            if (!(dataDefinition.Parent is DataDefinition) && data?.LevelNumber != null &&
+                data.LevelNumber.Value == 88)
+                DiagnosticUtils.AddError(node, "Level 88 data must be declared inside a data from level 01 to 49.");
```
smedilol: This check must be done by a Checker common to all DataDefinition. It can be removed.

TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
smedilol: Error message would be more clear with : "Level 77 is forbidden in global-storage section"

PR #1092, opened 2018-08-13
729 datadeclaration

cobarzanc:


smedilol:
I rewrote the code and pushed directly to develop

Issue #1086, opened 2018-08-08
Declaratives does not take all lines in account


collarbe:
```COBOL
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOZFCPT.

       ENVIRONMENT DIVISION.
       CONFIGURATION SECTION.
       SOURCE-COMPUTER. IBM-370
            WITH DEBUGGING MODE
            .
       SPECIAL-NAMES.  DECIMAL-POINT IS COMMA.

      *=================================================================
       PROCEDURE DIVISION.
      DDECLARATIVES.
      DDECLARATION SECTION.
      D      USE FOR DEBUGGING ON ALL PROCEDURES.
      DAFFICHAGE-PARAGRAPHE.
      D    display "in main program".
      DEND DECLARATIVES.
        .
       END PROGRAM TCOZFCPT.
```
The line      ` D      USE FOR DEBUGGING ON ALL PROCEDURES.` is not parse

Issue #1078, opened 2018-08-01
Double paranthese make parser crash


collarbe:
This code : 
```COBOL
  IF AYIDDM-PtlDta-Value
                            ((AYIDDM-Idx-DmlPtl idx-RpdSrch)
                            = spaces or low-values)
                            CONTINUE 
```

make parser crash with exception : 
  ```
 à TypeCobol.Compiler.Parser.CobolExpressionsBuilder.CreateConditionalExpression(ConditionalExpressionContext context)
   à TypeCobol.Compiler.Parser.CobolStatementsBuilder.CreateIfStatement(IfStatementContext context)
   à TypeCobol.Compiler.Parser.CodeElementBuilder.EnterIfStatement(IfStatementContext context)
   à TypeCobol.Compiler.Parser.Generated.CodeElementsParser.IfStatementContext.EnterRule(IParseTreeListener listener)
   à Antlr4.Runtime.Tree.ParseTreeWalker.Walk(IParseTreeListener listener, IParseTree t)
   à Antlr4.Runtime.Tree.ParseTreeWalker.Walk(IParseTreeListener listener, IParseTree t)
   à TypeCobol.Compiler.Parser.CodeElementsParserStep.ParseProcessedTokensLinesChanges(TextSourceInfo textSourceInfo, ISearchableReadOnlyList`1 documentLines, IList`1 processedTokensLinesChanges, PrepareDocumentLineForUpdate prepareDocumentLineForUpdate, TypeCobolOptions compilerOptions, PerfStatsForParserInvocation perfStatsForParserInvocation)

```

The error comes from double paranthese  `((AY`

Issue #1067, opened 2018-07-26
Not defined variable used in a WHEN does not create an error 


collarbe:
Giving the following code : 

```COBOL
           evaluate true
             when nullInd < x'80'
               display "test"
           end-evaluate.
```

The parser does not create an error on the variable nullInd which is not defined.

Issue #1043, opened 2018-07-11
Remove Antlr for Node phase


smedilol:
CUPS seems to work fine for Node phase.
I think we can remove Antlr (backup) for this phase.

smedilol:
Commit linked to this issue:
665e0f6
06e7c38

PR #1093, opened 2018-08-13
v1.2.3

smedilol:
# What's New
- Support for `DECLARATIVES` inside procedure and fix some parsing/generation issues #655, #893, #1086 

- Command line interface now support mass compilation. You can pass programs that depends on each others. #1020 

- Code completion : 
   - Add character `.` after completing a type name in data division #1029 
      - Note : if you want to add the `value` clause after the type name, then you must move your cursor back. But for now the common case is to use typed variable without `value` clause.
   - Automatically add keywords `input` , `in-out` and `output` (if and when needed) when completing call to a procedure. #1030 

   - Suggest children variable event if their parent have no name (unnamed group variable for example) #982 
      - Example : With following declaration, completion after `Group` will propose `Var1` and `Var2`:
```cobol
01 Group.
    05 .
        10 Var1 pic X.
        10 Var2 pic X.
```

# What's fixed
- Variables not defined and used in a `WHEN` statement were not detected as error. #1067 
- Pointers use with syntaxes `set MyPointer up by x` or `down by x` were not generated correctly when declared as procedure's parameter #1071 
- Inside Dependency file : reference to a type declared in the same dependency doesn't need to be prefixed with the dependency program name anymore #723 
- Double parenthesis make parser crash #1078 

# Misc
- New tests to measure performance regression #442
- Minor performance improvement #1051 
- C# solution rules:
   - Activate rule "missing new or override keywords"  #951
   - Activate rule "invalid "is" / catch, assignemnt/comparison to same variable" #951
- Grammar simplification #586
   - Remove intermediate rule that doesn't match a real Cobol concept and only lead to another single rule
- Remove Antlr grammar files and related code for Node phase #1043
   - Only CUPS grammar and related code exists for this phase

- Add **partial** support for `global-storage section` #894 #805 
   - This new section allow to share data between a program and its procedures
   - You use this new section in your code, but it will stay "as is " in generated code
   - Complete support is expected for next release


smedilol:
Forgot to create this PR from a dedicated branch.
I'll create a branch v1.2.3 from develop and start again.

Issue #1011, opened 2018-06-27
Incremented pointer must be limited to 22 chars and not inside copy


smedilol:
**FOLLOWING #780**

2 rules to add:
If a pointer is incremented with a set up/down by instruction, then its generated name will be:
its name + a hash of 8 characters.

We must then limit its name to 22 chars because our target compiler (IBM V4 for now) limit variable name to 30 chars.

As codegen cannot change COPY, pointer inside copy cannot be incremented.

Brochato:
The 22 chars limit is respected in the redefines declaration generation (inside a type or not).
But its not in it usage, in the SET Statement generation.

PR #1104, opened 2018-08-20
Issue #1011, Add error caption on pointer name to long (22 char) or b…

maxime645:
…elonging to copy

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
@@ -867,6 +867,20 @@ namespace TypeCobol.Compiler.Diagnostics
                             break;
                         }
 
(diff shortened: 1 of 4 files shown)


TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -867,6 +867,20 @@ public static void CheckStatement(Node node)
                             break;
                         }
 
+                        if (receiver.Name.Length > 22)
+                        {
+                            DiagnosticUtils.AddError(node,
+                                "Pointer name '" + receiver.Name + "' is over 22 characters.");
+                            break;
```
smedilol: After reading your code and the code already in place, I think the `break` can be removed.
If a pointer variable is not declared correctly we want to have all error message at once.

So can you please remove the 3 `break` in this foreach ?

Issue #734, opened 2017-12-01
Make EUROINFO_LEGACY_REPLACING_SYNTAX an option of the compiler


smedilol:
One team that will use our parser will already have the copy suffixed with our special EUROINFO_LEGACY_REPLACING_SYNTAX.
However other EI rules could apply.

So we need to change our compiler directive EUROINFO_LEGACY_REPLACING_SYNTAX to a simple compiler options.

The CLI must evolve to accept this option.

To avoid to change all command lines, we can activate this option by default if EUROINFO_RULES is enabled.
Then team that user our parser will have to deactivate this option manually.

PR #1117, opened 2018-08-23
Issue #734, Made Euroinfo replacing syntax as a CLI option

maxime645:


TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -28,6 +29,38 @@ public class TypeCobolOptions : IBMCompilerOptions
...
         /// </summary>
         public bool UseAntlrProgramParsing { get; set; }
+        /// <summary>
+        /// Shall we use EUROINFO_LEGACY_REPLACING_SYNTAX snippets for Parsing the Program.
+        /// </summary>
+        public bool UseEuroInformationLegacyReplacingSyntax { get; set; } = true;
+
+#if EUROINFO_RULES
+#else
+        private bool _useEuroInformationLegacyReplacingSyntax;
```
smedilol: I think you meant UseEuroInformationLegacyReplacingSyntax = false ?

maxime645: It's a refactoring mistake. I fixed it in the latest push

Issue #1116, opened 2018-08-23
LS : GoToDefinition can locate a Token inside a Copy instead of the main document


smedilol:
**Describe the bug**
When a COPY contains a CodeElement which is located at the same line and position that an CodeElement in the source.
If you ask to Go To Definition on the CodeElement on the source, we can use the CodeElement inside the COPY instead of the one in the source.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZS0OSM.
       data division.
       working-storage section.
       COPY YFOO replacing ==:toto:== by ==toto==.
      
       procedure division.
           perform TOTO
           goback.
       TOTO.
           continue.
       END PROGRAM DVZS0OSM.
```
YFOO.cpy
```cobol
      ******************************************************************
       01  YFOO.
           05  var1 pic X.
           05  var1 pic X.
           05  var1 pic X.
           05  var1 pic X.
           05  var1 pic X.
           05  var1 pic X.
```
if `perform TOTO` is on line 8, then GoTODefinition doesn't works. On line 8 in YFOO there is `var1`
If it's on line 9, then it works. There is nothing on line 9 of YFOO.

**Technical**

See following code in TypeCobolServer.OnDefinition :
```C#
var matchingCodeElement =
                        fileCompiler.CompilationResultsForProgram.ProgramClassDocumentSnapshot.NodeCodeElementLinkers
                            .Keys.FirstOrDefault(c => c.ConsumedTokens.Any(
                                t => t.Line == parameters.position.line + 1 &&
                                     parameters.position.character >= t.StartIndex &&
                                     parameters.position.character <= t.StopIndex + 1));
```

PR #1118, opened 2018-08-24
Issue #1116, Fix GoTo Definition

maxime645:


smedilol:
It's ok, but can you add a test with LSR?
I can show you how to do it if you want.

Issue #1126, opened 2018-09-06
When END-PROGRAM is missing procedure are generated as nested instead of stacked


smedilol:
**Describe the bug**
When END-PROGRAM is missing procedure are generated as nested program instead of stacked program.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVSFOO.
       PROCEDURE DIVISION.
       declare procedure Foo.
       procedure division.
           display "Hey"
           .
       end-declare.
           .
    
```

Generated code is
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVSFOO.
       PROCEDURE DIVISION.
      *declare procedure Foo.
           .
      *
      *declare procedure Foo.
      *_________________________________________________________________
       IDENTIFICATION DIVISION.
       PROGRAM-ID. c1502275Foo.
       PROCEDURE DIVISION
           .
           display "Hey"
           .
       END PROGRAM c1502275Foo.
```

**Expected behavior**
Expected genrerated code:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVSFOO.
       PROCEDURE DIVISION.
      *declare procedure Foo.
           .
       END PROGRAM DVSFOO.
      *
      *declare procedure Foo.
      *_________________________________________________________________
       IDENTIFICATION DIVISION.
       PROGRAM-ID. c1502275Foo.
       PROCEDURE DIVISION
           .
           display "Hey"
           .
       END PROGRAM c1502275Foo.
```

We should generate an END PROGRAM during codegen, because a lot of program will not have this statement and it's not mandatory.
Also generate a warning if END PROGRAM is missing.

PR #1146, opened 2018-10-01
Issue #1126, warn for missing END PROGRAM, generate it if is missing while declaring a procedure

maxime645:


Codegen/src/Nodes/ProgramEnd.cs
```diff
@@ -17,15 +17,26 @@ internal class ProgramEnd: Compiler.Nodes.End, Generated {
         this.OriginalProcName = originalProcName;
 
 
+    }
+    public ProgramEnd(string originalProcName) : base(new Compiler.CodeElements.ProgramEnd())
```
smedilol: Semantic is not good here.
We use this new constructor to create an End Node for a program.
parameter should be called `programName`.

Codegen/src/Nodes/ProgramEnd.cs
```diff
@@ -17,15 +17,26 @@ internal class ProgramEnd: Compiler.Nodes.End, Generated {
...
 
-	private IList<ITextLine> _cache = null;
+        private IList<ITextLine> _cache = null;
 	public override IEnumerable<ITextLine> Lines {
 		get {
 			if (_cache == null) {
 				_cache = new List<ITextLine>();
-			    _cache.Add(new TextLineSnapshot(-1, "END PROGRAM " + ProgramHashName.Head + OriginalProcName + ".", null));
-			}
+                if(ProgramHashName.Head == "")
```
smedilol: If and else branch are almost identical, can you factorize?

Issue #1135, opened 2018-09-24
Don't crash if Kibana config is not set


smedilol:
**Describe the bug**
Currently if Kibana config is not found and telemetry is activated, TypeCobol doesn't work (or perhaps a subset of TypeCobol, like generation)

If Kibana config is not found, telemetry should be disabled.

PR #1161, opened 2018-10-10
Issue #1135, insert condition to use Analytics if settings are set

maxime645:


Analytics/ElasticTelemetry.cs
```diff
@@ -25,6 +25,13 @@ private ElasticTelemetry()
                 //Load custom app.config for this assembly
                 var appConfig = ConfigurationManager.OpenExeConfiguration(Assembly.GetExecutingAssembly().Location); 
 
+                if (appConfig.AppSettings.Settings["SmtpServer"].Value == "your.smtp.server")
```
smedilol: It will works if the configuration is still the default.
But I think we should also manage the case where the configuration file is wrong.
Like someone put  `http://BadServer.com` in `ElasticServer` property.

smedilol: As `SmtpServer` is not used by ElasticAnalytics, I'd prefer that this test use an `ElasticXXX` property.

Issue #1088, opened 2018-08-09
Codegen : global is not kept for type bool


smedilol:
```cobol
       01 varA type Bool global.
```

Global is lost during code generation.

Generated should be:
```cobol
      *01 varA type Bool global.
       01  varA-value PIC X VALUE LOW-VALUE global.
           88  varA       VALUE 'T'.
           88  varA-false VALUE 'F'
                           X'00' thru 'S'
                           'U' thru X'FF'.
```

PR #1164, opened 2018-10-15
Issue #1088, keep boolean "global" on generation

maxime645:


TypeCobol/Compiler/Nodes/Attributes.cs
```diff
@@ -205,6 +206,21 @@ public object GetValue(object o, SymbolTable table)
 	    }
     }
 
+    internal class GlobalAttribute : Attribute
+    {
+        public object GetValue(object o, SymbolTable table)
+        {
+            var data = o as DataDefinition;
+            if (data == null) return null;
+            if (((DataDescriptionEntry) data.CodeElement).Global != null &&
```
smedilol: Can you factorize the cast to DataDescriptionEntry ?

PR #1125, opened 2018-09-03
937 formalized comments

Brochato:
This Pull Request includes the folowing issues:
 [Formalized comment syntax for documentation generation.](https://github.com/TypeCobolTeam/TypeCobol/issues/998)
 [Formalized comment tasks.](https://github.com/TypeCobolTeam/TypeCobol/issues/1023)
 [Formalized comment Token and Scanner.](https://github.com/TypeCobolTeam/TypeCobol/issues/1012)
 [Formalized comment Code Element enhancement.](https://github.com/TypeCobolTeam/TypeCobol/issues/1022)
 [Free Multilines Comment.](https://github.com/TypeCobolTeam/TypeCobol/issues/1121)

PR #1175, opened 2018-10-30
Formalized Comments, Multilines Comments and Documentation Generation

Brochato:
This is a refactor of the branche 937_FormalizedComments
See PR #1125


TypeCobol/AntlrGrammar/CobolWords.g4
```diff
@@ -563,7 +563,24 @@ tokens
     PRIVATE,
     IN_OUT,
 	STRICT,
-	QuestionMark
+	QuestionMark,
```
smedilol: Token should be written in upper case.
I made the mistake with QuestionMark.

We should fix all tokens in this PR.

smedilol:
Issues referenced in commit are also wrong.

smedilol:
See new PR #1198

PR #1177, opened 2018-11-02
787 line exceed80 chars

maxime645:


TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -101,6 +101,13 @@ public static void ScanTokensLine(TokensLine tokensLine, MultilineScanState init
             if (textLine.Type == CobolTextLineType.Comment ||
                (textLine.Type == CobolTextLineType.Debug && !tokensLine.InitialScanState.WithDebuggingMode))
             {
+                if (tokensLine.ColumnsLayout == ColumnsLayout.CobolReferenceFormat && tokensLine.Text.Length > 80)
+                {
+                    if (tokensLine.Type == CobolTextLineType.Comment)
```
smedilol: I say that only comment should raise a warning when line exceed 80 characters.
But now that I see the code, it reminds me that Debug line when debug mode is not active could also raise a warning.

So you can remove this `if (tokensLine.Type == CobolTextLineType.Comment)` to make warning works both for Comment and inactive Debug lines.

smedilol:
@maxime645 I let you merge this PR

PR #1182, opened 2018-11-06
New Codegen Rebased

mayanje:


mayanje:
You are right maxime645, adding a comment like: 
//-------------------------------------------------------------------------------------------
//This file is automatically generated by TypeCobol.TemplateTranspiler from file Skeleton.xml.
//DO NOT EDIT.
//-------------------------------------------------------------------------------------------

mayanje:
There is one issue concerning the Skeleton.xml file, the one used by the current Codegen uses the character % as prefix character for meta environment variables. This character has an ambiguity as it can be the C# integer modulo operator %. This must be fixed. Otherwise it is possible to use only one file.

smedilol:
Messages of commits need to be rewritten as discussed together

PR #1215, opened 2018-12-21
1209 gen nested functions

maxime645:


TypeCobol/Compiler/Nodes/Attributes.cs
```diff
@@ -291,8 +293,46 @@ public object GetValue(object o, SymbolTable table)
...
+
+            return AnyNestedFunctions(pgm);
+        }
+
+        private bool AnyNestedFunctions(Program pgm)
+        {
+            if (pgm == null) return false;
 
+            if (pgm.Children.OfType<ProcedureDivision>().SelectMany(c => c.Children)
+                .Any(c => c is FunctionDeclaration && ((FunctionDeclaration)c).CodeElement().Visibility == AccessModifier.Public && ((FunctionDeclaration)c).IsNested))
```
smedilol: This test is duplicated at least 3 times.
Can you factorize?

maxime645:
In common case, all function declarations should be generated as Stacked Programs. 
An exception is made for function declarations that contain an EXEC statement. Those are generated as nested programs, since they cannot be used when they are in stacked programs.

When the option -f Cobol85Nested is used, every function declarations will be generated as nested programs.

maxime645:
The following rules are applied for both common and Cobol85Nested cases.
For public nested programs, the following rules are applied:
-	The program will be generated in the main program
-	The program is common, so it will be accessible from all other nested programs of the main program
-	Corresponding entry points will be generated as well, so the nested program will be accessible from other external programs
-	Corresponding number of variables needed to call the procedure will be created in the linkage section

A private nested program will be generated in its belonging program with no further code generation, since it should only be accessible by it.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -750,6 +750,21 @@ public virtual void StartExecStatement(ExecStatement execStmt)
 
         public virtual void EndExecStatement()
         {
+            //procedures containing EXEC statements should be generated as nested
```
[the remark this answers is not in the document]
maxime645: I saw this with @mayanje following the bug reported by @sarterbe.
The event EndExecStatement (therefore StartExecStatement) is only fired when the Exec statement is located in the procedure division and at the top of the instructions, while OnExecStatement is fired when the Exec statement is located in the procedure division and anywhere else but at the top of the instructions.
This behaviour might be a bug we need to fix. The two SetFlag are a quick fix to generate correctly the lines we need.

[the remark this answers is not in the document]
maxime645: The search for the FunctionDeclaration node is not to make sure we are in a FunctionDeclaration, but rather to set the flag on this declaration that contains an Exec Statement. The test on _InsideProcedure avoids to look for a parent of type FunctionDeclaration if the Exec statement is not into a procedure.
But maybe we could store the FunctionDeclaration node instead of "if we are inside a procedure division" ?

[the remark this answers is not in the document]
smedilol: Ok, so for the EndExecStatement  and OnExecStatement can you add this explanation directly in the source code?
Otherwise someone could remove the duplication to "optimize" the code.

Codegen/test/resources/output/TypeCobol/GenerateAsNested/FunDeclareWithExec-PrivateOnly.rdz.cbl
```diff
@@ -0,0 +1,147 @@
...
+            EXIT.
+
+
+      *
+      *declare procedure check private
+      *   input mydate        TYPE Date
+      *  .
+      *_________________________________________________________________
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. a0508f35check.
```
smedilol: This program need to be generated as COMMON.
Every program that is generated as nested either from a public or private procedure must be declared with COMMON.

Codegen/test/resources/output/TypeCobol/GenerateAsNested/FunDeclareWithExec-PrivateOnly.rdz.cbl
```diff
@@ -0,0 +1,147 @@
...
+           END-EXEC  
+           CONTINUE.
+       END PROGRAM a0508f35check.
+      *
+      *declare procedure check private
+      *   input mydate        TYPE Date
+      *   output r1           PIC X(5)
+      *  .
+      *_________________________________________________________________
+       IDENTIFICATION DIVISION.
```
smedilol: This program need to be generated as COMMON.

Codegen/test/resources/output/TypeCobol/GenerateAsNested/FunDeclareWithExec-PrivateOnly.rdz.cbl
```diff
@@ -0,0 +1,147 @@
...
+             WHERE LOT_SOC             = :ERIM04-LOT-SOC
+           END-EXEC  
+           CONTINUE.
+       END PROGRAM efd9419fcheck.
+      *
+      *declare procedure checkName private
+      *   input myname        PIC X(15)
+      *  .
+      *_________________________________________________________________
+       IDENTIFICATION DIVISION.
```
smedilol: This program need to be generated as COMMON.

Codegen/test/resources/output/TypeCobol/GenerateAsNested/FunDeclareWithExec-PublicPrivate.rdz.cbl
```diff
@@ -0,0 +1,201 @@
...
+           END-EXEC  
+           CONTINUE.
+       END PROGRAM efd9419fcheck.
+      *
+      *declare procedure checkName private
+      *   input myname        PIC X(15)
+      *  .
+      *_________________________________________________________________
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. a02a7aa5checkName.
```
smedilol: This program must be generated with COMMON.

smedilol:
@maxime645  Ok, for me.
As there are a lots of "fix" commit, can you do a rebased branch and create a new PR?

Thanks,

Issue #1223, opened 2019-01-08
Detect clause PIC with usage Comp-1/-2


maxime645:
A variable with a Picture clause cannot be used with usage COMP-1 or COMP-2.
Currently, the error is only detected after compilation. To avoid a useless compilation, it would be nice to fire it while writing the code.

Exemple of code that should fire an error:
```
01 MyVar PIC 9(5) COMP-1
01 MyVar PIC 9(5) COMP-2
```

PR #1224, opened 2019-01-08
Issue #1223, add error on combinated use of COMP-1/-2 and picture

maxime645:


TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -712,6 +712,14 @@ private static void CheckParameter([NotNull] ParameterDescriptionEntry parameter
                 CrossCompleteChecker.CheckPicture(node, parameter);
             }
 
+            if (parameter.Usage != null &&
```
smedilol: There is duplicate code here. Can you factorize ?

Issue #1226, opened 2019-01-10
Documentation generation


maxime645:
**Describe the bug**
The JSON file that is generated with option -f Documentation is not correctly generated.
The file lacks of the main collection that should contain all elements of type formalized comment.

Currently, after the file generation, we have : 
```
{
  "__type": "DocumentationForProgram:#TypeCobol.Compiler.Nodes",
  "Deprecated": "",
  "Description": "My program description ",
  "Name": "DocGen",
  "Namespace": "DocGen",
  "Needs": [ "some needs", "description" ],
  "ReplacedBy": "MyFonction2 ",
  "Restriction": "Do not Use BOOL var ",
  "See": "Thank you for your attention ",
  "ToDos": [ "Add BOOL support", "implement a call counter" ]
} {
  "__type": "DocumentationForType:#TypeCobol.Compiler.Nodes",
  "Deprecated": "",
  "Description": "inline typedef ",
  "Name": "myType",
  "Namespace": "DocGen",
  "Restriction": "Do not Use BOOL var ",
  "Visibility": 0,
  "DocDataType": {
    "MaxOccurence": 1,
    "Picture": "X(01)"
  }
}
```

The formalized comment should be generated in a collection like : 
```
{
  "MainCollection": [
    {
      "__type": "DocumentationForProgram:#TypeCobol.Compiler.Nodes",
      "Deprecated": "",
      "Description": "My program description ",
      "Name": "DocGen",
      "Namespace": "DocGen",
      "Needs": [ "some needs", "description" ],
      "ReplacedBy": "MyFonction2 ",
      "Restriction": "Do not Use BOOL var ",
      "See": "Thank you for your attention ",
      "ToDos": [ "Add BOOL support", "implement a call counter" ]
    },
    {
      "__type": "DocumentationForType:#TypeCobol.Compiler.Nodes",
      "Deprecated": "",
      "Description": "inline typedef ",
      "Name": "myType",
      "Namespace": "DocGen",
      "Restriction": "Do not Use BOOL var ",
      "Visibility": 0,
      "DocDataType": {
        "MaxOccurence": 1,
        "Picture": "X(01)"
      }
    }
  ]
}
```

PR #1228, opened 2019-01-11
Issue #1226, fix JSON generation

maxime645:


Codegen/src/Generators/DocumentationGenerator.cs
```diff
@@ -29,23 +31,70 @@ public void Generate(CompilationUnit compilationUnit, ColumnsLayout columns = Co
         {
             Destination.Append("");
             //Add version to output file
```
smedilol: Can you keep TypeCobol_Version ?

smedilol: Ok, I didn't see it.

Issue #1229, opened 2019-01-12
With MultiLine comments incremental scanning is broken


mayanje:
**FOLLOWING OF #1012**

The fact is that the Scanning with MultilineScanState in presence of multi line comments or formal comments is not taken in account by incremental scanning. This can produce bad incremental code element parsing.

As in, fact the incremental scanning was not broken, but not adapted to support Multi-line comments.
The support of Incremental Scanning in presence of MultiLine Comments is made possible by:

#1234 Bug for incremental mode and multi line changes
#1238 Make Incremental Scanning support Multi Line comments

PR #1230, opened 2019-01-12
WI #1229 Review of incremental scanning for multi line comments.

mayanje:


TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -490,6 +490,17 @@ public void UpdateTokensLines(System.Action onVersion = null)
                     currentTokensLinesVersion = currentTokensLinesVersion.next;
                     if (onVersion != null)
                         onVersion();
+
+                    //JCM: #1229 - So all incremental changes must have their code element reset to NULL to make Code Element Incremental Parsing invalidate them.
+                    foreach (var change in documentChanges)
```
smedilol: Is it specific to multiline comment?
If not, it would be better to have a separated issue to explain the problem. And then a separate commit.

TypeCobol/Compiler/Scanner/ScannerStep.cs
```diff
@@ -202,15 +204,53 @@ public static class ScannerStep
...
             MultilineScanState scanState = ScanTokensLineWithContinuations(lineToScanIndex, lineToScan, textSourceInfo, documentLines, prepareDocumentLineForUpdate, compilerOptions, copyTextNameVariations, tokensLinesChanges, initialScanState, out nextLineToScanIndex, out nextLineToScan);
 
             // Scan the following lines until we find that the scan state at the beginning of the next line has been updated
-            while (nextLineToScan != null && nextLineToScan.InitialScanState != null && !nextLineToScan.InitialScanState.Equals(scanState))
+            while (nextLineToScan != null && nextLineToScan.InitialScanState != null)
             {
-                scanState = ScanTokensLineWithContinuations(nextLineToScanIndex, nextLineToScan, textSourceInfo, documentLines, prepareDocumentLineForUpdate, compilerOptions, copyTextNameVariations, tokensLinesChanges, scanState, out nextLineToScanIndex, out nextLineToScan);
+                if (nextLineToScan.InitialScanState.Equals(scanState))
+                {//JCM: #1229 : We should break ==> but we must before check that tokens are not changing.
+                    TokensLine currentScanLine = nextLineToScan;
```
smedilol: Same question:
Is it specific to multiline comment?

mayanje: Ensuring that after an incremental change that the multilinescanstate is not changing is not enough, because a change can impact lines that were already in the desired multilinescansate but there tokens have been changed, so its true this situation appears often with multiline comments  that can hide syntactical source code, but not only.

smedilol: Ok.
For the new issue about incremental bug, can you also update documentation:
https://github.com/TypeCobolTeam/TypeCobol/wiki/TypeCobolIncrementalPipeline

smedilol:
See PR #1236 and a new PR should also be created for bug with multi line comment.

PR #1231, opened 2019-01-12
1225 syntax coloring

mayanje:


smedilol:
Ok, but we need to merge PR about WI #1229 Review of incremental scanning for multi line comments first

Issue #1225, opened 2019-01-10
Enhance TypeCobol LSP to support Syntax Coloring Custom Notification


mayanje:
The goal here is to implement a custom LSP notification in relation with the TypeCobol Incremental scanner to inform a client whenever a region of a document has been rescanned.
The notifcation parameters shall at least include the target region and the tokens concerned.

mayanje:
As the character for a multi line comment has changed, instead of using * now % is used, 
%<<, %>>, %<<<, %>>> must be explicitly marked as comment tokens.

PR #1242, opened 2019-01-23
WI #1225 Implement Custom Syntax Coloring LSP notification.

mayanje:
First implementation notify only MultiLine comments.


TypeCobol/LanguageServices/Editor/ILanguageServer.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// Are we supporting Syntax Coloring Notifications.    
+        /// </summary>
+        bool UseSyntaxColoring { get; set; }
+
+        /// <summary>
+        /// Called when a token scanning has been performed.
+        /// </summary>
+        /// <param name="changes">The list of document change instances, if this parameter is null then the whole document has been rescanned.</param>
+        /// <param name="compilationDocument">The underlying CompilationDocument instance</param>
+        void UpdateTokensLines(IList<DocumentChange<ITokensLine>> changes, CompilationDocument compilationDocument);
```
smedilol: As the first parameter can be null, I think we need a default value here.
```
 void UpdateTokensLines(CompilationDocument compilationDocument, IList<DocumentChange<ITokensLine>> changes = null);
```

TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -476,6 +482,8 @@ public void UpdateTokensLines(System.Action onVersion = null)
                 if (scanAllDocumentLines)
                 {
                     ScannerStep.ScanDocument(TextSourceInfo, compilationDocumentLines, CompilerOptions, CopyTextNamesVariations, initialScanStateForCopy);
+                    //Notify any Language Server listener that the whole document has been rescanned
+                    this.LanguageServer?.UpdateTokensLines(null, this);
```
smedilol: There is already a listener at the end of this method:
tokensLinesChanged
Can you use to avoid to change CompilationDocument?

mayanje: CanI call the listener if the entire document has been changed, by passing null as changing?

smedilol: Can you create a second generic listerner in this case ?
I think it's better to have a generic behavior that we can reuse for different purpose.

mayanje: Yes I can.

smedilol: Or yes maybe passing null or create a DocumentChangedEvent<ITokensLine>

TypeCobol/LanguageServices/Editor/ILanguageServer.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TypeCobol.Compiler;
+using TypeCobol.Compiler.Concurrency;
+using TypeCobol.Compiler.Scanner;
+using TypeCobol.Compiler.Text;
+
+namespace TypeCobol.LanguageServices.Editor
```
smedilol: Now that we have a generic listener in CompilationDocument can you move this class to the TypeCobol.LanguageServer project ?

TypeCobol/Compiler/FileCompiler.cs
```diff
@@ -79,6 +80,33 @@ public class FileCompiler
 
         public EventHandler<ExecutionStepEventArgs> ExecutionStepEventHandler { get; set; }
 
+        private ILanguageServer _LanguageServer;
+        /// <summary>
+        /// Any connection to a Language Server instance.
+        /// </summary>
+        public ILanguageServer LanguageServer
```
smedilol: A property "LanguageServer" shouldn't be in TypeCobol project.
Can you move it to TypeCobol.LanguageServer project ?

mayanje: It's done,  with code refactoring.

TypeCobol/Compiler/FileCompiler.cs
```diff
@@ -12,6 +12,7 @@
 using TypeCobol.Compiler.Scanner;
 using TypeCobol.Compiler.Text;
 using TypeCobol.Tools;
+using TypeCobol.LanguageServices.Editor;
```
smedilol: There is one using left, I think you can remove it.

mayanje:
Yes I will merge will "Squash and merge" as soon as appvoyer checks pass

Issue #1245, opened 2019-01-24
Update CSharp version


maxime645:
All the projects in the solution are currently running with C#6. We want to upgrade to CSharp 7.3.

PR #1246, opened 2019-01-25
Issue #1245 Update CSharp version to 7.3

maxime645:


CSCupRuntime/CSCupRuntime.csproj
```diff
@@ -17,21 +16,27 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <OutputPath>..\bin\Debug\</OutputPath>
+    <LangVersion>7.3</LangVersion>
```
smedilol: I I think you can move LangVersion to file ..\TypeCobol\Common.props
This will avoid to change all csproj files.

smedilol: Maybe not all csproj use the common.props file

maxime645: Right, I forgot about the common.props file. All done

TypeCobol/Common.props
```diff
@@ -42,6 +42,7 @@
     <WarningLevel>4</WarningLevel>
 	<CodeAnalysisRuleSet>TypeCobol.ruleset</CodeAnalysisRuleSet>
     <RunCodeAnalysis>true</RunCodeAnalysis>
+    <LangVersion>7.3</LangVersion>
```
smedilol: Can you put the LangVersion at the top of this file, so it applies for all configuration ?

Issue #1243, opened 2019-01-24
Continuation line works when outside literal


smedilol:
**Describe the bug**
The continuation line character (`-` in column 7) is allowed in any statement.
- It can be used to continue a literal on multiple lines
- To split a statement in multiple lines even if it's not necessary in this case

Our parser treat the second usage as an error.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       data division.
       working-storage section.
       01 foo pic X(50).
       PROCEDURE DIVISION.
      
             move 'CHAMPS 082 A 085 INTERDITS EN OUVERTURE'
      -        to foo
      
           goback
           .
       END PROGRAM DVZF0OSM.
```
There is an error on the move statement.

**Expected behavior**
There should be no error on this program.

PR #1249, opened 2019-01-28
Issue #1243 Remove diagnostic: first char after continuation line should be a quotation mark

maxime645:


smedilol:
Tests are related to the scanner.
Can you add the of the issue as a parser test?

Issue #1247, opened 2019-01-25
Kibana is still to slow


smedilol:
**FOLOWING OF #1131**
When use by the command line, Analytics done with Kibana is still too slow.
For LanguageServer it seems to be ok.
As we can scan our sources code to find the kind of TypeCobol syntax used, we can remove these events from Kibana:
- FunctionDeclared
- TypeDeclared
- FunctionCalled
- TypeUsed

But do we keep them for LanguageServer ?

The only events left for command line would be:
- Duration
- ReturnCode
- Exceptions
- Diagnostic

For LanguageServer :
- Duration
- ReturnCode
- Exceptions
- Diagnostic
- Go to definition
- Completion
- SignatureHelper

smedilol:
After discussion, remove 

- FunctionDeclared
- TypeDeclared
- FunctionCalled
- TypeUsed

For CLI and LanguageServer

PR #1280, opened 2019-02-12
WI #1247 Remove useless track of events

maxime645:
On a source that used to take 30 seconds to generate, we win 5 to 8 seconds


mayanje:
Rather than remove it totally, maybe we can use a C# Directive like:
#if USE_TELEMETRY
......
#endif

smedilol:
No, it'll just be dead code that is never gonna be activated.
If we want to revert the event, we can revert the commit.

PR #1299, opened 2019-03-06
1031 outline rebase

maxime645:


maxime645:
See rebased version #1416

Issue #1323, opened 2019-03-22
Make codegen work with Typedef inside COPY


mayanje:
With issue #1019 it is now possible to declare a TYPEDEF whose body is inside a COPY.
However TypeCobol does not support TYPEDEF entirely declared inside a Copy.
But Codegen shall works as if it is possible. But a semantic error must be generated.

maxime645:
The entirety of a TypeDef shouldn’t be in a copy. 
The declaration line (01 myType TypeDef) should in any case be in the TypeCobol Source. Only the bodies of weak and strong TypeDefs can be stored in a copy.
Using a TypeDef strict should be forbidden.

smedilol:
TypeCobol is abandonned, this issue is not needed anymore.

PR #1324, opened 2019-03-22
WI #1323 Adapt the generator to variable having a Typedef in a copy

mayanje:


Codegen/test/resources/output/TypeCobol/TypedefCopys/TypedefInsideCopy.rdz.tcbl
```diff
@@ -0,0 +1,20 @@
...
+       configuration section.
+       source-computer. ibm-370
+      *                 with debugging mode
+                        .
+       special-names. decimal-point is comma.
+       data division.
+       working-storage section.
+     
+
+       COPY MyCopyC.
```
smedilol: As the COPY contains a Typedef, this line must be commented.

But you have chose a good test case.
Your COPY also contains data declaration outside the typedef.
So I'd say that this case must be refused. Codegen cannot comment inside the COY.


If we want to generated the code, then :
Typedef must be allowed in COPY if there are only typedef in the copy file.
Otherwise there must be an error (you cannot mix data declaration and typedef)


Note that for Cobol 2002 and if you don't want to generate code to Cobol 85, then you can mix data declaration and typedef in a COPY.

smedilol: We need to discuss this case at our next project meeting.

Issue #662, opened 2017-09-06
When LanguageServer starts, it should returns TypeCobol version


smedilol:
To ease debugging, it should be nice if TypeCobol version was returned by TypeCobolServer when it starts.
IDZ could then display it.

collarbe:
I see two options for this issue. 

- First, is to add TypeCobol version in InitializeResult object returned by LSP after the initialization. 
- Second, create a custom method in TypeCobol LSP that returns TypeCobol's version.

PR #1349, opened 2019-04-10
WI #662 Add Experimental Properties as server capabilities.

mayanje:


TypeCobol.LanguageServer/VsCodeProtocol/Initialize Method/ServerCapabilities.cs
```diff
@@ -85,5 +85,10 @@ public class ServerCapabilities
         /// The server provides rename support.
         /// </summary>
         public bool renameProvider { get; set; }
+
+        /// <summary>
+        /// Experimental value.
+        /// </summary>
+        public object experimental;
```
smedilol: Why using the word `experimental` ?

mayanje: Because it is possible to use this field, without having to create a custom LSP notification. So we can get all additional capabilities information from the server as a result of the initialize request, thus we stay compatible with LSP specification.

PR #1359, opened 2019-04-26
1358 codegen line map

mayanje:


TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -24,6 +24,7 @@ public class TypeCobolConfiguration
         public List<string> CopyFolders = new List<string>();
         public List<string> InputFiles = new List<string>();
         public List<string> OutputFiles = new List<string>();
+        public List<string> LineMapFiles = new List<string>();
```
smedilol: Do we need a list of lineMap files? I mean, an unique file should be enough I think.

mayanje: I was thinking about that, but it seems that, but this is the current implementation for mass generation if needed, the fact is what happend if several input file are used, for what file to generate the mapping file ?

smedilol: Yes, right.

Issue #1363, opened 2019-05-09
CLI : Accept parameter "glm|genlinemap=" but do nothing


smedilol:
Our current production version is 1.2.6.
Our next qualification version wll be 1.3.0.

1.3.0 accept a new parameter : "glm|genlinemap="
Our IDE client can switch between 1.2.6 and 1.3.0 but it doesn't which version will accept parameters "glm|genlinemap=".
So the client will always pass this parameter.

Our production version must also accept this parameter.
The goal is to create a version 1.2.7 which accept parameter "glm|genlinemap=" but does nothing with it for now.

PR #1364, opened 2019-05-09
WI #1363 Accept -glm option but do nothing with it

maxime645:


maxime645:
This release prepares the production version for the next qualification release
# New

- Add recognition of option -glm. This option is not used until v1.3 (#1363)

Issue #1373, opened 2019-05-20
Error in diagnostic list when too many warnings


maxime645:
**Describe the bug**
Due to the low "maximum number of diagnostics" default preference, an error can be hidden if there is too many warnings.
The server is sending the 30 first diagnostics regardless of the severity.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOZBMX0.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 myvar pic X.
      
       PROCEDURE DIVISION.
      
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
           evaluate myvar
             when 'a'
                display 'hello'
           end-evaluate.
      
      
           move 'error' to var.
      
      
       END PROGRAM TCOZBMX0.
```
The maximum number of diagnostic is set to 1.
No error is displayed. Remove the evaluate, the error will appear.

**Expected behavior**
The server should return the error diagnostics in priority.

maxime645:
After discussion, the default value for the maximum number of diagnostics should be augmented to 200

PR #1374, opened 2019-05-20
WI #1373 Prioritize errors over warnings on returned diagnostics to t…

maxime645:
…he client


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -587,10 +587,23 @@ private void RefreshCustomSymbols()
         private void ProgramClassChanged(object cUnit, ProgramClassEvent programEvent)
         {
             var compilationUnit = cUnit as CompilationUnit;
-            var fileUri = OpenedDocumentContext.Keys.FirstOrDefault(k => compilationUnit != null && k.LocalPath.Contains(compilationUnit.TextSourceInfo.Name));
+            var fileUri = OpenedDocumentContext.Keys.FirstOrDefault(k =>
+                compilationUnit != null && k.LocalPath.Contains(compilationUnit.TextSourceInfo.Name));
 
-            var diags = compilationUnit?.AllDiagnostics().Take(TypeCobolConfiguration.MaximumDiagnostics == 0 ? 100 : TypeCobolConfiguration.MaximumDiagnostics);
-            DiagnosticsEvent(fileUri, new DiagnosticEvent() { Diagnostics = diags});
+            var test = compilationUnit?.AllDiagnostics().GroupBy(d => d.Info.Severity);
```
smedilol: test ?
It's also a duplicate line of 
` var severityGroups = compilationUnit?.AllDiagnostics().GroupBy(d => d.Info.Severity);`

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -587,10 +587,23 @@ private void RefreshCustomSymbols()
...
+            System.Collections.Generic.IEnumerable<Diagnostic> diags = new List<Diagnostic>();
+
+            var severityGroups = compilationUnit?.AllDiagnostics().GroupBy(d => d.Info.Severity);
```
smedilol: OrderBy doesn't work directly?

TypeCobol.LanguageServer/Workspace.cs
maxime645: You are right, I forgot to remove that line before commit

TypeCobol.LanguageServer/Workspace.cs
maxime645: OrderBy would work but the errors would be at the end of the list. We would need to reverse that list and change the order of appearance of the diagnostics. Since many of our LSR tests are already handling warnings and errors, they would be false.
The only way to change the priority of the errors without doing to much changes in the LSR test is to move up the errors as a group. In this way, the order of the diagnostics won't change within the groups.

smedilol: Ok, can you add this comment on the source ?

PR #1381, opened 2019-06-05
1204 remove server mode in cli

fm-117:


fm-117:
I have restored the -1 flag in the OptionSet with empty delegate asssociated.

PR #1384, opened 2019-06-07
1343 forbid typedef in copy

fm-117:


TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
```diff
@@ -93,13 +93,28 @@ public static void CheckTypeDefinition(TypeDefinition typeDefinition)
                 string message = "TYPEDEF \'" + typeDefinition.Name + "\' has no description.";
                 DiagnosticUtils.AddError(typeDefinition, message, MessageCode.SemanticTCErrorInParser);
             }
+
+            if (typeDefinition.IsInsideCopy())
```
smedilol: It's ok but for performance reason, this test should be called only during the CodeElement phase.
Method CheckTypeDefinition(TypeDefinition typeDefinition) is called on crossCheck phase, meaning that this test apply to all TypeDefinition.

If this test is moved to CodeElement phase, it'll only be called if the DataTypeDescriptionEntry (the CodeElement associated to the TypeDefinition) need to be refreshed.

So can you can create another static method in class TypeDefinitionChecker. This new method can be called  by CobolCodeElementBuilder.

fm-117: It was planned to include this check in TypeDefinitionEntryChecker.CheckTypedef but the IsInsideCopy method always returns false in this context because no tokens have been consumed yet. The location is not available at this phase.

smedilol: Ok right I forgot about that.

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -173,6 +173,7 @@ public static Nodes.TypeDefinition CreateBuiltIn(DataType type)
 		}
 		private static Nodes.TypeDefinition CreateBase(DataType type) {
 			var entry = new DataTypeDescriptionEntry();
+		    entry.Visibility = AccessModifier.Public;
```
maxime645: Both Visibilities can be set in the method CreateBuiltIn

fm-117: See 924ed27d5967cdb01052e052ad645d8c7f7dec8a

PR #1402, opened 2019-06-24
1399 new statement free cobol6.1

fm-117:
Also includes codegen test for INITIALIZE and OCCURS UNBOUNDED syntax.


TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -189,6 +189,11 @@ codeElement:
 	returnStatement |
 	returnStatementEnd |
 	sortStatement |
+
+//	 [COBOLV6]
+	// --- Dynamic allocation statements ---
+	freeStatement |
```
smedilol: I'll rather a generic name here, like `cobol2002Statement`.
Then redefines it in Cobol2002CodeElements.g4 to include the `freeStatement`

fm-117: See ab10d10110b03c9420497858c93592d7498b6a42.

TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -204,14 +208,21 @@ codeElement:
...
 
 // what is here is not important as long as it is not epsilon
 tcCodeElement: PUBLIC | PRIVATE | UNSAFE;
 //	[/TYPECOBOL]
 
 
+//	[COBOL2002]
+// dummy definition of Cobol 2002 or V6 code elements, not used but required by Antlr.
+// see Cobol2002CodeElements.g4 for the proper definition.
+cobol2002Statement: UNSAFE;
```
mayanje: Ok dummy production has been created in order to not have many changes to be handled in CodeElementBuilder.
But at the first glance, this grammar in fact ambigüe even if theses are error cases. because UNSAFE is used as first token in both rules: tcCodeElement and cobol2002Statement, I think in this situation the first rule will match. 
So I think If this is the choice it's better than to use a distinct dummy token for each grammar extension like:
tcCodeElement: TC_EXTENSIONS;
cobol2002Statement: CBL2002_EXTENSIONS;

And also a token which will never be returned by the Scanner, because UNSAFE can be returned.

fm-117: Ok to include specific dummy tokens in next change (ALLOCATE support) but not directly in this PR as it is easier to keep the TokenList as it is.

Codegen/test/resources/input/CobolV6/DynamicAllocationStatements.rdz.cbl
```diff
@@ -0,0 +1,22 @@
...
+000000 01 var4.
+000000   05 var5 POINTER.
+000000 01 var6.
+000000   05 var7 POINTER.
+000000
+000000 PROCEDURE DIVISION.
+000000     FREE var0
+000000     FREE var1 var2 var3
+000000     FREE var5 OF var4
+000000     FREE var7 IN var6
```
mayanje: It will be nice to also use a TypeCobol feature like: FREE var4::var5 and FREE var6::var7

fm-117: See 985dc0d39396202ff2509ab5aa5e3abe930a0e55.

PR #1405, opened 2019-07-05
1398 new statement allocate cobol6.1

fm-117:


TypeCobol/Compiler/CodeElements/Statement/AllocateStatement.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Must be a level-01 or level-77 item defined in the LINKAGE SECTION.
+        /// If data-name-1 is specified, the RETURNING phrase can be omitted.
+        /// Otherwise, the RETURNING phrase must be specified.
+        /// Cannot be reference modified.
+        /// Cannot be a group item that contains an unbounded table.
+        /// <remarks>Mutually exclusive with <see cref="AllocatedSize"/></remarks>
+        /// </summary>
+        public SymbolReference AllocatedArea { get; set; }
```
smedilol: AllocatedArea must be of type Variable.

TypeCobol/AntlrGrammar/Cobol2002CodeElements.g4
```diff
@@ -44,6 +45,13 @@ initializeStatement:
 	(THEN? REPLACING initializeReplacingDirective+)?
 	(THEN? TO? DEFAULT)?;
 
+// New Cobol v6.1 ALLOCATE statement to obtain dynamic storage.
+// 'INITIALIZED' is defined here as a contextual keyword. The dataNameReference is therefore not allowed to be named 'INITIALIZED' in this statement.
+allocateStatement:
+	ALLOCATE ((arithmeticExpression CHARACTERS) | { !string.Equals(CurrentToken.Text, "INITIALIZED", System.StringComparison.InvariantCultureIgnoreCase) }? dataNameReference)
+	({ string.Equals(CurrentToken.Text, "INITIALIZED", System.StringComparison.InvariantCultureIgnoreCase) }? KeywordINITIALIZED=UserDefinedWord)?
+	(RETURNING dataItemReference)?;
```
smedilol: Use variable2 here.
The difference with dataItemReference is that it's mapped to Variable class.

Then use CoboLExpressionsBuilder.CreateVariable to instantiate Variable.

TypeCobol/AntlrGrammar/Cobol2002CodeElements.g4
```diff
@@ -44,6 +45,13 @@ initializeStatement:
 	(THEN? REPLACING initializeReplacingDirective+)?
 	(THEN? TO? DEFAULT)?;
 
+// New Cobol v6.1 ALLOCATE statement to obtain dynamic storage.
+// 'INITIALIZED' is defined here as a contextual keyword. The dataNameReference is therefore not allowed to be named 'INITIALIZED' in this statement.
+allocateStatement:
+	ALLOCATE ((arithmeticExpression CHARACTERS) | { !string.Equals(CurrentToken.Text, "INITIALIZED", System.StringComparison.InvariantCultureIgnoreCase) }? dataNameReference)
```
smedilol: Use storageArea2 instead of dataNameReference.
storageArea2 is mapped to ReceivingStoreageArea.

Then use CoboLExpressionsBuilder.CreateStorageArea to instantiate ReceivingStorageArea.

fm-117:
We'll use storageArea2 for dataNameReference as it ias already available and we'll create pointerStorageArea along with correponding CreateStorageArea method in CobolExpressionsBuilder. StorageDataType will be set to DataPointer, a new value to be created in the enum.

fm-117:
See 46e601a4cf8a54a195cd8319dd6eb3c157e5560d for the AllocateStatement.
See 698d2e1514be81be2e0777182142b7b0c5a33407, same thing for the FreeStatement.

TypeCobol/Compiler/CodeElements/Statement/AllocateStatement.cs
fm-117: See comment below.

TypeCobol/AntlrGrammar/Cobol2002CodeElements.g4
```diff
@@ -44,6 +45,13 @@ initializeStatement:
 	(THEN? REPLACING initializeReplacingDirective+)?
 	(THEN? TO? DEFAULT)?;
 
+// New Cobol v6.1 ALLOCATE statement to obtain dynamic storage.
+// 'INITIALIZED' is defined here as a contextual keyword. The storageArea2 is therefore not allowed to be named 'INITIALIZED' in this statement.
+allocateStatement:
+	ALLOCATE ((arithmeticExpression CHARACTERS) | { !string.Equals(CurrentToken.Text, "INITIALIZED", System.StringComparison.InvariantCultureIgnoreCase) }? storageArea2)
```
smedilol: It doesn't works like that ? 
```
	ALLOCATE ((arithmeticExpression CHARACTERS) | storageArea2)
	({ string.Equals(CurrentToken.Text, "INITIALIZED", System.StringComparison.InvariantCultureIgnoreCase) }? KeywordINITIALIZED=UserDefinedWord)?
	(RETURNING pointerStorageArea)?;
```

fm-117: It does but it will successfully parse 'ALLOCATE INITIALIZED' statement which is incorrect. storageArea2 is not allowed to be named 'INITIALIZED' in this context as we cannot distinguish between the data name and the optional keyword.
The COBOL compiler rejects 'INITIALIZED' as an identifier so it seems more robust to forbid it in this case.

Issue #1331, opened 2019-03-29
Codegen : option to keep/remove commented TypeCobol intrusction


smedilol:
**Is your feature request related to a problem? Please describe.**
Base problem is that our generated code is very big compare to the TypeCobol source file.
One solution, is to reduce unuseful comment in the generated code.

Currently our generated code, keep the following TypeCobol instructions as comment :
- Typedef declaration
- operator `::`
- Function/procedure declaration
- set boolean to xxx
- set qualified index with `::` to xxx
- ...

Our users don't look at the generated code anymore, so these commented TypeCobol instructions are not useful.
But we want to be able to generate code with all these lines so TypeCobol beginners can understand the code.

**Describe the solution you'd like**
By default, don't comment any TypeCobol instructions in the generated code. But remove them completely.

The formalized comment of function/procedure can still be kept above the generated stacked program.

Create an TypeCobol.CLI option so we can keep all these instructions commented (same behavior as today).

## Note from the code review of PR #1410 
If the original TypeCobol call to procedure is removed, then we must improve the generated commment.
This improvment in the generated code can be done regardless of the TypeCobol.CLI option to keep all TypeCobol instructions as comment.

Just below this kind of comment:
```cobol
      *    Equivalent to call ebb060ea in module Callee
```
Add a comment to indicate which TypeCobol procedure is called

Which give the following result:
```cobol
      *    Equivalent to call ebb060ea in module Callee
      *     Call TypeCobolModule::ProcedureName
```

smedilol:
Following Code review https://github.com/TypeCobolTeam/TypeCobol/pull/1410, we could improve the generated code for call procedure.
Indicate which TypeCobol procedure is called by a standard Cobol call.

smedilol:
No longer needed as TypeCobol is abandonned.

PR #1410, opened 2019-07-09
1233 Generate programs with 8-chars-long names (hash)

fm-117:


CLI/test/ressources/dependencies_1/output_expected/ProcedureCall-Public.rdz.tcbl
```diff
@@ -90,7 +90,7 @@
              OR TC-Callee-ebb060ea-Idt not = 'ebb060ea'
                PERFORM TC-LOAD-POINTERS-Callee
            END-IF
-      *    Equivalent to call ebb060eaMyPublicProcedure in module Callee
+      *    Equivalent to call ebb060ea in module Callee
```
maxime645: I think we should add a comment to clarify that the program ebb060ea is the generated code for function check here as well

smedilol: We have the original TypeCobol call commented above.
We can add this in #1331 which will remove the original TypeCobol call.

fm-117: Or simply replace the comment by 'Equivalent to call <TC-procedure-name> in module <module-name>', that is not mentioning the generated code in comment but only refering to TC names ?

smedilol: This comment is useful when you debug a TypeCobol program, because our tools will only show the Cobol program name (here : ebb060ea) and not the procedure name.

Issue #1360, opened 2019-04-30
Fix global name resolution


smedilol:
**Describe the bug**
The current resolution for symbol marked as global is not correct.

Here's the rule from the IBM Cobol v6 Enterprise:

> Names within programs
> When a program, program B, is directly contained within another program,
> program A, both programs can define a condition-name, a data-name, a file-name,
> or a record-name using the same user-defined word. When such a duplicated name
> is referenced in program B, the following steps determine the referenced resource
> (these rules also apply to classes and contained methods):
> 1. The referenced resource is identified from the set of all names that are defined
> in program B and all global names defined in program A and in any programs
> that directly or indirectly contain program A. The normal rules for qualification
> and any other rules for uniqueness of reference are applied to this set of names
> until one or more resources is identified.
> 2. If only one resource is identified, it is the referenced resource.
> 3. If more than one resource is identified, no more than one resource can have a
> name local to program B. If zero or one of the resources has a name local to
> program B, the following rules apply:
> v If the name is declared in program B, the resource in program B is the
> referenced resource.
> 66 Enterprise COBOL for z/OS, V6.1 Language Reference
> v If the name is not declared in program B, the referenced resource is:
> – The resource in program A if the name is declared in program A
> – The resource in the containing program if the name is declared in the
> program that contains program A
> This rule is applied to further containing programs until a valid resource is
> found.

But currently we never stop the symbol resolution when we found a matching name. We continue to resolve the variable in containing program.

**To Reproduce**
In the following example, there should be no ambiguity of symbol `a`.
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZFSCT0.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 a pic X global.
      
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Nested.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 a pic X.
       01 b pic X.
       procedure division.
           move a to b
           .
       END PROGRAM Nested.
       END PROGRAM DVZFSCT0.
```

This also apply to Types.
In the following example, there should be no ambiguity of symbol `a`.
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZFSCT0.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 a typedef pic X global.
      
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Nested.
[...]

PR #1415, opened 2019-07-16
WI #1360 Quit searching when node found but always search in GS table

fm-117:


fm-117:
Abandoned, need more testing.

PR #1435, opened 2019-08-20
1430 ls reload copy

maxime645:


TypeCobol.LanguageServer/FileWatchers/CopyWatcher.cs
```diff
@@ -0,0 +1,73 @@
...
+            var directory= new FileInfo(e.FullPath).Directory;
+            if (File.Exists(directory.FullName + Path.DirectorySeparatorChar + "~.lock"))
+                return;
+
+            _TypeCobolWorkSpace.CompilationProject.ClearImportedCompilationDocumentsCache();
+
+            Action refreshAction = () => { _TypeCobolWorkSpace.RefreshOpenedFiles(); };
+            lock (_TypeCobolWorkSpace.MessagesActionsQueue)
+            {
+                if (_TypeCobolWorkSpace.MessagesActionsQueue.All(mw => mw.Action != refreshAction))
```
smedilol: The comparison : `mw.Action != refreshAction` doesn't works because `refreshAction` is instanciated each time  `OnChanged` method is called.

As there also a refreshAction created in DependenciesFileWatcher class, can you ensure that there is only one  refreshAction either from DependenciesFileWatcher or CopyWatcher in the MessagesActionsQueue ?

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -439,6 +441,11 @@ public void DidChangeConfigurationParams(IEnumerable<string> arguments)
 
             //Dispose previous watcher before setting new ones
             _DepWatcher.Dispose();
+            _CopyWatcher.Dispose();
```
fm-117: It is unusual to Dispose object before using it. Maybe the method should be renamed Reset and separated from the .NET Dispose pattern ?

TypeCobol.LanguageServer/FileWatchers/CopyWatcher.cs
maxime645: It's done

PR #1437, opened 2019-08-21
791 preserve value clause in proc params and typed variables

fm-117:


CLI/test/ressources/value_clauses/input/ValueClauses.rdz.tcbl
```diff
@@ -0,0 +1,34 @@
...
+      
+       01 StructureType typedef strict PRIVATE.
+          05 champ PIC X(1).
+      
+       01 ScalarType typedef strict PRIVATE PIC X(1).
+      
+       01 C-DatePivot   TYPE Date VALUE '20190101'.
+       01 C-DefaultRate TYPE Rate VALUE 3.55.
+      
+       01 Structure TYPE StructureType     VALUE 'A'.
```
smedilol: Can you use `01 Structure VALUE 'A' TYPE StructureType.` ?
The goal is to check that the generated line doesn't include `TYPE StructureType`

Codegen/src/Nodes/TypedDataNode.cs
```diff
@@ -200,6 +205,48 @@ internal static string ExtractAnyCobolScalarTypeDef(ColumnsLayout? layout, TypeD
...
                     }
                     FlushConsumedTokens(layout, i - 1, customtype.CodeElement.ConsumedTokens, sb, out bHasPeriod);
+
+                    //Add any VALUE clause from dataDescriptionEntry
+                    if (bHasPeriod)
+                    {
+                        sb.Length--; //Trim trailing PeriodSeparator if any.
+                        bHasPeriod = false;
+                    }
+                    i = 0;
```
mayanje: Why to continue here if bHasPeriod is true?

fm-117: At this point we have the type info from the typedef but we still need to add the VALUE clause from the dataDef. It is also possible to use a tokenFilter to avoid outputting the PeriodSeparator bu I found it easier to simply remove it.

mayanje: But I mean if we have encountered a dot in any declaration, there is nothing else that can follow it.

mayanje: So I understand now that, first the PICTURE type tokens from the TypeDef are flushed followed by the any VALUE tokens.

fm-117:
See rebase version : #1466.

Issue #1443, opened 2019-08-29
TC Report Test failures are not detected in case of diagnostics.


mayanje:
The method ParseWithNodeListenerReportCompare in the file
TypeCobol.Test\Report\ReportTestHelper.cs

does not signal that the report is not generated because the call to the method  parser.Results.AllDiagnostics() returns diagnostics.

So this method should now return true or false, so that Tests can check a TRUE assertion value.

Because for instance now the test "TestReportCall" is no longer working because of the following diagnostic: _Warning: "when other" is missing_

PR #1444, opened 2019-08-29
WI #1443 Add boolean retun value.

mayanje:


TypeCobol.Test/Report/ReportTestHelper.cs
```diff
@@ -73,8 +74,17 @@ public static void ParseWithNodeListenerReportCompare(string fileName, string re
...
                         }
                     }
+                    else
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    return false;
```
fm-117: The two blocks `else { return false; }` can be simplified as `return false;`at the end of the try block.

TypeCobol.Test/Report/ReportTestHelper.cs
```diff
@@ -73,8 +74,17 @@ public static void ParseWithNodeListenerReportCompare(string fileName, string re
                             string result = sw.ToString();
                             string expected = File.ReadAllText(output, format.Encoding);
                             TypeCobol.Test.TestUtils.compareLines(input, result, expected, PlatformUtils.GetPathForProjectFile(output));
+                            return true;
                         }
                     }
+                    else
+                    {
+                        return false;
```
maxime645: Can you return more information than just false in the case if there are diagnostics or the report is null ?
Running the test in debug just to know what happened can be prevented if the information is in the output.

mayanje: Do you want to return the Diagsnotics list?

TypeCobol.Test/Report/ReportTestHelper.cs
mayanje: Just a matter of style

TypeCobol.Test/Report/ReportTestHelper.cs
maxime645: Not necessarily the Diagnostics list, but at least a message explaining why the test failed (e.g. "diagnostics encountered", "report not generated correctly"...)

mayanje: Is a return code enough?

PR #1461, opened 2019-09-05
1386 on hover display variable definition

fm-117:


maxime645:
There are some missing cases on qualified names, indexes, conditions and call parameters

fm-117:
See #1503

PR #1463, opened 2019-09-06
1427 dependencies copy

maxime645:


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -527,7 +527,16 @@ private void RefreshCustomSymbols()
...
-                _customSymbols = Tools.APIHelpers.Helpers.LoadDependencies(TypeCobolConfiguration.Dependencies, TypeCobolConfiguration.Format, _customSymbols, TypeCobolConfiguration.InputFiles, TypeCobolConfiguration.CopyFolders, DiagnosticsErrorEvent); //Refresh Dependencies
+                _customSymbols = Tools.APIHelpers.Helpers.LoadDependencies(TypeCobolConfiguration.Dependencies, TypeCobolConfiguration.Format, _customSymbols, TypeCobolConfiguration.InputFiles, 
+                    TypeCobolConfiguration.CopyFolders, DiagnosticsErrorEvent, out List<RemarksDirective.TextNameVariation> usedCopies, out List<CopyDirective> missingCopies); //Refresh Dependencies
+
+                if(missingCopies.Count > 0)
+                {
+                    MissingCopiesEvent(TypeCobolConfiguration.Dependencies?.FirstOrDefault(), new MissingCopiesEvent() { Copies = missingCopies.Select(c => c.TextName).Distinct().ToList() });
+                    var message = "COPY need to be downloaded. When it's done restart the server";
```
smedilol: Isn't it possible to avoid a server restart.
Missing copies is very common case. It will be very boring to restart the server each time it happens.

maxime645: I forgot to remove the notification. All create/delete/update actions on a copy lead to an automatic symbol reloading

CLI/src/CLI.cs
```diff
@@ -109,36 +109,54 @@ class CLI
...
+                //Collect the missing copies
+                File.WriteAllLines(config.HaltOnMissingCopyFilePath, missingCopies.Select(c => c.TextName).Distinct());
+
+                //Create extraced copies file even if copy are missing
+                CreateExtracedCopiesFile(config, usedCopies);
 
-                if (diagDetected)
-                    throw new CopyLoadingException("Diagnostics detected while parsing Intrinsic file", null, null, logged: false, needMail: false);
+                //If copies are missing, don't try to parse main input files.
+                throw new MissingCopyException("Some copy are missing", config.Dependencies.FirstOrDefault(), null, logged: false, needMail: false);
```
fm-117: Why use FirstOrDefault ? The origin of the error could be located in any of the dependency files.

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -146,6 +146,9 @@ private void WarningTrigger(object sender, string message)
         /// <param name="missingCopies">List of missing copies name</param>
         private void MissingCopiesDetected(object fileUri, MissingCopiesEvent missingCopiesEvent)
         {
+            //Warning the main file could not be opened
+            //This event can be use when a dependency have not been loaded
```
fm-117: 'can be used'

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -527,7 +527,14 @@ private void RefreshCustomSymbols()
             try
             {
                 _customSymbols = Tools.APIHelpers.Helpers.LoadIntrinsic(TypeCobolConfiguration.Copies, TypeCobolConfiguration.Format, DiagnosticsErrorEvent); //Refresh Intrinsics
-                _customSymbols = Tools.APIHelpers.Helpers.LoadDependencies(TypeCobolConfiguration.Dependencies, TypeCobolConfiguration.Format, _customSymbols, TypeCobolConfiguration.InputFiles, TypeCobolConfiguration.CopyFolders, DiagnosticsErrorEvent); //Refresh Dependencies
+                _customSymbols = Tools.APIHelpers.Helpers.LoadDependencies(TypeCobolConfiguration.Dependencies, TypeCobolConfiguration.Format, _customSymbols, TypeCobolConfiguration.InputFiles, 
+                    TypeCobolConfiguration.CopyFolders, DiagnosticsErrorEvent, out List<RemarksDirective.TextNameVariation> usedCopies, out List<CopyDirective> missingCopies); //Refresh Dependencies
+
+                if(missingCopies.Count > 0)
+                {
+                    MissingCopiesEvent(TypeCobolConfiguration.Dependencies?.FirstOrDefault(), new MissingCopiesEvent() { Copies = missingCopies.Select(c => c.TextName).Distinct().ToList() });
```
fm-117: Here we check if config.Dependencies is not null but not in CLI.cs line 153.

TypeCobol/Tools/APIHelpers/Helpers.cs
```diff
@@ -73,11 +73,42 @@ public static SymbolTable LoadIntrinsic(List<string> paths, DocumentFormat intri
...
-
             var parser = new Parser(intrinsicTable);
+
+            parser.Init(path, new TypeCobolOptions { ExecToStep = ExecutionStep.SemanticCheck }, format, copyFolders);
+            parser.Parse(path); //Parse the dependency file
+
+            return parser.Results;
+        }
+
+        public static IEnumerable<string> GetDependenciesMissingCopies([NotNull] List<string> paths, DocumentFormat format, SymbolTable intrinsicTable, List<string> copyFolders, EventHandler<DiagnosticsErrorEvent> diagEvent)
```
fm-117: The return type could be a `IDictionary<string, IList<string>>` to give the list of missing copies for each dependency file.

maxime645: The result of this method is not meant to be displayed to the user, only to download the missing copies. It is useless to return the list of missing copies per file here.
However, a MissingCopyException writes a file that misses some copies. MissingCopyExceptions can be thrown after a LoadDependencies, it can be interesting to return such a list there.

CLI/src/CLI.cs
maxime645: Your right, it has been changed. Now, for each dependency that misses some copys, we throw an exception with the name of the file as argument

TypeCobol.LanguageServer/Workspace.cs
maxime645: The purpose of a MissingCopiesEvent is to trigger a copy loading client side. My question is more why do we even need to give a filepath ? @smedilol

smedilol: The MissingCopiesParams requires a TextDocumentIdentifier that's why the filePath is required.
@mayanje Can we remove the property TextDocumentIdentifier in MissingCopiesParams ?
If so, it will the subject of another issue.

TypeCobol/Tools/APIHelpers/Helpers.cs
smedilol: MissingCopyException is only used in CLI and it already got the filePath for which copies are missing.

smedilol:
See new PR #1510

PR #1473, opened 2019-09-18
1450 goback stoprun

LANOYDO:

PR #1474, opened 2019-09-19
1455 warning paragraph section empty

LANOYDO:


maxime645:
@smedilol, there is a problem with one of the tests, here. 
The test 'CompilerDirecting.pgm' uses many times the compiler directive BASIS. I spoke with @sarterbe about it, we don't use it in our company and I am not sure if the behaviour of the directive has been implemented.
We need to discuss if this test is relevant, if we need to update it or if in these cases of empty paragraphs, the warning is accurate

maxime645:
See new version #1473

Issue #1455, opened 2019-09-05
Warning if paragraph/section is empty


smedilol:
**Is your feature request related to a problem? Please describe.**
Generate a warning if a paragraph or a section is empty.

Cobol example
```cobol
      *Warning INIT is empty
       INIT.
           .

      *Ok
       TASK-1.
           display "task-1"
           .
```

PR #1479, opened 2019-09-23
WI #1455, add Diagnostic warning when a paragraph or a section is empty

LANOYDO:
a section/paragraph is empty when it has no child or when its child/children is/are an {End}


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -723,19 +725,38 @@ private static SymbolReference ResolveProcedureName(SymbolTable table, SymbolRef
...
+            if (found.Count > 1)
+                DiagnosticUtils.AddError(node, nodeTypeName + " \'" + node.Name + "\' already declared");
+
+            // a section/paragraph is empty when it has no child or when its child/children is/are an {End} node
+            bool empty = true;
+            int nodesCount = node.Children.Count; 
+            // checks if all children are of type {End} : if not one then empty = false 
+            for (int i = 0; i < nodesCount; i++)
+            {
+                if  ((node.Children[i].Children.Count == 1 && node.Children[i].Children[0].GetType() == typeof(Nodes.End)) == false)
```
smedilol: Can you add a comment to explain why you test `node.Children[i].Children.Count == 1`? I guess it's because of Sentence nodes.

smedilol: Ok, but what is the type of "node.Children[i]" ?
One should think that an empty paragraph contains no Children or maybe just a child of type End.
So the test would be:
`if  ((node.Children.Count == 1 && node.Children[0].GetType() == typeof(Nodes.End)) == false)`

LANOYDO: done by using explicit word "sentence"
https://github.com/TypeCobolTeam/TypeCobol/pull/1479/commits/90e9535abb7b0fc6faada8125915a1be4e97a24a

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -724,18 +724,22 @@ private static SymbolReference ResolveProcedureName(SymbolTable table, SymbolRef
                 DiagnosticUtils.AddError(node, nodeTypeName + " \'" + node.Name + "\' already declared");
 
             // a section/paragraph is empty when it has no child or when its child/children is/are an {End} node
-            bool empty = true;  // default value
-            foreach (Node sentence in node.Children)
+            if (!node.Name.Equals("INIT-LIBRARY", StringComparison.InvariantCultureIgnoreCase))
```
mayanje: @LANOYDO, @smedilol @fm-117 
So this is a restriction, as soon as  one of  the Section or a Paragraph is named INIT-LIBRARY this check is no longer performed?
I say that because I know that INIT-LIBRARY paragraph name is checked by the LibraryChecker, maybe we can identify that the paragraph is the INIT-LIBRARY paragraph and only ignore the check in this situation.
This information can be set by the LibraryChecker by adding for instance a IsInitLibrary boolean property to the Paragraph, but in this case we must assume that the LibraryChecker is called before the SectionOrParagraphUsageChecker.

smedilol: We can remove this test.
An `exit` statement can be put in paragraph INIT-LIBRARY to avoid the warning.

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1479/commits/01c239da31a567402e43d3e2e6c655c577177b67

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -716,19 +718,35 @@ private static SymbolReference ResolveProcedureName(SymbolTable table, SymbolRef
...
+        protected static void Check<T>(string nodeTypeName, T node, [NotNull] IList<T> found) where T : Node
         {
-            if (found.Count > 1) DiagnosticUtils.AddError(node, "Symbol \'" + node.Name + "\' already declared");
+            if (found.Count > 1)
+                DiagnosticUtils.AddError(node, nodeTypeName + " \'" + node.Name + "\' already declared");
+
+            // a section/paragraph is empty when it has no child or when its child/children is/are an {End} node
+            bool empty = true;  // default value
+            foreach (Node sentence in node.Children)
+            {
```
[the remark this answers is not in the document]
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1479/commits/01c239da31a567402e43d3e2e6c655c577177b67

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -716,19 +719,48 @@ private static SymbolReference ResolveProcedureName(SymbolTable table, SymbolRef
             return null;
         }
 
-        protected static void Check<T>(T node, [NotNull] IList<T> found) where T : Node
+        protected static void Check<T>(string nodeTypeName, T node, [NotNull] IList<T> found) where T : Node
```
fm-117: Maybe we should mark this method as `private` instead of `protected` because it assumes that T is either Section or Paragraph to work properly. Inheritors may not be aware of that constraint, there are currently no inheritors so this is optional.

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1479/commits/b3bf2dca5a1639dffc0b0068fe759f303b177681

Issue #1316, opened 2019-03-18
Create an error if paragraph INIT-LIBRARY is not present for program with public procedures


smedilol:
Paragraph INIT-LIBRARY is used by generated code for program with public procedures.
Currently there is no error, before code generation if this paragraph is missing.
You'll only get an error after compilation.

The goal is to get an error before code generation.

PR #1488, opened 2019-10-03
WI #1316 Create an error if paragraph INIT-LIBRARY is not present for…

LANOYDO:
… program with public procedures


TypeCobol.Test/Parser/Programs/TypeCobol/LibraryFunctionDeclarativeCheck.rdzPGM.txt
```diff
@@ -1,5 +1,5 @@
 ﻿--- Diagnostics ---
-Line 38[9,12] <27, Error, Syntax> - Syntax error : Inside a library only function declaration or declaratives are allowed  / sentence-1
+Line 38[9,12] <27, Error, Syntax> - Syntax error : A program which contains public procedure must have INIT-LIBRARY as first paragraph.
```
fm-117: It would be better to have a test for each case :
- first paragraph is not INIT-LIBRARY
- invalid declarations in public library
- missing INIT-LIBRARY

In fact the second case corresponds to SECTION declarations
```cobol
MySection SECTION.
    DISPLAY "1"
```

LANOYDO: fixed; new commit:
https://github.com/TypeCobolTeam/TypeCobol/tree/4e9f45bcbfdc878acbe170739c3bbd60dcc8a94a

TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -883,14 +882,25 @@ public static void CheckLibrary([NotNull] ProcedureDivision procedureDivision)
...
-                            " / " + child.ID);
+                            : procedureDivision)
+                        : child;
+                    if (firstParagraphChecked)
+                    {
+                        // this case corresponds to SECTION declarations or statements not inside a paragraph
+                        string nodeInfo = " " + node.Name + " / " + node.ID;
+                        if (nodeInfo == "  / ")  nodeInfo = "";
+                        DiagnosticUtils.AddError(node,
+                            "Inside a library only function declaration or declaratives are allowed" + nodeInfo);
```
smedilol: Can you change the message to:
"A program which contains public procedure cannot contains section or statement not under a paragraph."

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1488/commits/6c457786fae470a996fff3b9f4509cce5e20546a

Issue #1480, opened 2019-09-24
Omittable FunctionParamater


maxime645:
**Describe the bug**
FOLLOWING #791 
When a FunctionParameter is typed and is declared as omittable in a function declaration, it will be generated with the question mark in the DataDefinition.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       identification division.
       program-id. MyPGM.
       data division.
       working-storage section.

       01 GroupType TYPEDEF STRICT public.
           05 GroupType-DATA1      PIC X(5).
           05 GroupType-DATA2      PIC X(5).

       procedure division.
       INIT-LIBRARY.
           GOBACK
           .

        DECLARE PROCEDURE MyProc
                            INPUT MyProc-DATA1 ? TYPE GroupType.
        DATA DIVISION.
        WORKING-STORAGE SECTION.
        01 AnotherGroupType TYPE GroupType.
        PROCEDURE DIVISION.
           MOVE AnotherGroupType TO MyProc-DATA1
           .
        END-DECLARE.
       end program MyPGM.
```
The dataDefinition of MyProc-DATA1 in the Linkage Section of MyProc will be :
```cobol
       01 MyProc-DATA1 ? .
           02 ScalarCharType-DATA1 PIC X(5).
           02 ScalarCharType-DATA2 PIC X(5).
```

**Expected behavior**
The question mark shouldn't be generated

Expected (Type)Cobol code : 
```cobol
       01 MyProc-DATA1.
           02 ScalarCharType-DATA1 PIC X(5).
           02 ScalarCharType-DATA2 PIC X(5).
```

Issue #1477, opened 2019-09-20
Unwanted empty lines on generated code


maxime645:
**Describe the bug**
FOLLOWING #791
There are some unwanted break lines in the generated code.
When a DataDef is declared with a TYPE on two lines, the point finishing the DataDef will be generated alone on a second line.
At the end of the DataDef, there is also an unwanted empty line.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. PGM1.
      
       DATA DIVISION.
       Working-STORAGE SECTION.
      
       01 TypeDefInMainProgram TYPEDEF STRICT PUBLIC.
           05 DATA1 PIC X.
           05 DATA2 PIC X.
           
       01 TypeDefInMainProgram1
            TYPE TypeDefInMainProgram.
            
       01 TypeDefInMainProgram2 TYPE TypeDefInMainProgram.
      
       PROCEDURE DIVISION.
            GOBACK
            .
       END PROGRAM PGM1.
```

Expected (Type)Cobol code : 
```cobol
      *TypeCobol_Version:0.1(alpha)
       IDENTIFICATION DIVISION.
       PROGRAM-ID. PGM1.
      
       DATA DIVISION.
       Working-STORAGE SECTION.
      
      *01 TypeDefInMainProgram TYPEDEF STRICT PUBLIC.
      *    05 DATA1 PIC X.
      *    05 DATA2 PIC X.
           
      *01 TypeDefInMainProgram1
      *     TYPE TypeDefInMainProgram.
       01 TypeDefInMainProgram1.
           02 DATA1 PIC X.
           02 DATA2 PIC X.
            
      *01 TypeDefInMainProgram2 TYPE TypeDefInMainProgram.
       01 TypeDefInMainProgram2.
           02 DATA1 PIC X.
           02 DATA2 PIC X. 
      
       PROCEDURE DIVISION.
            GOBACK
            .
       END PROGRAM PGM1.
```

PR #1489, opened 2019-10-04
WI #1480 Fix linkage generation when using omittable typed parameters

fm-117:
Fixes issues #1480 and #1477
`ExtractTypeInfo` strategy has been fixed to neither output QuestionMark nor PeriodSeparator. In theory, `ExtractAfterLevel` strategy is also concerned but in practice this strategy is never used for parameter descriptions.
We can either :
- leave the `ExtractAfterLevel` strategy untouched (staying in line with its current name)
- add case for QuestionMark to handle the case of future use of this strategy with param descriptions.


fm-117:
See #1505

Issue #1454, opened 2019-09-05
Warning if debugging mode is active


smedilol:
**Is your feature request related to a problem? Please describe.**
Debugging sometimes stay active on program.
Debugging mode should only be used when you actually have something to debug.

Generate a warning when Debugging mode is active so the developer will not forget to deactivate it.

Cobol example
```cobol
TODO
```

PR #1492, opened 2019-10-08
WI #1454, add Diagnostic warning when a debugging mode is active

LANOYDO:


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -372,6 +372,15 @@ public override bool Visit(DataDefinition dataDefinition)
             return true;
         }
 
+        public override bool Visit(SourceComputer sourceComputer)
+        {
+            if (sourceComputer.Lines.Any(l => (l as CodeElementsLine)?.ScanState.WithDebuggingMode == true))
```
fm-117: Why use Lines property ?? `sourceComputer.CodeElement.DebuggingMode?.Value` should tell us whether debugging mode is active or not.

mayanje: This a scanner consequence of having WITH DEBUGGING MODE clause set within the SourceComputerParagraph code element of the SourceComputer node.

Yes a more explicit way would be to check that the consumed tokens of the SourceComputerParagraph code element contains at least the tokens DEBUGGING MODE.

smedilol: +1, checking property DebuggingMode of CodeElement SourceComputerParagraph is clearer.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -372,6 +372,15 @@ public override bool Visit(DataDefinition dataDefinition)
...
+            {
+                DiagnosticUtils.AddError(sourceComputer, "Debugging mode is active", MessageCode.Warning);
```
smedilol: It'd be nice to position the diagnostic on the line that contains "DEBUGGING MODE" rather than  SOURCE-COMPUTER.
You can pass the token sourceComputer.DebuggingMode.Token to AddError method to do so.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
mayanje: Sure

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/tree/94dd2d71d6dd982de8de0c257ff90aba9e55e698

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/tree/94dd2d71d6dd982de8de0c257ff90aba9e55e698

TypeCobol/Compiler/Parser/DiagnosticUtils.cs
```diff
@@ -18,9 +18,17 @@ internal static class DiagnosticUtils
             var parserDiag = new ParserDiagnostic(message, e.StartIndex + 1, e.StopIndex + 1, e.ConsumedTokens[0].Line, null, code);
             e.Diagnostics.Add(parserDiag);    
         }
-		internal static void AddError(CodeElement e, string message, Scanner.Token token, string rulestack = null, MessageCode code = MessageCode.SyntaxErrorInParser) {
+		internal static void AddError(CodeElement e, string message, Scanner.Token token, string rulestack = null, MessageCode code = MessageCode.SyntaxErrorInParser, bool withStackRule = true) {
```
fm-117: Could youcheck the behavior when you call the roiginal method with rulestack set to null and no modification. I'm not sure any modification is required.

LANOYDO: done in last commit

smedilol:
Following today's meeting, the only change since last approval of everyone is an update of an unit test.
So it's ok to merge this PR in this state.

Issue #1487, opened 2019-10-03
Anonymous array types with index are not supported


fm-117:
**Describe the bug**
An exception is thrown during code generation when using an anonymous indexed array typedef. This bug affects both unnamed and FILLER arrays.

**To Reproduce**
```cobol
       01  MyType TYPEDEF STRICT.
           05 FILLER OCCURS 10 INDEXED BY I-MyType.
                10 item           PIC X.

       01 MyTable type MyType.
```
This code will throw NullReferenceException in TypeCobol.Codegen.Nodes.TypedDataNode.BuiltIndexMap.

This code however generates without error :
```cobol
       01  MyType TYPEDEF STRICT.
           05 MyType-TBL OCCURS 10 INDEXED BY I-MyType.
                10 item           PIC X.

       01 MyTable type MyType.
```

**Expected behavior**
Anonymous or FILLER array types should be expanded without error.

Expected (Type)Cobol code : 
```cobol
       01 MyTable.
          05 FILLER OCCURS 10 INDEXED BY I-MyType.
             10 item PIC X.
```

fm-117:
May be related to #216.

PR #1495, opened 2019-10-09
WI #1487 Fix bugs on indexes generation

fm-117:
Link to issue : #1487 
This PR fixes 2 bugs :
- when an index is declared in a typedef on an anonymous node, a NullReferenceException was previously thrown
- when an index is declared on a typed node (either named or anonymous), generated names in declaration and usages were previously different, leading to an incorrect generated code

We now support indexes on anonymous declarations and match names in declaration and usages correctly. A hash is added to the index name as soon as it is used in a qualified reference or declared in a typedef.


TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -282,20 +282,24 @@ public void SetFlag(Flag flag, bool value, bool bRecurse = false)
...
+                        qn.Add(current.Name);
                     }
-                    parent = parent.Parent;
+
+                    current = current.Parent;
                 }
+
                 qn.Reverse();
                 _qualifiedName = new URI(qn);
                 return _qualifiedName;
```
LANOYDO: blank lines are not usefull here

fm-117:
See #1506

Issue #1386, opened 2019-06-07
Hover : Display the definition of a variable


smedilol:
**Is your feature request related to a problem? Please describe.**
When the mouse is over a reference to a variable. The hover could display information about its definition.

TypeCobol code that you'd like to write.
```cobol
        01 var1 pic 9(10).
        01 var2.
           05 var3 pic X(10).
            move var1 to var2
```
When mouse is over "var1" on the line "move var1 to var2", hover must show `01 var pic 9(10).`
When mouse is over "var2" on the line "move var1 to var2", hover must show: 
```        
        01 var2.
           05 var3 pic X(10).
```

PR #1503, opened 2019-10-14
1386 on hover display variable definition rebase

fm-117:
Issue #1386 
Rebased version of PR #1461


TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -544,6 +545,17 @@ protected override Hover OnHover(TextDocumentPosition parameters)
                         }
                     }
                     break;
+                default:
+                    // Use userFilterToken to find the variable under mouse cursor.
+                    if (userFilterToken != null)
+                    {
+                        var candidates = matchingNode.SymbolTable.GetVariablesExplicit(new URI(userFilterToken.SourceText)).ToList();
```
smedilol: All information about Tokens<->SymbolInformation<->StorageArea<->Node are stored in CodeElement and Node.
See 
`CodeElement.SymbolInformationForTokens`
Then when you have the SymbolInformation you can iterate list:
- `CodeElement.StorageAreaReads`
- `CodeElement.StorageAreaWrites`

To check if a StorageArea contains this SymbolInformation 

Then with the StorageArea you can retrieve the corresponding DataDefinition already resolved by CrossChecker with
`node.GetDataDefinitionFromStorageAreaDictionary()`.

If one element is not found (eg. CodeElement.SymbolInformationForTokens doesn't contains the token, it means the token is not a SymbolInformation) and you can stop (there is no variable to hover).

smedilol: A quick code to illustrate:
```c#
// Use userFilterToken to find the variable under mouse cursor.
                    if (userFilterToken != null)
                    {

                        if(matchingNode.CodeElement.SymbolInformationForTokens.TryGetValue(userFilterToken, out var symbolInformation) )
                        {
                            //Search if a StorageArea is related to this SymbolInformation
                            StorageArea storageArea = null;
                            foreach (var tempStorageArea in matchingNode.CodeElement.StorageAreaReads)
                            {
                                if (tempStorageArea.SymbolReference == symbolInformation)
                                {
                                    storageArea = tempStorageArea;
                                    break;
                                }
                            }
                            foreach (var tempStorageArea in matchingNode.CodeElement.StorageAreaWrites)
                            {
                                if (tempStorageArea.StorageArea?.SymbolReference == symbolInformation)
                                {
                                    storageArea = tempStorageArea.StorageArea;
                                    break;
                                }
                            }
                            //End of search

                            if (storageArea != null)
                            {
                                //Get the cached DataDefinition for this StorageArea
                                var dataDef = matchingNode.GetDataDefinitionFromStorageAreaDictionary(storageArea);

                                if (dataDef != null)
                                {
                                    //Display the DataDefinition
                                    message = dataDef.ToString();
                                }
                            }

                        }
                    }
```

smedilol: Beware GetHashCode() method is missing in Token class.
So `matchingNode.CodeElement.SymbolInformationForTokens.TryGetValue(userFilterToken, out var symbolInformation) ` will not return anything.
A quick hash code for Token:
```C#
public override int GetHashCode()
        {
            return this.Type.GetHashCode() * 37
                   + this.Column * 37
                   + this.Length * 37 
                   + this.Line *37
                   + this.Text.GetHashCode();
        }
```
But I need to check if this hashCode can be improved.

TypeCobol/Compiler/Scanner/Token.cs
```diff
@@ -21,24 +21,23 @@ public class Token : Antlr4.Runtime.IToken, IVisitable
...
-                   && this.HasClosingDelimiter == tokenCompare.HasClosingDelimiter
-                   && this.LiteralValue == tokenCompare.LiteralValue
-                   && this.TokenType == tokenCompare.TokenType
-                   && this.Text == tokenCompare.Text;
+            if (obj is Token token)
+            {
+                return token.Type == this.Type &&
+                       token.StartIndex == this.StartIndex &&
+                       token.StopIndex == this.StopIndex &&
+                       token.TokensLine == this.TokensLine;
```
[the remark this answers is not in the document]
fm-117: I tried to use TokenSource in the beginning but I had trouble since this property may or may not be set. Two distinct tokens may both have a null `TokenSource` and comparing them will yield wrong result.
It appears that `Type`, `StartIndex`, `StopIndex` and `TokensLine` are always set and are constitutive to the Token. A new `TokenLine` is created in `CompilationDocument` (creation or update) so two tokens defined one in a COPY, the other in a Program will have different TokensLine and will be considered distinct.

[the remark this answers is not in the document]
smedilol: Can you test this case:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. RB1LEQU1.
       data division.
       working-storage section.
       COPY YTOTO
       01 toto pic X.
      
       procedure division.
           move toto to toto
           .
       end program RB1LEQU1.
```
And COPY YTOTO contains:
     01 toto pic 9.
at line 6.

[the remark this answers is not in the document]
fm-117: Actually the token in the copy should be at line 9 to match the one in the move statement because the comparison is made on userFilterToken which is positioned where the hover is requested.
Here are my findings in debug mode :
- the SymbolInformationForTokens dictionary does not include tokens from COPY so it won't create any conflict
- using the Watcher in VS, I managed to compare two similar tokens written one in a COPY, the other in main program : the two tokens are considered distinct by the Equals method and they differ only in their TokensLine which is the expected behavior.

Also, the hover request on a variable defined in a COPY won't return anything.

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -984,4 +1020,60 @@ public CodeElementWrapper(CodeElement codeElement)
...
+        /// <param name="level">Starting level, default is 0.</param>
+        /// <returns>IEnumerable of Tuple, Item1 is depth, Item2 is ITextLine.</returns>
+        private static IEnumerable<Tuple<int, ITextLine>> GetLinesWithLevel(Node node, int level = 0)
+        {
+            if (node == null) yield break;
+
+            // Lines of current level in node
+            if (node.CodeElement?.ConsumedTokens != null)
+            {
+                var lines = new List<ITextLine>();
```
smedilol: Why a list here ?
It seems you only need to remember the last line seen.

fm-117: Yes, you're right. This has been fixed with a47405251aca00c87ab802227c12827430bb7377.

PR #1506, opened 2019-10-14
WI #1487 Fix bugs on indexes generation

fm-117:
Rebased version of #1495 
Link to issue : #1487 
This PR fixes 2 bugs :
- when an index is declared in a typedef on an anonymous node, a NullReferenceException was previously thrown
- when an index is declared on a typed node (either named or anonymous), generated names in declaration and usages were previously different, leading to an incorrect generated code

We now support indexes on anonymous declarations and match names in declaration and usages correctly. A hash is added to the index name as soon as it is used in a qualified reference or declared in a typedef.


Codegen/src/Nodes/TypedDataNode.cs
```diff
@@ -486,9 +486,11 @@ internal static List<ITextLine> LineToTextLines(string line)
             }
 
             list_items.AddRange(pathVariables);
-            string qn = string.Join(".", list_items.ToArray());
-            AddIndexMap(rootNode, rootNode.QualifiedName.ToString(), indexes, map);
-            AddIndexMap(ownerDefinition, qn, indexes, map);
+            string qn = string.Join(".", list_items.Where(item => item != null));
+
+            // If indexes directly come from their DataDef, use the computed path (qn). If they come from a TypeDef, the path is the rootNode.QualifiedName.
+            AddIndexMap(rootNode, rootNode == ownerDefinition ? qn : rootNode.QualifiedName.ToString(), indexes, map);
```
LANOYDO: For visibility reason the line 492 should be
AddIndexMap(rootNode, (rootNode == ownerDefinition ? qn : rootNode.QualifiedName.ToString()), indexes, map);

Codegen/test/resources/output/TypeCobol/IndexesAndAnonymousDataDef.rdz.tcbl
```diff
@@ -0,0 +1,114 @@
...
+      *   05 Items Type ElementType
+      *      OCCURS 10 INDEXED BY Idx-InNamedArray.
+          05 Items
+             OCCURS 10 INDEXED BY Idx-InNamedArray.
+           06 item-4 PIC X.
+                                                   
+       01 TableWithFillerArray.
+      *   05 FILLER Type ElementType
+      *      OCCURS 10 INDEXED BY Idx-InFillerArray.
+          05
```
smedilol: FILLER is lost here.
The bug was already there before this PR. I'll create a new issue.

PR #1471, opened 2019-09-17
WI #1453, add Diagnostic when using ALTER

LANOYDO:

Issue #1453, opened 2019-09-05
Warning if ALTER is used


smedilol:
**Is your feature request related to a problem? Please describe.**
Generate a warning if `ALTER` is used.

PR #1531, opened 2019-10-28
WI #1453 Add Diagnostic when using ALTER

LANOYDO:
It's a rebase version: PR is #1471


LANOYDO:
WI #1453 Add Diagnostic when using ALTER closed

Issue #1533, opened 2019-10-31
Detect cobol construct with a mix of Debug and "Normal" lines


smedilol:
**Is your feature request related to a problem? Please describe.**
Detect when a Cobol construct use a mix of Debug and "normal" lines and the behavior will change depending on debug is active or not.

TypeCobol code that you'd like to write.
The following code will not produce the same results depending on debug is active or not.
```cobol
       if condition1
           continue
      Delse
           set condition2 to true
       end-if
```

A warning must be positioned on the "else" statement.

We need others tests cases.

smedilol:
Will be implemented in our private quality extension.

PR #1534, opened 2019-11-04
WI #1533 Detect cobol construct with a mix of Debug and "Normal" lines

LANOYDO:


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -61,6 +63,36 @@ public override bool BeginCodeElement(CodeElement codeElement)
...
+        {
+            // detect CodeElement with a mix of Debug and "Normal" lines in debugging mode         
+            if (DebuggingMode && end.CodeElement.Type != CodeElementType.SentenceEnd)
+            {
+                CodeElement openingCodeElement = end.Parent.CodeElement;
+                if (openingCodeElement != null)
+                {
+                    CodeElement endCodeElement = end.CodeElement;
+                    bool isStartDebug = IsCodeElementInDebug(openingCodeElement);
+                    bool isEndDebug = IsCodeElementInDebug(endCodeElement);
```
mayanje: @LANOYDO One things to know is that in cobol many constructions have optional END instruction, so optional END have no associated End node, thus this algorithm will never recognize something like.
````cobol
      d    evaluate xxxx
      d         when 'toto' display "found"
                when other display "not found"
```

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -61,6 +63,36 @@ public override bool BeginCodeElement(CodeElement codeElement)
...
+                    if (isStartDebug == isEndDebug && openingCodeElement is CodeElements.IfStatement)
```
fm-117: This limit the check only on IfStatements, Is it what we want to do in this issue ?

smedilol:
I propose to close this PR for now.
Wait for full specifications.

Issue #1515, opened 2019-10-17
Codegen : Call procedure pointer causes exception


smedilol:
**Describe the bug**
Bug when you generate the following code:

**To Reproduce**
Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       LINKAGE SECTION.
       01  PP                            PROCEDURE-POINTER.
      
       PROCEDURE DIVISION.
           CALL PP
           GOBACK
           .
       END PROGRAM MyPGM.
      
```
The line which cause the problem is `CALL PP`
Exception
```
   at TypeCobol.Server.CLI.runOnce2(TypeCobolConfiguration config, AbstractErrorWriter errorWriter)
   at TypeCobol.Server.CLI.runOnce(TypeCobolConfiguration config)


   at TypeCobol.Codegen.Nodes.ProcedureStyleCall.get_Lines()
   at TypeCobol.Codegen.Generators.DefaultGenerator.d__40.MoveNext()
   at TypeCobol.Codegen.Generators.DefaultGenerator.LinearGeneration[A](LinearNodeSourceCodeMapper mapper, LinearNodeSourceCodeMapper clonedMapper, IReadOnlyList`1 Input, LineMappingCtx lmCtx, Int32 startLine, Int32 endLine)
   at TypeCobol.Codegen.Generators.DefaultGeneratorWithLineMap.LinearGeneration[A](LinearNodeSourceCodeMapper mapper, IReadOnlyList`1 Input, LinearNodeSourceCodeMapper clonedMapper)
   at TypeCobol.Codegen.Generators.DefaultGenerator.TreeToCode()
   at TypeCobol.Codegen.Generators.MixedTransformGenerator.Generate(CompilationUnit compilationUnit, ColumnsLayout columns)
   at TypeCobol.Server.CLI.runOnce2(TypeCobolConfiguration config, AbstractErrorWriter errorWriter)

```

**Expected behavior**
No exception.

fm-117:
The call is mistakenly parsed as a Procedure call (TC-specific) instead of a regular COBOL procedure-pointer call. After that, Codegen will try to generate code for the call but will fail as there is no FunctionDeclaration.

It looks like there is an ambiguity in ANTLR grammar for this kind of CALL

```
cobolCallStatement:
	CALL programNameOrProgramEntryOrProcedurePointerOrFunctionPointerVariable
[...]

PR #1550, opened 2019-11-15
WI #1515 Correct grammar ambiguity between variable CALL and proc CAL…

fm-117:
…L during cross checking
Issue is #1515


fm-117:
This PR is abandoned because the fix consisting in swapping nodes is incomplete. See original issue #1515 for details.

Issue #1552, opened 2019-11-15
Codegen : picture clause must not include line break


smedilol:
**Describe the bug**
Bug reported with TypeCobol version > v1.4.0-prerelease

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. RB1LEQU1.
       procedure division.
       declare procedure Foo2 public
                 input  EquipementObject PIC X(3)
                        EquipementMainReference
                                         PIC S9(4)
                        EquipementSecondaryReference
                                         PIC S9(4)
                        EquipementOccurrence
                                         PIC S9(4).
       procedure division.
       end-declare.
       end program RB1LEQU1.
```
The generated code will "cut" picture clause at an incorrect position for `EquipementOccurrence` variable:

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. ce3384b8.
       DATA DIVISION.
       LINKAGE SECTION.
      *RB1LEQU1.Foo2 - Params :
      *		input(EquipementObject: pic X(3), EquipementMainReference: pic
      *S9(4), EquipementSecondaryReference: pic S9(4), EquipementOccurre
      *nce: pic S9(4))
       01 EquipementObject PIC X(3).
       01 EquipementMainReference
                                  PIC
           S9(4).
       01 EquipementSecondaryReference
            PIC S9(4).
       01 EquipementOccurrence
                                  PIC S9(
           4).
       PROCEDURE DIVISION
             USING BY REFERENCE EquipementObject
                   BY REFERENCE EquipementMainReference
                   BY REFERENCE EquipementSecondaryReference
                   BY REFERENCE EquipementOccurrence
           .
      *RB1LEQU1.Foo2 - Params :
      *		input(EquipementObject: pic X(3), EquipementMainReference: pic
      *S9(4), EquipementSecondaryReference: pic S9(4), EquipementOccurre
      *nce: pic S9(4))
       END PROGRAM ce3384b8.
```

Expected (Type)Cobol code : 
The picture clause itself (not the word `PIC` or `PICTURE` must not include a line break.
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. ce3384b8.
       DATA DIVISION.
       LINKAGE SECTION.
      *RB1LEQU1.Foo2 - Params :
      *		input(EquipementObject: pic X(3), EquipementMainReference: pic
[...]

PR #1558, opened 2019-11-26
1552 unwanted line breaks

fm-117:
Fix for issue #1552.

The idea behind this fix is that the line breaks added by `TypedDataNode.FlushConsumedTokens` were not considered as line breaks but as standard characters. This would lead to errors when splitting based on length. Now the splitting process starts by splitting over line breaks and then apply length restrictions.


smedilol:
@rooksdo I removed myself as a reviewer. I let you check if everything is ok from a TypeCobol developer perspective.

Issue #1484, opened 2019-09-26
Completion between :: and :: raises an exception


mayanje:
When the completion is invoked between two sequences of :: and ::, that is to say the cursor locate like that :: | ::
This raises an exception
![CobonColonColonColonompletion](https://user-images.githubusercontent.com/19149269/65701663-fce85b80-e081-11e9-894b-55776e07dc25.png)

fm-117:
May be related to #1428.

Issue #1428, opened 2019-08-08
Completion after `:` crash


smedilol:
**Describe the bug**
Completion after `:` instead of `::` cause a crash.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
        01 Var1 type Date.
            move Var1:
```
Ask completion.

**Expected behavior**
2 possibilities:
- Smart : Detect that a `:` is missing and it's not possible to use Reference Modification, then propose standard completion after `::`  and add `:` automatically
   - With a syntax like this `Var1(Var2:`  it's not possible to know if the user want to use a reference modification or forget to add `:`.
   - If it's a reference modification, then all variables of type numeric should be proposed.
- Simple : Completion should show an empty list or a list with a message indicating that the syntax is not correct for autocompletion.

smedilol:
Another case which might be related:
```
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 Group1.
         05 Var1 pic X(10).
       01 Var2 pic 9.
       01 Var3 pic X(10).
       LINKAGE SECTION.
      *=================================================================
       procedure division.
           MOVE Group1::Var1(1:) TO Var3
           .
       END PROGRAM MyPGM.
```
If you ask completion after `(1:` you get an Exception in SymbolTable.

PR #1562, opened 2019-11-27
WI #1428 #1484 Avoid exception when completing after or between colon

fm-117:
Fix for issues #1428 and #1484.


TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -413,9 +413,10 @@ public static IEnumerable<CompletionItem> GetCompletionForQualifiedName(Position
                                 ((firstSignificantToken != null && ((t.StartIndex >= firstSignificantToken.EndColumn && t.Line == firstSignificantToken.Line) || t.Line > firstSignificantToken.Line)) 
                                 || firstSignificantToken == null) 
                                 && ((t.EndColumn <= position.character && t.Line == position.line + 1) || t.Line < position.line + 1))
-                        .Select(t => t.Text);
-                qualifiedName.Reverse();
```
mayanje: Why the call to _qualifiedName.Reverse();_ is no longer needed?

fm-117: It was useless since the beginning because it is the `Reverse()` from LINQ (instead of `Reverse()` from generic List). The Reverse method from LINQ doesn't change the original IEnumerable on which it is called, it produces a new IEnumerable configured to iterate in reverse order. Here the return value of the call is not used so the call itself is useless.

Issue #1559, opened 2019-11-26
Invalid generated code when decrementing pointer


fm-117:
**Describe the bug**
The generated code for decrementing a pointer using a numeric variable is wrong. Does not compile on z/OS.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCSAMPLE8.
       data division.
       working-storage section.
       01 MyPointer POINTER.
       01 n PIC 9.
       procedure division.
           SET MyPointer UP BY n
           SET MyPointer UP BY 1
           SET MyPointer DOWN BY n
           SET MyPointer DOWN BY 1
           .
       end program TCSAMPLE8.
```

Codegen will produce
```cobol
      *    SET MyPointer DOWN BY n
           ADD -n to MyPointerae9be588
```
which is not a valid COBOL syntax.

**Expected behavior**
Use `SUBTRACT` keyword instead.

Expected (Type)Cobol code : 
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCSAMPLE8.
       data division.
       working-storage section.
      *01 MyPointer POINTER.
       01  MyPointer POINTER.
       01 redefines MyPointer.
           02 MyPointerae9be588 pic S9(05) comp-5.
                            
       01 n PIC 9.
       procedure division.
      *    SET MyPointer UP BY n
           ADD n to MyPointerae9be588
                                
      *    SET MyPointer UP BY 1
           ADD 1 to MyPointerae9be588
                                
      *    SET MyPointer DOWN BY n
           SUBTRACT n to MyPointerae9be588
                                  
      *    SET MyPointer DOWN BY 1
           SUBTRACT 1 to MyPointerae9be588
                                  
           .
       end program TCSAMPLE8.
```

fm-117:
It appears that only skeletons need to be modified. See skeleton named "POINTER.INCREMENT" in skeleton.xml.

PR #1564, opened 2019-11-27
WI #1559 Fix code for decrementing a pointer

LANOYDO:


CLI/src/config/skeletons.xml
```diff
@@ -109,21 +109,22 @@
...
+        }
+        }
+        else
+        {
+        item += (@Model.incrementDirection == "Up" ? "ADD " : "SUBTRACT ") + @Model.sender.ToString();
+        item += (@Model.incrementDirection == "Up" ? " to " : " from ") ;
+        foreach (var receiver in @Model.displayableReceivers)
+        item += receiver + ", ";
+        }
+        item = item.Remove(item.Length - 2);
```
fm-117: Indentation is used to have a somewhat more readable generated .cs file, so it would be better if you preserve the original style.
Also, here you can write C# code so you can factorize the test on the increment direction like this :
```
...
else
{
    if (@Model.incrementDirection == "Up")
    {
        item += "ADD " + @Model.sender.ToString() + " to ";
    }
    else
    {
        item += "SUBTRACT " + @Model.sender.ToString() + " from ";
    }
    foreach 
    ...
```

LANOYDO: Fix indentation
https://github.com/TypeCobolTeam/TypeCobol/pull/1564/commits/af7ec2ab236ede632a9fdbfddf7fc7c26735d840

PR #1568, opened 2019-12-03
WI 1513 log file path option

rooksdo:
[Issue 1513.docx](https://github.com/TypeCobolTeam/TypeCobol/files/3915827/Issue.1513.docx)

See attached document for modifications in CLI project concerning this issue.


TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -28,6 +28,10 @@ public class TypeCobolConfiguration
         public ExecutionStep ExecToStep = ExecutionStep.Generate; //Default value is Generate
         public string ErrorFile = null;
         public string skeletonPath = "";
+        public string LogFile = null;
+
+        //Log file name
+        public const string LogFileName = "TypeCobol.CLI.log";
```
smedilol: I'd rather name this const `DefaultLogFileName` so we can clearly understand its purpose.

CLI/src/CLI.cs
```diff
@@ -36,18 +36,14 @@ class CLI
...
             Console.WriteLine(debugLine);
-
-            TextWriter textWriter;
-            if (config.ErrorFile == null) textWriter = Console.Error;
-            else textWriter = File.CreateText(config.ErrorFile);
-            AbstractErrorWriter errorWriter;
-            if (config.IsErrorXML) errorWriter = new XMLWriter(textWriter);
-            else errorWriter = new ConsoleWriter(textWriter);
+            TextWriter textWriter = config.ErrorFile == null ?  Console.Error : File.CreateText(config.ErrorFile);
+            AbstractErrorWriter errorWriter = config.IsErrorXML ? (AbstractErrorWriter) new XMLWriter(textWriter) : new ConsoleWriter(textWriter);
```
smedilol: You can remove the cast to (AbstractErrorWriter)

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
rooksdo: Agreed.

CLI/src/CLI.cs
rooksdo: C# compiler doesn't allow for this ternary operation without cast to first return value, since XMLWriter and ConsoleWriter do not implicitly cast to the same object class.

Issue #1566, opened 2019-11-29
Random bug with TypeCobol.LanguageServer during generation


smedilol:
**Describe the bug**
Random bug during code generation.
![image](https://user-images.githubusercontent.com/13764482/69874345-06c54f80-12bb-11ea-9127-0e16636a0f02.png)

**To Reproduce**
No information from the user. It's random.

rooksdo:
Ok for merge as soon as possible. Tests runs are all ok.
Specific tests concerning concurrency will need to be run.

fm-117:
May not be fully resolved but improving the concurrency checks in workspace should help (see PR #1569).

Issue #1565, opened 2019-11-29
Document not recognized by TypeCobol.LanguageServer 


smedilol:
**Describe the bug**
Close RDZ.
Open RDZ.
The user gets the following message:
![image](https://user-images.githubusercontent.com/13764482/69874099-6ec76600-12ba-11ea-86ce-4daa4ccfddac.png)

Restart TypeCobol.LanguageServer doesn't help.
RDZ must be restarted.

Idea: maybe 2 instances of TypeCobol.LanguageServer are running.

rooksdo:
Ok for merge as soon as possible. Tests runs are all ok.
Specific tests concerning concurrency will need to be run.

fm-117:
May not be fully resolved but improving the concurrency checks in workspace should help (see PR #1569).

PR #1569, opened 2019-12-03
WI #1565 #1566 improve concurrency handling in LS Workspace

fm-117:
Both issues #1565 et #1566 are hard to reproduce and concern the collection of opened documents in the Workspace. They seem to be both linked with concurrency issues on this dictionary so I've improved the concurrency handling and protected the dictionary from accesses outside of the class.


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -42,13 +42,13 @@ public class Workspace
         private CopyWatcher _CopyWatcher;
         private System.Timers.Timer _semanticUpdaterTimer;
         private bool _timerDisabled;
-
+        private readonly Dictionary<Uri, DocumentContext> _openedDocuments;
```
mayanje: We should probably start using _Concurrent​Dictionary<Uri, DocumentContext>_ this will avoid to use a lock variable.

fm-117: I've considered using it at first, but it does not fit for all usages in the class. Sometimes it translates well (simple access can be replaced with corresponding method in `ConcurrentDictionary`), but for more complex operations like refreshing all opened documents, it is not well suited. Locking would still be necessary and so I prefer to leave as it is.

Issue #1313, opened 2019-03-15
Completion bug for output parameter with empty type


smedilol:
**Describe the bug**
NullReferenceException when you ask completion after `var1::` when var1 is an output typed parameter and when the type of var1 only contains unnamed children.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       identification division.
       program-id. MyPGM.
       environment division.
       configuration section.
       source-computer. ibm-370
      *                 with debugging mode
                        .
       special-names. decimal-point is comma.
       data division.
       working-storage section.
      
      
       01 Type1 typedef strict.
          05 pic X.
      
       01 var1 type type1.
      
       procedure division.
           call foo output var1::
           .
       declare foo  output b type type1.
       end-declare.
      
       end program MyPGM.
```
and/or steps to reproduce the behavior:
Ask completion after var1::

Note, the bug doesn't happens when var1 is in `input`.

**Expected behavior**
No error, but nothing should appear in the completion list.

smedilol:
Issue has been merged into develop.
We must manually report it into branch v1.4.1

PR #1573, opened 2019-12-05
WI #1313 Avoid completion bug for output parameter with empty type

LANOYDO:


TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -253,13 +253,13 @@ public static void OnNode(Node node)
...
 
-                    var callArgName = actual.StorageArea.ToString();
                     var found = node.GetDataDefinitionFromStorageAreaDictionary(actual.StorageArea);
-                   if (found == null)
+                    if (found == null)
                     {
                         continue;
                     }
 
+                    var callArgName = actual.StorageArea.ToString();
```
fm-117: Ok with this fix, but we also need to check that `searchedStorageArea` is not null within `GetDataDefinitionFromStorageAreaDictionary`.

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1573/commits/ada299ba51eb65b02b486e60ab6de3e72c064bcb

Issue #1586, opened 2019-12-11
#1539 sub-issue for Equals/GetHashCode implementation


rooksdo:
This issue is a subset of issue #1539 to be integrated in v1.4. Targeted classes to be modified for implemention of Equals/GetHashCode are shown below (with specified method to redefine):

Type | Equals | GetHashCode
-- | -- | --
TypeCobol.Compiler.CodeElements.CodeElement | O | X
TypeCobol.Compiler.CodeElements.Expressions.AbstractQualifiedName | O | O
TypeCobol.Compiler.CodeElements.SymbolInformation | O | O
TypeCobol.Compiler.Diagnostics.Diagnostic | O | X
TypeCobol.Compiler.Scanner.MultilineScanState | O | O
TypeCobol.Compiler.Scanner.Token | O | O
TypeCobol.LanguageServer.VsCodeProtocol.SignatureInformation | O | X
TypeCobol.Compiler.CodeElements.SymbolReference | X | X
TypeCobol.Compiler.Preprocessor.ImportedToken | X | X

And the following table adds information about further checks to consider for implemention

Type | Action
-- | --
TypeCobol.Compiler.CodeElements.CodeElement | Add GetHashCode and Iequatable, check present comparision algorithm
TypeCobol.Compiler.CodeElements.Expressions.AbstractQualifiedName | Seems manageable
TypeCobol.Compiler.CodeElements.SymbolInformation | Add IEquatable ?
TypeCobol.Compiler.Diagnostics.Diagnostic | Add GetHashCode and Iequatable, check present comparision algorithm
TypeCobol.Compiler.Scanner.MultilineScanState | Add IEquatable ?
TypeCobol.Compiler.Scanner.Token | Add IEquatable ?
TypeCobol.LanguageServer.VsCodeProtocol.SignatureInformation | Add GetHashCode and Iequatable, check present comparision algorithm
TypeCobol.Compiler.CodeElements.SymbolReference | Check whether redefining equality is necessary?
TypeCobol.Compiler.Preprocessor.ImportedToken | CopyDirective needs to be checked in order to distinguish between native   and imported token

smedilol:
TypeDefinition was ok in PR #1555, why remove it from this issue ?

rooksdo:
[...]

Issue #204, opened 2016-04-04
CodeElement.get_Column out of range


wiztigers:
Since #196, parsing of sample file `IPCTPPRT.PGM` fails with a `System.ArgumentOutOfRangeException` thrown in `CodeElement.get_Column`.
`CodeElement.ConsumedTokens` must be `null` or empty ...

wiztigers:
I confirm this error wasn't present before meging the PR #196.

To assert this, one can checkout revision d9971b1f0f169970706f728a5d57680ca2f2f49c and run the sample parsing after applying the following patch (to remove another error in `MoveStatement.cs` that was fixed in 16864b9efdacef7e03fc6be459e3c3c3369669e5) :
```git
diff --git a/TypeCobol/Compiler/CodeElements/Statement/MoveStatement.cs b/TypeCobol/Compiler/CodeElements/Statement/MoveStatement.cs
index 5903fa7..62cc610 100644
--- a/TypeCobol/Compiler/CodeElements/Statement/MoveStatement.cs
+++ b/TypeCobol/Compiler/CodeElements/Statement/MoveStatement.cs
@@ -53,13 +53,21 @@ namespace TypeCobol.Compiler.CodeElements
         public Dictionary<QualifiedName,bool> Symbols {
             get {
                 var symbols = new Dictionary<QualifiedName,bool>();
-                if (Sending is Identifier)
-                    symbols.Add(((Identifier)Sending).Name, false);
-                foreach (var identifier in Receiving)
-                    symbols.Add(identifier.Name, true);
+                               var name = asUserDefinedName(Sending);
+                               if (name != null) symbols.Add(name, false);
+                foreach (var identifier in Receiving) {
+                                       name = asUserDefinedName(identifier);
+                                       if (name != null) symbols.Add(name, false);
+                               }
                 return symbols;
             }
             private set { throw new System.InvalidOperationException(); }
         }
+               private QualifiedName asUserDefinedName(Expression expression) {
+                       if (!(expression is Identifier)) return null;
+                       if (expression is SpecialRegister) return null;
+                       if (expression is FunctionReference) return null;
+                       return ((Identifier)expression).Name;
+               }
     }
 }

```

wiztigers:
@laurentprudhon, thanks in advance for your analysis.
I can fix this later, but I've must address some other tickets before.

wiztigers:
I confirm that `IPCTPPRT.PGM` is the only file in error in our production sample.

[...]

PR #1587, opened 2019-12-12
WI #1586 Sub-issue for Equals/GetHashCode implementation

LANOYDO:


TypeCobol/Compiler/CodeElements/Expressions/QualifiedName.cs
```diff
@@ -85,26 +85,28 @@ public int IndexOf(string item)
...
-            hash = hash * 7 + Count.GetHashCode();
-            foreach (string part in this)
-                hash = hash * 7 + part.GetHashCode();
-            return hash;
+            unchecked
+            {
+                int hash = 13;
+                hash = (hash * 397) ^ Count;
+                foreach (string part in this)
+                    hash = (hash * 397) + part.GetHashCode();
```
smedilol: Why `+` and not `^`  ?

TypeCobol/Compiler/CodeElements/CodeElement.cs
```diff
@@ -157,16 +159,44 @@ public override string ToString()
...
+
+            if (ConsumedTokens != null && codeElement.ConsumedTokens != null)
+            {
+                if (ConsumedTokens.Count > 0 && codeElement.ConsumedTokens.Count > 0)
+                {
+                    return Type == codeElement.Type && ConsumedTokens[0].Equals(codeElement.ConsumedTokens[0]);
+                }
+                else
+                {
+                    return Type == codeElement.Type && Object.ReferenceEquals(ConsumedTokens, codeElement.ConsumedTokens);
```
smedilol: If we reach this line, it means `this` and codeElement are 2 differents instances. 
Their ConsumedTokens contains no elements.
And their ConsumedTokens properties point to the same list. 

I wonder if it can happens.

I've tested to throw an exception and our tests run fine.

TypeCobol.LanguageServer/VsCodeProtocol/Signature Help/SignatureInformation.cs
```diff
@@ -41,14 +42,37 @@ public SignatureInformation(string label, string documentation, params Parameter
...
+            {
+                var hashCode = 13;
+                hashCode = (hashCode * 397) ^ label.GetHashCode();
+                var docHashCode = documentation?.GetHashCode() ?? 0;
+                hashCode = (hashCode * 397) ^ docHashCode;
+                if (parameters != null)
+                {
+                    foreach (var param in parameters)
+                    {
+                        hashCode = (hashCode * 397) ^ param.GetHashCode();
```
fm-117: You should combine values from parameter itself because in the Equals method parameters are compared using their content (label and documentation) and here you are using the hashcode derived from the reference of the parameter object.

TypeCobol/Compiler/CodeElements/CodeElement.cs
fm-117: About the comparison of CodeElements with no ConsumedToken, see issue #204. Properties based on ConsumedTokens collections have been modified to allow for null value of the collection.

TypeCobol/Compiler/CodeElements/Expressions/QualifiedName.cs
LANOYDO: fixed by using ^.
https://github.com/TypeCobolTeam/TypeCobol/pull/1587/commits/1be2e096b4f27bb53cd71184aacdcd855f430e29

TypeCobol/Compiler/CodeElements/CodeElement.cs
LANOYDO: Equals, if ConsumedTokens are not both not_null, returns false;
Equals, if counts of ConsumedTokens are different, returns false.
Equals, if counts of ConsumedTokens are the same and superior to 0, returns equality of Type and ConsumedTokens[0].
Equals, if counts of ConsumedTokens are both equal to 0, throws an exception.
https://github.com/TypeCobolTeam/TypeCobol/pull/1587/commits/1be2e096b4f27bb53cd71184aacdcd855f430e29

Issue #1590, opened 2019-12-12
Hover on procedure name fires error message


rooksdo:
### Description
Pressing F3 key on procedure name fires an error message, as in the example below:
```cobol
       declare procedure GetPersonById private{cursor here}
               input  PersonId type Date.
       Procedure division.
      
           goback
           .
       end-declare.
```

Stacktrace:
![image](https://user-images.githubusercontent.com/13764482/70780924-f1095d00-1d85-11ea-94ea-519d89da7545.png)

You can also try to use "hover" on an undeclared variable:
```cobol
            move a to xxxxxxxxxxxxx
```

### Solution
Shield against null value (see onHover method in TypeCobolServer.cs)

smedilol:
@LANOYDO I've updated the issue with the stacktrace

smedilol:
@LANOYDO the move statement doesn't seems to reproduce the bug.
But I can reproduce it with a set:
```cobol
            set xxxxxxxxxxxx to true
```

Issue #1579, opened 2019-12-05
CTRL+SPACE triggered unexpected error message


rooksdo:
**Describe the bug**
CTRL+SPACE (completion bind key) triggered unexpected error message

![image](https://user-images.githubusercontent.com/57907033/70260435-8efe9580-1790-11ea-8d65-f97fb4583281.png)

**To Reproduce**
Happened only once (see call trace for origin)

**Expected behavior**
Default behavior expected from CTRL+SPACE (a.k.a completion)

PR #1595, opened 2019-12-13
WI #1579 #1590 Fix hover and completion exceptions

LANOYDO:
see issues https://github.com/TypeCobolTeam/TypeCobol/issues/1579 and https://github.com/TypeCobolTeam/TypeCobol/issues/1590

### DIFF
--- a/TypeCobol/Compiler/Nodes/Node.cs
+++ b/TypeCobol/Compiler/Nodes/Node.cs
@@ -821,6 +821,7 @@ namespace TypeCobol.Compiler.Nodes {
         /// <returns>Correpsonding DataDefinition</returns>
         public DataDefinition GetDataDefinitionFromStorageAreaDictionary(StorageArea searchedStorageArea, bool? isReadDataDefiniton=null)
         {
+            if (searchedStorageArea == null) return null;
             DataDefinition searchedElem = null;
             if (isReadDataDefiniton == null)
             {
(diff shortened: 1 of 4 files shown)


TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -821,6 +821,7 @@ private void Dump(StringBuilder str, int i)
         /// <returns>Correpsonding DataDefinition</returns>
         public DataDefinition GetDataDefinitionFromStorageAreaDictionary(StorageArea searchedStorageArea, bool? isReadDataDefiniton=null)
```
smedilol: Can you add attribute `[CanBeNull]` before StorageArea searchedStorageArea ?
This way Resharper will know that the value can be null.

TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -848,6 +849,7 @@ public DataDefinition GetDataDefinitionFromStorageAreaDictionary(StorageArea sea
         
         public DataDefinition GetDataDefinitionForQualifiedName(QualifiedName qualifiedName, bool? isReadDictionary=null)
```
smedilol: Same here:
Can you add attribute [CanBeNull] before QualifiedName qualifiedName ?
This way Resharper will know that the value can be null.

TypeCobol/Compiler/Nodes/Node.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1595/commits/59139c835bd4751fb47886a0d3305206f1272b62

TypeCobol/Compiler/Nodes/Node.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1595/commits/59139c835bd4751fb47886a0d3305206f1272b62

Issue #783, opened 2018-01-10
Move : completion after a special keyword doesn't work


smedilol:
Completion after this kind of code doesn't work:
```cobol
move space to 
move quote to 
```

smedilol:
We need to discuss about which type are compatible with things like `space`, `quote`, ...

smedilol:
This issue can be started for quote and space

fm-117:
Related to #1541. See if the updated code also covers those cases.

PR #1596, opened 2019-12-17
WI #783 Complete after special keyword space quote

LANOYDO:


TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -678,7 +681,7 @@ public static IEnumerable<CompletionItem> GetCompletionForOf(FileCompiler fileCo
             {
                 case TokenType.ADDRESS:
                 {
-                    var contextToken = tokensUntilCursor?.Skip(2).FirstOrDefault(); //Try to get the token that may define the completion context
+                    var contextToken = tokensUntilCursor.Skip(2).FirstOrDefault(); //Try to get the token that may define the completion context
```
smedilol: The `?` after tokensUntilCursor has disappear. It shouldn't.

TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -587,11 +587,14 @@ public static IEnumerable<CompletionItem> GetCompletionForTo(FileCompiler fileCo
                     }
                 }
 
-                if (sendingVar?.RepeatedCharacterValue?.Token != null &&
-                    (sendingVar.RepeatedCharacterValue.Token.TokenType == TokenType.SPACE
-                     || sendingVar.RepeatedCharacterValue.Token.TokenType == TokenType.SPACES))
+                var tokenType = sendingVar?.RepeatedCharacterValue?.Token?.TokenType;
+                if (tokenType != null)
```
fm-117: At first I did not understand how this test would work since `TokenType` is a ValueType (enum) but in fact here the variable `tokenType` is of type `Nullable<TokenType>` so this test is equivalent to `if (tokenType.HasValue)`. I think this is clearer than comparing to null but it is ok.

TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
smedilol: Ok fine.

Issue #1248, opened 2019-01-28
string on two line with continuation line


maxime645:
**Describe the bug**
The way we allow the user to write a string on two lines with continuation line char is wrong and won’t compile.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)

The following code is recognized as a correct way to write a string on two lines.
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       data division.
       working-storage section.
       01 foo pic X(50).
       PROCEDURE DIVISION.
      
             move "CHAMPS 082 A 085 INTERDITS EN OUVERTURE"
      -        ""Texte supplémentaire"
      -        to foo
      
           goback
           .
       END PROGRAM DVZF0OSM.
```

If there is only one " at line 9, the diagnostic suggests to add a second ".

**Expected behavior**
The diagnostic should suggest to remove the final " at superior line and forbid the use of 
the code above.

fm-117:
In fact this syntax is valid and is used in NIST test `NC215A.cbl` ([at line 50](https://github.com/TypeCobolTeam/TypeCobol/blob/d6a78b3ff7fc618d84b005c89a01e9cbe495160e/TypeCobol.Test/ThirdParty/Nist/NC215A.cbl#L49)). This is combination of a continuation line and escaped double-quote within a string literal delimited by double quotes. (!) To escape the `"` character in a string literal delimited by double-quotes, the character has to be doubled. So here the line break happens right between the two consecutive double-quotes characters and therefore is valid.

This is supported by IBM compiler _BUT_ it is reported as a warning in RDZ COBOL editor by real-time compilation.

We should close this issue and not change the current behavior.

rooksdo:
See also #1594

smedilol:
I've tested with IBM compiler and this syntax is not valid.
Real-time syntax check of RDZ is not accurate.

You need to test with Remote Syntax check or compile the program.

The example from NIST test `NC215A.cbl` ([at line 50](https://github.com/TypeCobolTeam/TypeCobol/blob/d6a78b3ff7fc618d84b005c89a01e9cbe495160e/TypeCobol.Test/ThirdParty/Nist/NC215A.cbl#L49)) is not the same.
In Nist, it's in  ENVIRONMENT DIVISION / CONFIGURATION SECTION for an alphabet statement.

Here's all errors reported when I compile it:
```cobol
       IDENTIFICATION DIVISION.
[...]

PR #1598, opened 2019-12-17
WI #1248 Check validity of continuation line

LANOYDO:


fm-117:
Ok to improve the error message but with the proposed code, the invalid syntax described in the issue will still be accepted. We should correct the parser so as to reject invalid continuation lines and suggest proper syntax in the error message.

fm-117:
Closed as issue is no longer relevant.

Issue #1539, opened 2019-11-12
Implement Equals/GetHashCode activate ruleset and implement IEquatable<T>


smedilol:
**Describe the bug**
Some class implement the Equals methode but not GetHashCode().
It can causes problem especially when used in Dictionnary.

In all cases, review or Implement Equals() and GetHashCode() at least in:
- ImportedToken
- SymbolReference
- StorageArea
- CodeElement
- Node
- TypeDefinition

Activate ruleset about Equals() and GetHashCode() to find other missing.

For performance reason:
use ReferenceEquals() at the beggining of Equals() method because in the common case we retrieve reference to the same instance.
Also implement IEquatable<T>

fm-117:
Here is a template code for the Equals/GetHashCode methods :

```csharp
class SomeObject : IEquatable<SomeObject>
{
    public bool Equals(SomeObject other)
    {
        // other can be null if called directly, it is not null when called from Equals(SomeObject)
        if (object.ReferenceEquals(null, other)) return false;

        // optimization : use reference equality first
        if (object.ReferenceEquals(this, other)) return true;

        /*
         * NOTE : operator == can be used instead of object.ReferenceEquals only if it is not redefined.
         * This is already a rule in MS ruleset (no equality operator overloading in reference types).
         * To be clearer and safer, we should use object.ReferenceEquals.
         */

        // Same type, not null, compare constitutive constants from the object.
        throw new NotImplementedException();
    }

    public override bool Equals(object obj)
    {
        if (obj is SomeObject someObject)
        {
            return Equals(someObject);
        }

        // null or different type
        return false;
    }

    public override int GetHashCode()
    {
        // Combine contitutive constants from the object into a hash function.
        // The same data used in Equals(SomeObject) must be used in the hash and both methods should be updated when the object changes.
        // primes + bit XOR are recommended
        throw new NotImplementedException();
    }
}
```

rooksdo:
Follow up in issue #1588

PR #1600, opened 2019-12-18
1588 review equals get hash code

LANOYDO:
This is the follow up of issue https://github.com/TypeCobolTeam/TypeCobol/issues/1539
- Last implementations,
- activation equals/gethshcode rules as errors.


TypeCobol/Compiler/CodeElements/CodeElement.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
fm-117: Not required, no other code modification in this file ??

TypeCobol/Compiler/CodeElements/CodeElement.cs
```diff
@@ -24,6 +25,7 @@ protected CodeElement(CodeElementType type)
             SymbolInformationForTokens = new Dictionary<Token, SymbolInformation>();
         }
 
+
```
fm-117: Why ?

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -23,18 +24,34 @@ public class DataType : ICobolLanguageLevel, IVisitable {
...
+            {
+                var hashCode = 13;
+                hashCode = (hashCode * 397) ^ Name.GetHashCode();
+
+                return hashCode;
+            }
+        }
 
-		public static bool operator ==(DataType x, DataType y)
+        public static bool operator ==(DataType x, DataType y)
```
fm-117: Isn't that declaration supposed to raise an error since we are redefining equality operator on a reference type ? (rule CA1046)

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -23,18 +24,34 @@ public class DataType : ICobolLanguageLevel, IVisitable {
...
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = 13;
+                hashCode = (hashCode * 397) ^ Name.GetHashCode();
+
+                return hashCode;
+            }
+        }
```
fm-117: The original comparison mechanism used the CobolLanguageLevel because basically the idea was to use reference equality for Cobol 85 DataTypes and then compare names for other types.
I think we should use all three fields in Equals/GetHashCode implementations just to be on the safe side.

TypeCobol/Compiler/CodeElements/Expressions/StorageArea.cs
```diff
@@ -110,17 +110,17 @@ public enum StorageAreaKind
 	/// defined special registers (see list in a comment just below).
 	/// </summary>
 	public class DataOrConditionStorageArea: StorageArea {
```
fm-117: You did not change the Equals/GetHashCode for StorageArea.

TypeCobol/Compiler/CodeElements/Header/FunctionDeclarationHeader.cs
```diff
@@ -232,23 +233,27 @@ public class ParametersProfile: CodeElement, ParameterList, IVisitable
...
-		    if (!okay) return false;
-		    okay = AreEqual(InoutParameters, o.InoutParameters);
-		    if (!okay) return false;
-		    okay = AreEqual(OutputParameters, o.OutputParameters);
-		    return okay;
-	    }
+	    public override bool Equals(object other)
+        {
+            return Equals(other as ParametersProfile);
+        }
```
fm-117: Some spaces/tabs indent mix

TypeCobol/Compiler/CodeElements/Header/FunctionDeclarationHeader.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Text;
+﻿using System;
+using System.Text;
```
fm-117: Not needed

TypeCobol/Compiler/Nodes/ParametersProfileNode.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace TypeCobol.Compiler.Nodes
 {
-    public class ParametersProfileNode : GenericNode<ParametersProfile>, ParameterList
+    public class ParametersProfileNode : GenericNode<ParametersProfile>, ParameterList, IEquatable<ParametersProfileNode>
```
fm-117: Equals and GetHashCode could be forwarded to ParametersProfile to avoid duplicate code.

TypeCobol/Compiler/Scanner/Token.cs
```diff
@@ -16,7 +16,7 @@ public class Token : Antlr4.Runtime.IToken, IVisitable, IEquatable<Token>
         private int startIndex;
         private int stopIndex;
 
-		/// <summary>Empty constructor for mock.</summary>
+        /// <summary>Empty constructor for mock.</summary>
```
fm-117: Ok but why not changing the next line which also has tabs ?

LANOYDO: because i see no indent difference when i use Visual Studio editor.

TypeCobol/Compiler/CodeElements/Header/FunctionDeclarationHeader.cs
LANOYDO: This namespace is used by IEquatable.

TypeCobol/Compiler/CodeElements/CodeElement.cs
LANOYDO: This namespace is used by IEquatable.

TypeCobol/Compiler/CodeElements/CodeElement.cs
LANOYDO: Initialise the property.

TypeCobol/Compiler/CodeElements/Expressions/StorageArea.cs
LANOYDO: In StorageArea, there is no Equals/GetHashCode.

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
LANOYDO: When operator equals is overloaded and Equals/GetHashCode are not overrided, CA1046 fires an exception: 'DataType' should not overload the equality operator.
CA1046 is active when Equals/GetHashCode are not overrided.

TypeCobol/Compiler/Nodes/ParametersProfileNode.cs
LANOYDO: done.
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8bb22a0bed7c51638872bc38edfbcb12c9e2ab34

TypeCobol/Compiler/CodeElements/Header/FunctionDeclarationHeader.cs
LANOYDO: done.
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8bb22a0bed7c51638872bc38edfbcb12c9e2ab34

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
LANOYDO: done.
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8bb22a0bed7c51638872bc38edfbcb12c9e2ab34

TypeCobol/Compiler/CodeElements/CodeElement.cs
LANOYDO: One "using System;" has been deleted.
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8bb22a0bed7c51638872bc38edfbcb12c9e2ab34

LANOYDO:
> Missing in this PR :
> 
> * TypeCobol.Compiler.CodeElements.StorageArea
> * TypeCobol.Compiler.Nodes.Node
> 
> In these classes, there are no Equals/GetHashCode methods. No need to override.

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
fm-117: Ok but we should either :
- remove these two operator overloads as this is the only case of operator overloading in the solution and can be confusing. Note that the usages should be all updated.
- or forward the implementation to Equals method to avoid code duplication.
Wait for @smedilol view on this question.

TypeCobol/Compiler/CodeElements/Expressions/StorageArea.cs
fm-117: Yes but that's the point. @smedilol asked for a custom equality comparison mechanism for both StorageArea and Node so the idea is to check whether it is useless or not and to decide which implementation to use.

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -23,18 +24,37 @@ public class DataType : ICobolLanguageLevel, IVisitable {
...
```
fm-117: CobolLanguageLevel and RestrictionLevel should be used in comparison along with Name.

TypeCobol/Compiler/CodeElements/Expressions/StorageArea.cs
```diff
@@ -67,6 +67,24 @@ public override string ToString()
...
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (13 * 397) ^ base.GetHashCode();
+            }
+        }
+
```
fm-117: This is useless. This is equivalent to use the default implementations from the framework. We should discuss if StorageArea actually needs a custom equality comparison.

TypeCobol/Compiler/CodeElements/Header/FunctionDeclarationHeader.cs
```diff
@@ -232,62 +233,71 @@ public class ParametersProfile: CodeElement, ParameterList, IVisitable
...
-		    int hash = 17;
-		    foreach (var p in Parameters) hash = hash * 23 + p.GetHashCode();
-		    return hash;
-	    }
+        private static bool AreEqual(IList<ParameterDescriptionEntry> mine, IList<ParameterDescriptionEntry> hers)
+        {
+            if (mine.Count != hers.Count) return false;
+            for (int c = 0; c < mine.Count; c++) {
+	            if (!mine[c].Equals(hers[c])) return false;
+	            if (!mine[c].Equals(hers[c])) return false;
```
fm-117: Duplicate lines

TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -717,6 +717,19 @@ public Node Get(string uri, int startIndex)
...
+            return Object.ReferenceEquals(this, obj as Node);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (13 * 397) ^ base.GetHashCode();
+            }
+        }
```
fm-117: Same remark as for StorageArea. No need to add code that does the same as the framework.

TypeCobol/Compiler/Nodes/ParametersProfileNode.cs
```diff
@@ -40,43 +40,6 @@ public IList<ParameterDescription> Parameters
...
-            }
-            return true;
-        }
-
-        public override int GetHashCode()
-        {
-            int hash = 17;
-            foreach (var p in Parameters) hash = hash * 23 + p.GetHashCode();
-            return hash;
-        }
```
fm-117: Why delete the methods ?? The idea here is to have the IEquatable/Equals/GetHashCode and forward the implementation to the CodeElement because we already have an equality comparison mechanism for the ParametersProfile class.

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8b7686fa9f02c00e2a7150589c65d7fd66b787c8

TypeCobol/Compiler/CodeElements/Expressions/StorageArea.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8b7686fa9f02c00e2a7150589c65d7fd66b787c8

TypeCobol/Compiler/CodeElements/Header/FunctionDeclarationHeader.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8b7686fa9f02c00e2a7150589c65d7fd66b787c8

TypeCobol/Compiler/Nodes/Node.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8b7686fa9f02c00e2a7150589c65d7fd66b787c8

TypeCobol/Compiler/Nodes/ParametersProfileNode.cs
LANOYDO: It's right; it's fixed.
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/8b7686fa9f02c00e2a7150589c65d7fd66b787c8

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -9,8 +9,9 @@
 
 namespace TypeCobol.Compiler.CodeElements
 {
-	public class DataType : ICobolLanguageLevel, IVisitable {
-		public string Name { get; private set; }
+	public class DataType : ICobolLanguageLevel, IVisitable, IEquatable<DataType>
+    {
```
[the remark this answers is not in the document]
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/f938281db0e24b2c886be36dac2baf758d00710f

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -23,34 +24,56 @@ public class DataType : ICobolLanguageLevel, IVisitable {
...
+        private static bool Equals([CanBeNull]DataType x, [CanBeNull]DataType y)
         {
-            //Data instance for Cobol85 are unique so we can compare reference
             if (Object.ReferenceEquals(x, y)) return true;
-            if (x?.CobolLanguageLevel == CobolLanguageLevel.Cobol85 ||
-                y?.CobolLanguageLevel == CobolLanguageLevel.Cobol85) return false;
+            if (Object.ReferenceEquals(x, null)) return false;
+            if (Object.ReferenceEquals(null, y)) return false;
+            
+            bool isEqual = string.Equals(x.Name, y.Name, StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/f938281db0e24b2c886be36dac2baf758d00710f

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
rooksdo: OK with @smedilol.

TypeCobol/Compiler/CodeElements/CodeElement.cs
```diff
@@ -1,5 +1,5 @@
-using Antlr4.Runtime;
 using System;
+using Antlr4.Runtime;
```
fm-117: In the end, it appears this file hasn't been modified so could you revert to the original file from develop.

LANOYDO: File TypeCobol/Compiler/CodeElements/CodeElement.cs is identical identical from develop branch one
https://github.com/TypeCobolTeam/TypeCobol/pull/1600/commits/7b1c8285f6ae818069349c70ab185fa012fdd39a

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -9,58 +9,79 @@
...
+            return astVisitor.Visit(this);
+        }
 
-	    public override bool Equals(object obj) {
-			var other = obj as DataType;
-			if (other == null) return false;
-			return other == this;
-		}
+        public override bool Equals(object obj) {
+            return Equals(this, obj as DataType);
```
smedilol: This implementation of Equals doesn't respects our guideline :
https://github.com/TypeCobolTeam/TypeCobol/issues/1539


Can you adapt the code,

LANOYDO: It's already done.
public override bool Equals(object obj) {
            return Equals(this, obj as DataType);
        }

        public bool Equals(DataType dataType)
        {
            return Equals(this, dataType);
        }

        private static bool Equals([CanBeNull]DataType x, [CanBeNull]DataType y)
        {
            if (Object.ReferenceEquals(x, y)) return true;
            if (Object.ReferenceEquals(x, null)) return false;
            if (Object.ReferenceEquals(null, y)) return false;

            return string.Equals(x.Name, y.Name, StringComparison.OrdinalIgnoreCase)
                   && x.RestrictionLevel.Equals(y.RestrictionLevel)
                   && x.CobolLanguageLevel.Equals(y.CobolLanguageLevel);
        }

        public override int GetHashCode()
        {
            unchecked
            {
                var hashCode = 13;
                hashCode = (hashCode * 397) ^ Name.GetHashCode();
                hashCode = (hashCode * 397) ^ RestrictionLevel.GetHashCode();
                hashCode = (hashCode * 397) ^ CobolLanguageLevel.GetHashCode();

                return hashCode;
            }
        }

PR #1602, opened 2019-12-20
WI #1592 Show line of copy statement if error(s) is(are) inside copy

### LINKED ISSUE #1592: Copy errors shown in TypeCobol program file
### Description
Copy errors are shown in TypeCobol source program file at the same position (ex: error on line 5 in copy shows an error on line 5 in program).

LANOYDO:
Fixes #1592.


TypeCobol/Compiler/AntlrUtils/ParserDiagnosticErrorListener.cs
```diff
@@ -65,10 +65,15 @@ public class ParserDiagnostic : Diagnostic
             // TO DO - IMPORTANT : this is the INITIAL line number, and not the CURRENT line number
             // This is enough to pass all unit tests, but will return false informations in real usage !
             // for real line number, use a Snapshot
-            if (Line < 0 && OffendingSymbol != null)
+            if (OffendingSymbol != null)
             {
-                CodeElement e = OffendingSymbol as CodeElement;
-                if (e != null && e.ConsumedTokens.Count > 0) Line = e.ConsumedTokens[0].Line;
+                var copyToken = (offendingSymbol as ImportedToken)?.CopyDirective?.COPYToken;
+                if (copyToken != null) Line = copyToken.Line;
```
rooksdo: Why use offendingSymbol when OffendingSymbol was used for non-null test?

TypeCobol/Compiler/CupParser/NodeBuilder/CupParserTypeCobolProgramDiagnosticErrorReporter.cs
```diff
@@ -169,10 +169,18 @@ public class CupParserDiagnostic : Diagnostic
             // TO DO - IMPORTANT : this is the INITIAL line number, and not the CURRENT line number
             // This is enough to pass all unit tests, but will return false informations in real usage !
             // for real line number, use a Snapshot
-            if (Line < 0 && OffendingSymbol != null)
+            CodeElement e = offendingSymbol?.value as CodeElement;
```
fm-117: The idea here is to find accurate information for the parser error location, so a pseudo-code would be :
```csharp
if (Line < 0 && OffendingSymbol != null)
{
    if (offendingSymbol is inside copy)
    {
        //Retrieve copy token and use its line as location
    }
    else
    {
        //Fallback to first consumed token location
        ...
    }
}
```

To determine whether the offending symbol is inside a copy or not, you'll have to use two different methods because in ANTLR, we are dealing with tokens and in CUP we have codeElements so the code will be slightly different between the two classes.

TypeCobol.Test/Parser/Programs/Cobol2002/TypedefInCopy/typedef-strong.rdzMix-EI.txt
```diff
@@ -1,9 +1,9 @@
-Line 1[27,33] <27, Error, Syntax> - Syntax error : Custom EI rule : Only TYPEDEF STRICT is allowed.
        IDENTIFICATION DIVISION.
        PROGRAM-ID. MainProgram.
        data division.
        working-storage section.
 Line 5[27,33] <27, Error, Syntax> - Syntax error : Custom EI rule : Only TYPEDEF STRICT is allowed.
+Line 5[27,33] <27, Error, Syntax> - Syntax error : Custom EI rule : Only TYPEDEF STRICT is allowed.
```
fm-117: Ideally, we should avoid duplicate errors. I guess that what is happening here is that we get one error from the copy and one from the usage of the type. So to avoid that, the checker should only raise one error but not both.

TypeCobol/Compiler/AntlrUtils/ParserDiagnosticErrorListener.cs
LANOYDO: done : use of only OffendingSymbol.
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/c75e93cbbbe150f7b2561c1696660261a82adfbe

TypeCobol/Compiler/CupParser/NodeBuilder/CupParserTypeCobolProgramDiagnosticErrorReporter.cs
LANOYDO: It's right.
Modification has been made in these 2 classes by removing code about line is < 0 because it never happens (result of property IToken.Line or CodeElement.Line).
Line is modified when a copy is detected.
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/c75e93cbbbe150f7b2561c1696660261a82adfbe

TypeCobol.Test/Parser/Programs/Cobol2002/TypedefInCopy/typedef-strong.rdzMix-EI.txt
LANOYDO: done: adding to  the message "Syntax error : Custom EI rule : Only TYPDEF STRICT is allowed" "in definition of type <name_of_the_typedef>.".
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/c75e93cbbbe150f7b2561c1696660261a82adfbe

TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
```diff
@@ -49,7 +49,7 @@ public static void CheckTypedef(DataTypeDescriptionEntry typedef, CodeElementsPa
 #if EUROINFO_LEGACY_TYPEDEF
             if (typedef.RestrictionLevel != RestrictionLevel.STRICT)
             {
-                string message = "Custom EI rule : Only TYPEDEF STRICT is allowed.";
+                string message = $"Custom EI rule : Only TYPEDEF STRICT is allowed in definition of type {typedef.DataName.Name}.";
```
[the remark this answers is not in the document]
LANOYDO: Reason is to avoid duplicate errors; see the previous conversation:
Ideally, we should avoid duplicate errors. I guess that what is happening here is that we get one error from the copy and one from the usage of the type. So to avoid that, the checker should only raise one error but not both.
----adding to the message "Syntax error : Custom EI rule : Only TYPDEF STRICT is allowed" "in definition of type <name_of_the_typedef>.".

TypeCobol.Test/Parser/Programs/Cobol2002/TypedefInCopy/typedef-strong.rdzMix-EI.txt
smedilol: In fact, here we get 2 errors, because there are 2 typedef NOT strict.
To improve readability, I propose to add the name of the COPY inside the message.
Something like:
Line 5[27,33] <27, Error, Syntax> - Syntax error in XXXX.cpy : Custom EI rule : Only TYPEDEF STRICT is allowed.

LANOYDO: done; retrieve if exists copy name and add to messages in method Cobol2002Checker.CheckTypedef
example message "Syntax error : Custom EI rule : Only TYPEDEF STRICT is allowed in definition of type TypedefWeak1 in 'TypedefWeak' copy"
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/b3858d1f5a5d9ac4d5ab15c99198b4222fbf0e3d

TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
```diff
@@ -27,29 +28,33 @@ public static void CheckRedefines(DataRedefinesEntry redefines, CodeElementsPars
...
-                string message = "EXTERNAL clause cannot be specified with TYPEDEF clause";
+                string message = "EXTERNAL clause cannot be specified with TYPEDEF clause" + copyErrorMessage;
                 foreach (var external in context.externalClause())
                     DiagnosticUtils.AddError(typedef, message, external);
             }
 
             if (typedef.HasExplicitVisibility && typedef.IsGlobal)
             {
-                string message = $"GLOBAL clause cannot be specified with {typedef.Visibility.ToString().ToUpper()} access modifier";
+                string message = $"GLOBAL clause cannot be specified with {typedef.Visibility.ToString().ToUpper()} access modifier" + copyErrorMessage;
```
smedilol: From An API perspective, the COPY name shold be a separate parameter of `DiagnosticUtils.AddError`.
It will also allow to change the positon of the copy name inside the error message easily.

Or even better, can the COPY name could be retrieved automatically within the AddError method.

LANOYDO: done in AddError method.
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/96bff72e166deda2c10a40bb4a19f899b6c4dfdd

TypeCobol/Compiler/AntlrUtils/ParserDiagnosticErrorListener.cs
```diff
@@ -57,18 +57,18 @@ public override void SyntaxError(IRecognizer recognizer, IToken offendingSymbol,
...
+            var copyDirective = (OffendingSymbol as ImportedToken)?.CopyDirective;
+            if (copyDirective != null)
             {
-                CodeElement e = OffendingSymbol as CodeElement;
-                if (e != null && e.ConsumedTokens.Count > 0) Line = e.ConsumedTokens[0].Line;
+                Line = copyDirective.COPYToken.Line;
+                bool endsWithDot = Message.EndsWith(".");
+                if (endsWithDot) Message = Message.Substring(0, Message.Length - 1);
+                Message += $", in copy '{copyDirective.TextName}'";
+                if (endsWithDot) Message += ".";
```
fm-117: I think it is slightly better to have indication of the copy at the beginning of the message so the developer can quickly determine whether the error is in his own file or another. Somehing like : `Error in copy <copy-name> : <original message>`. This also simplifies the code : no need to care about dots.

TypeCobol/Compiler/AntlrUtils/ParserDiagnosticErrorListener.cs
```diff
@@ -57,18 +57,18 @@ public override void SyntaxError(IRecognizer recognizer, IToken offendingSymbol,
     public class ParserDiagnostic : Diagnostic
     {
 		public ParserDiagnostic(string message, IToken offendingSymbol, string ruleStack, MessageCode code = MessageCode.SyntaxErrorInParser, Exception exception = null) :
-			base(code, offendingSymbol == null ? -1 : offendingSymbol.Column, offendingSymbol == null ? -1 : (offendingSymbol.StopIndex < 0 ? -1 : (offendingSymbol.StopIndex+1)), offendingSymbol == null ? -1 : offendingSymbol.Line, message, exception)
+			base(code, offendingSymbol == null ? -1 : offendingSymbol.Column, offendingSymbol == null ? -1 : (offendingSymbol.StopIndex < 0 ? -1 : (offendingSymbol.StopIndex+1)), offendingSymbol == null ? -1 : offendingSymbol.Line, message, exception, offendingSymbol)
```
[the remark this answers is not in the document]
LANOYDO: removed
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/06cfe9438dc717ddb44511944de87226fd560015

TypeCobol/Compiler/AntlrUtils/ParserDiagnosticErrorListener.cs
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1602/commits/7c04857a0a5144ffc90df51a3c52657caa9fab40

TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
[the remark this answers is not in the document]
fm-117: Maybe use this error message : "type <type name> must be marked as STRICT." ?

Issue #1576, opened 2019-12-05
Completion on List indexed object shows List index as suggestion


rooksdo:
**Describe the bug**
Completion on a list structure type doesn't always show suggestions.

**To Reproduce**
In the main program (say PGM1.tcbl), a reference to a typedef dependency (say MyDependency.tcbl) is given as input
```cobol
DECLARE PROCEDURE Foo PUBLIC
           INPUT  ListObjects      TYPE MyDependency::ListObject
````
### Typedef definitions in MyDependency.tcbl

Definition of a public typedef named ListObject 
```cobol
01 ListObject typedef strict public.
   05 ListObjectGroupData type ObjectData occurs 200
                                          indexed by ListIdx.
```
This typedef uses another typedef named ObjectData defined as
```cobol
01 ObjectData typedef strict public.
   05 Data1                  PIC X(3).
```
The following statement suggests ListIdx as a completion suggestion:
```cobol
Perform varying index from 1 by 1 until
                  ListObjects::ListObjectGroupData::{cursor here} (index) equal space
```

**Expected behavior**
Completion should suggest Data1 as possible choice

rooksdo:
Fixes #1605

PR #1605, opened 2019-12-24
WI #1576 Modify completion on indexed list of typed element to show i…

LANOYDO:
…ts sub-element(s)


TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -424,14 +424,13 @@ public static IEnumerable<CompletionItem> GetCompletionForQualifiedName(Position
                     foreach (var variable in possibleVariables)
                     {
                         var children = new List<Node>();
+                        //if it's a typed variable, we have to search for children in the type
+                        var typeChildren = GetTypeChildren(node.SymbolTable, variable);
```
fm-117: Ok with the code but the call to GetTypeChildren may be expensive (use of SymbolTable) so it would be better to check whether the call is needed or not. The special case here are the arrays which are typed and also have children (indexes).

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1605/commits/a6ceab3fa390bd3cec1b2a4f4829f41abed0104e

TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -424,15 +424,18 @@ public static IEnumerable<CompletionItem> GetCompletionForQualifiedName(Position
                     foreach (var variable in possibleVariables)
                     {
                         var children = new List<Node>();
-                        if (variable.Children != null && variable.Children.Count > 0) //It's a variable with levels inside
-                            children.AddRange(variable.Children);
-                        else //It's a typed variable, we have to search for children in the type
+                        if (variable.Children?.All(c => c is IndexDefinition) ?? true)
```
smedilol: It works, but I think the code is hard to understand.
Plus this code (like the previous one) need to call the SymbolTable.
We know have cached variable.
If it's a typed varialble you can use `variable.TypeDefinition` to get the TypeDefinition.

I propose this code:
```C#
                        var children = new List<Node>();

                        //if it's a typed variable, propose 1st children of the type
                        if (variable.TypeDefinition != null) 
                        {
                            children.AddRange(variable.TypeDefinition.Children.Where(t => t.Name != null || t.Children.Where(u => u.Name != null) != null));
                        }
                        
                        //It's a variable with levels inside
                        //Note that Index are also children
                        if (variable.Children?.Count > 0) 
                            children.AddRange(variable.Children);

                        //No children => nothing to do
```

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1605/commits/89a9149e4842f4faff4b936987bcaf014a59a81d

TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -424,15 +424,19 @@ public static IEnumerable<CompletionItem> GetCompletionForQualifiedName(Position
...
-                            children.AddRange(variable.Children);
-                        else //It's a typed variable, we have to search for children in the type
+
+                        //if it's a typed variable, propose 1st children of the type
+                        if (variable.TypeDefinition != null)
                         {
-                            var typeChildren = GetTypeChildren(node.SymbolTable, variable);
-                            if (typeChildren != null)
-                                children.AddRange(typeChildren.Where(t => t.Name != null || t.Children.Where(u => u.Name != null) != null));
+                            children.AddRange(variable.TypeDefinition.Children.Where(t => t.Name != null || t.Children.Where(u => u.Name != null) != null));
```
fm-117: `t.Name != null || t.Children.Where(u => u.Name != null) != null` is always `true` because LINQ Where method always produces a non-null result. So in its current form, this line is equivalent to `children.AddRange(variable.TypeDefinition.Children);`.
Is this intended ? I know it was written originally but this is different now because you're not using the `GetTypeChildren` method so :
- before your modification : every node returned by GetTypeChildren were selected
- after your modification : every child from variable is selected

fm-117: So after reviewing the code of `GetTypeChildren`, I think it's ok. It is equivalent to selecting all children from the type, you can change the line to `children.AddRange(variable.TypeDefinition.Children);`

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1605/commits/99ff165773be105e3be39d1adc6750d3d026e678

Issue #1440, opened 2019-08-27
Completion on half written words


maxime645:
**Describe the bug**
The completion doesn't work fine when the type we try to complete is half written and the node not properly ended.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Callee.
      
       DATA DIVISION.
       working-storage section.
      
       01 STR.
           05 DATA1 PIC X.
           05 DATA2 TYPE B
           05 FILLER PIC X.
      
       PROCEDURE DIVISION.
      
       END PROGRAM Callee.
```
Try to complete variable DATA2 to "Bool".
When the B is written and we call autocompletion, it won't find anything. Yet, when nothing is written and call autocomplete, it gives all the possibilities.
After first analysis, there is no node associated to the line "05 DATA2 TYPE B"

**Expected behavior**
Autocompletion should return the compatible types.

fm-117:
This is not a completion bug but rather a parsing problem. In the given example :
- without the `B` after `TYPE`, CUP will produce all expected nodes
- with `B` after `TYPE`, CUP parser stops after DATA1 and no other nodes are produced (no DATA2, no PROCEDURE DIVISION, no END PROGRAM)

Without nodes for the matching token/codeelement, no completion can be performed.
It means that this issue is related to CUP and the way it handles errors and sometimes stops the whole process.

fm-117:
In fact this is more related to ANTLR which produces wrong CodeElements (then CUP has trouble building nodes with such input).

When reading tokens, ANTLR comes to an error at `05 DATA2 TYPE B` because there is no PeriodSeparator. The problem here is that the Scanner will produce an `IntegerLiteral` token instead of a `LevelNumber` for the `05` of the next line. From there, the parsing seems to be broken and as previously explained, the completion is broken as well. Note that the completion works as expected if we add a `.` after `TYPE B` or if there is no additional data declaration after `TYPE B` (because in both cases the next token will be correctly recognized).

[...]

PR #1610, opened 2020-01-03
WI #1593 Check completion for unknown values when using WHEN

### LINKED ISSUE #1593: Check completion for WHEN
**Description**
This issue is related to issue #887 and #1311 

**Task**
Check for completion results on unknown values and add missing test scenarios

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       data division.
       working-storage section.
        01 W-OPTION         PIC X.
           88 W-OPTION-1    Value '1'.
           88 W-OPTION-2    Value '2'.
       PROCEDURE DIVISION.
           evaluate true
               when W-OPTION-1 display "creation"
      *Case 1
      *OK Completion suggests W-OPTION, W-OPTION-1 and W-OPTION-2
               when
           end-evaluate.
      
           evaluate true
               when W-OPTION-1 perform creation
      *Case 2
      *KO Completion suggests CREATION, ERREUR and INIT-LIBRARY but should show
      *W-OPTION, W-OPTION-1 and W-OPTION-2
               when
           end-evaluate.
      
           goback.
       CREATION.
           display 'creation' .
       ERREUR.
           display 'error' .
       END PROGRAM Pgm.
```

**Rules for evaluate/when**
Inside an evaluate:
1. A `when` followed by another `when` can have no instruction.
2. The last `when` of an evaluate or if a `when` is followed by `when other`, then it must have at least one instruction
3. There can be 0 to 1 `when other`.
4. If present, the `when other` must be the last `when`

```cobol
           evaluate true
      * Ok when is followed by another when
               when W-OPTION-1
               when W-OPTION-2
                 display "Ok"
           end-evaluate.

           evaluate true
               when W-OPTION-1
                 display "test"
      * KO no instruction under this when
               when W-OPTION-2
           end-evaluate.

           evaluate true
               when W-OPTION-1
                 display "test"
      * KO no instruction under this when
               when W-OPTION-2
               when other
                 display "test"
           end-evaluate.

           evaluate true
               when W-OPTION-1
                 display "test"
               when W-OPTION-2
                 display "test"
      * KO no instruction under this when
               when other
           end-evaluate.

           evaluate true
               when W-OPTION-1
               when W-OPTION-2
                 display "test"
               when other
                 display "test"
           end-evaluate.

           evaluate true
      * KO when other must be last
               when other
                 display "test"
               when W-OPTION-1
               when W-OPTION-2
                 display "test"
           end-evaluate.

           evaluate true
               when W-OPTION-1
               when W-OPTION-2
                 display "test"
               when other
                 display "test"
      * KO more than 1 when other
               when other
                 display "test2"
           end-evaluate.


           evaluate true
      * KO when without condition but otherwise it's ok
      * to have no instruction
               when 
               when W-OPTION-2
                 display "test"
               when other
                 display "test"
           end-evaluate.
```

LANOYDO:
Fixes #1593.


TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -89,27 +89,73 @@ private IEnumerable<CodeElementWrapper> CodeElementFinder(FileCompiler fileCompi
             int linesCount = fileCompiler.CompilationResultsForProgram.ProgramClassDocumentSnapshot.PreviousStepSnapshot.Lines.Count;
             if (linesCount != 0 && lineIndex < linesCount)
             {
+                bool isFirstLine = true;
                 while (codeElements.Count == 0 && lineIndex >= 0)
                 {
                     var codeElementsLine =
-                        fileCompiler.CompilationResultsForProgram.ProgramClassDocumentSnapshot.PreviousStepSnapshot.Lines[lineIndex];
+                        fileCompiler.CompilationResultsForProgram.ProgramClassDocumentSnapshot.PreviousStepSnapshot
```
smedilol: Can you explain in why the problem happens for WHEN ?
Can the problem happens for other statements?

LANOYDO: You write a perform with when statements.
When you write only a new WHEN and Ctrl+Space, the completion searchs the nearest codeElement (it can be display, perform, ...).
The problem is that the WHEN is not identified as a codeElement but only as token.
Solution is to detect WHEN on the line of Ctrl+Space if there is no codeElement on this line.

smedilol: Ok, in this case, we should rather look at the grammar of WHEN to allow a WHEN without anything after it.

smedilol: So after your investigation and our discussions:
- We must change the grammar to accept a WHEN without anything else after. This ensure the WhenCondition codeElement is created
- It's also pretty sure we need to change the grammar of Evaluate/When/End evaluete (CUPS grammar) to accept a WHEN not followed by any statements.

By doing this, we'll have proper CodeElement and Node.

But this means our grammar is too permissive and we need to check the syntax manually.
@mayanje @fm-117 @LANOYDO @rooksdo  What do you think ?

I think it's better to change the grammar because we can then rely on CodeElement and Node.
Otherwise we would have to deal with tokens only in LanguageServer. That means try to do the job of the parser to deal with the token WHEN.

fm-117: This is somewhat close to #1440 where parsing is interrupted and no completion data can be found. I still think that the grammar should be as strict as possible but the recovery mechanism should be improved to allow such missing parts. Improving the parser recovery (ANTLR and CUPS) would be useful for all completion scenarios not only the WHEN case.
Easir said than done of course...

smedilol: I think we first need to investigate why for an incomplete move instruction we still have a move CodeElement and Node.

In following code, `move` will be detected and we even have an error because `var1` is not referenced.
```cobol
       IDENTIFICATION DIVISION.    
       PROGRAM-ID. DVZZBCO1.
       PROCEDURE DIVISION.
           move var1
           .

       END PROGRAM DVZZBCO1.
```

rooksdo:
We should test completion result for the following COBOL statements and compare with issue at hand : 

Empty move
```cob
move {expected variable or numeric/alphanumeric litteral}
```

Empty if
```cob
if {expected conditional statement}
```

Evaluate with missing when condition
```cob
evaluate true
   when {expected conditional statement}
end-evaluate
```

TypeCobol.LanguageServer/TypeCobolServer.cs
LANOYDO: move var1
.
sentence-0
        [[MoveStatement]] [12,15:move]<MOVE> --> [17,20:var1]<UserDefinedWord>
        end

move 
.
 sentence-0
        [[MoveStatement]] No Consumed Tokens
        [[MoveStatement]] [12,15:move]<MOVE> --> [12,15:move]<MOVE>
        end

-------- with non modified code
evaluate true
     when bool1
end-evaluate
.
 [[EvaluateStatement]] [12,19:evaluate]<EVALUATE> --> [21,24:true]<TRUE>


evaluate true
     when 
end-evaluate
.
sentence-0
        [[EvaluateStatement]] [12,19:evaluate]<EVALUATE> --> [21,24:true]<TRUE>

          end
        end

-------- with modified code (issue1593/PR1610)
evaluate true
     when bool1
end-evaluate
.
      [[EvaluateStatement]] [12,19:evaluate]<EVALUATE> --> [21,24:true]<TRUE>

        ?
          [[WhenCondition]] [17,20:when]<WHEN> --> [22,26:bool1]<UserDefinedWord>

        ?
          end
        end

evaluate true
     when 
end-evaluate
.
      [[EvaluateStatement]] [12,19:evaluate]<EVALUATE> --> [21,24:true]<TRUE>
        ?
          [[WhenDummy]] [17,20:when]<WHEN> --> [17,20:when]<WHEN>

        ?
          end
        end

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -102,7 +102,7 @@ private IEnumerable<CodeElementWrapper> CodeElementFinder(FileCompiler fileCompi
                         {
                             //Ignore all the EndOfFile token 
                             var tempCodeElements = codeElementsLine.CodeElements.Where(c => c.ConsumedTokens.Any(t => t.TokenType != TokenType.EndOfFile));
-
+                            var toList = tempCodeElements.ToList();
```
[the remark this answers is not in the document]
LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1610/commits/4d3efda1f28cb8abe37a2eb0ab1ca14cc74f0e3f

rooksdo (dismissed):

smedilol:
See new PR #1810

PR #1628, opened 2020-01-14
1260 dataflow domain refactorings

fm-117:
code refactoring before integrating semantic domain objects into develop.


TypeCobol/Compiler/Scopes/Domain.cs
```diff
@@ -0,0 +1,245 @@
...
+            /// <summary>
+            /// Total count of symbols in this entry.
+            /// </summary>
+            public int Count => _symbols != null ? _symbols.Count : (_symbol != null ? 1 : 0);
+
+            /// <summary>
+            /// Convenience property to retrieve the first symbol of this entry.
+            /// Returns null if this entry is empty.
+            /// </summary>
+            public TSymbol Symbol => _symbols != null ? (_symbols.Count > 0 ? _symbols[0] : null) : _symbol;
```
smedilol: Can be simplified to:
_symbols?.Count > 0 ? _symbols[0] : _symbol;

smedilol:
Can you write a quick explanation about the class renaming?
I'd like to know the dld name and new name to quickly know what to compare in the code.

fm-117:
Here is a list of what have been refactored :
- interface `TypeCobol.Compiler.Scopes.IDomain` has been deleted
- class `TypeCobol.Compiler.Scopes.Domain` now represents a set of symbols, it is used instead of `Dictionary<string, MultiSymbols>`in the old code
- `Scope<T>.MultiSymbols` and `Scope<T>.Entry` have been merged into `Domain<T>.Entry`
- `TypeCobol.Compiler.Symbols.LevelRenumber` now acts for both `TypeCobol.Compiler.Symbols.LevelRenumber` and `TypeCobol.Compiler.Types.TypeLevelRenumber`
- `TypeCobol.Compiler.Symbols.ProgramExpander` replaces `TypeCobol.Compiler.Symbols.SymbolExpander` and `TypeCobol.Compiler.Types.TypedefExpander` (using private nested classes)
- `TypeCobol.Compiler.Domain.Validator.TypeResolver` replaces 3 classes at once : `SymbolTypeResolver`, `SymbolValidator` and `TypeValidator` (in the same namespace).

TypeCobol/Compiler/Scopes/Domain.cs
fm-117: I have been thinking about it but I think it is less readable. I can change it if you prefer.
Also here is what it would look like for Count property :
```csharp
public int Count => _symbols?.Count ?? (_symbol != null ? 1 : 0);
```

smedilol: Ok the Count

fm-117: Done for both `Count` and `Symbol`. See c00b3b2691a7b3942b0ce36ba6bfa52d3f486501.

fm-117:
Marking this PR as Ready for review. There will be other changes but in a new separate PR to keep this one small and easily reviewable.

smedilol:
I still need time to review this last item:
- class TypeCobol.Compiler.Scopes.Domain now represents a set of symbols, it is used instead of Dictionary<string, MultiSymbols>in the old code

But otherwise it's ok for me.

Issue #1616, opened 2020-01-08
Generation error on program multiple definition


rooksdo:
**Description**
The following TypeCobol generates de following error 

![image](https://user-images.githubusercontent.com/57907033/71963116-3ab95c00-31fb-11ea-8a2d-ecb98707b7e2.png)
when the call to procedure Foo is uncommented:
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZZBCO0.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 tt pic x.
       PROCEDURE DIVISION.

      *    Generation fails with line below uncommented 
      *    call DVZZSTKD::Foo input tt
           
           goback.
       END PROGRAM DVZZBCO0.
      
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZZSTKD.
       PROCEDURE DIVISION.
       declare procedure Foo public
           input t pic x.
       procedure division.
           goback.
       end-declare.
      
       END PROGRAM DVZZSTKD.
````
**Expected behavior**

No generation error should be fired.

PR #1650, opened 2020-01-28
WI #1616  Remove exception on call procedure defined in stacked program

LANOYDO:
fixes https://github.com/TypeCobolTeam/TypeCobol/issues/1616


CLI/src/CLI.cs
```diff
@@ -272,7 +272,8 @@ private void AddProgramsToRootTable(SymbolTable rootTable, IEnumerable<Program>
                     var previousPrograms = rootTable.GetPrograms();
                     foreach (var previousProgram in previousPrograms)
                     {
-                        previousProgram.SymbolTable.GetTableFromScope(SymbolTable.Scope.Namespace).AddProgram(program);
+                        var namespaceSymbolTable = previousProgram.SymbolTable.GetTableFromScope(SymbolTable.Scope.Namespace);
```
fm-117: This certainly fixes the bug but could you investigate why it is needed to add programs into namespace SymbolTable and why sometimes it creates duplicates ?

LANOYDO: - It is needed to add programs main and stacked (nested are inside symbolTable of program) into Namespace-SymbolTable, for each program, even from different file, so parser can understand call of something inside another program.
- Just after CUP analyse, for each main and stacked program, ProgramClassbuilder creates a program with a TableOfNamespaces associated.
TableOfNamespaces (SymbolTable with scope Namespace) contains all the programs already analyzed, at this point. Thus, the SymbolTable of scope Namespace of a stacked program contains a more or less complete list of programs.
- This is why sometimes it creates duplicates in case of stacked program.

modification AddProgramsToRootTable and SymbolTable in order managing programs with same name
https://github.com/TypeCobolTeam/TypeCobol/pull/1650/commits/aee6bdd991f556ad40e12637d1f12234efcbefe6

fm-117: Althouhg I don't get why we get duplicates only for stacked, I think we can improve this part but it requires a bit of refactoring of the `CLI` class. Instead of having multiple instances of parsers and multiple Namespace SymbolTable, we could use the same SymbolTable all the way. We'll discuss about that.

smedilol: Stacked programs are already added to namespace SymbolTable by ProgramClassBuilder, so I think that's why we get duplicate.

To fix this bug, we must first clarify if a stacked program can be referenced from another source file.
From a TypeCobol perpsective:
- We don't need stacked program because we already have procedure.
- Our generated code doesn't works for procedure inside stacked program

As mentionned by Lanoydo, we still need to handle the case of 2 programs with the same name.

LANOYDO: Code has been modify by no longer filling the symbol table with programs stacked from another source file.
https://github.com/TypeCobolTeam/TypeCobol/pull/1650/commits/7562492ef5d810d8b308bdc276630234305d83b8

LANOYDO: Modify test  to verify handling the case of 2 programs with the same name
https://github.com/TypeCobolTeam/TypeCobol/pull/1650/commits/8757ce7ea03a9b960648331204bc55bc744e224e

smedilol: After discussion with @sarterbe, stacked cannot be call by others programs in our environment.
So there is no need to support public procedure in stacked programs.

CLI/src/CLI.cs
```diff
@@ -267,8 +267,9 @@ private void AddProgramsToRootTable(SymbolTable rootTable, IEnumerable<Program>
         {
             if (_configuration.ExecToStep >= ExecutionStep.SemanticCheck)
             {
-                foreach (var program in programs)
+                foreach (var program in programs.Where(p => !p.IsStacked))
```
smedilol: Following my last remarks, only the main program of a source file need to be in namespace SymbolTable.
Can you also filter nested programs here ?

CLI/src/CLI.cs
LANOYDO: done.
https://github.com/TypeCobolTeam/TypeCobol/pull/1650/commits/3680098497555087b762ae698ab7cf256eb2db6a

CLI/src/CLI.cs
LANOYDO: Modifed in order to a stacked program cannot be call by other programs (p.IsMainProgram):
https://github.com/TypeCobolTeam/TypeCobol/pull/1650/commits/3680098497555087b762ae698ab7cf256eb2db6a

CLI/src/CLI.cs
```diff
@@ -267,8 +267,9 @@ private void AddProgramsToRootTable(SymbolTable rootTable, IEnumerable<Program>
         {
             if (_configuration.ExecToStep >= ExecutionStep.SemanticCheck)
             {
-                foreach (var program in programs)
+                foreach (var program in programs.Where(p => p.IsMainProgram))
                 {
+                    // a stacked program should not be referenced from another source file
```
fm-117: Comment is not in-sync with the latest code since we also filter out nested programs.

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1650/commits/d13214d60251eda9d164857a4179ed920b19dd0f

PR #1654, opened 2020-01-29
WI 1545 ValueClauseAreaB

rooksdo:


TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
```diff
@@ -43,6 +46,19 @@ public static void OnCodeElement(DataDescriptionEntry data, CodeElementsParser.D
...
+
+                //Check VALUE clause starts in Area B
+                //Retrieve VALUE token through context
+                var valueToken =
+                    (Token)context?.children.OfType<CodeElementsParser.ValueClauseContext>().SingleOrDefault()?.Start;
+
+                if (valueToken != null)
+                {
+                    //Check starting position of VALUE keyword
+                    if (DocumentFormat.GetTextAreaTypeInCobolReferenceFormat(valueToken) != TextAreaType.AreaB)
```
mayanje: So this check shall only be done if the Document format is ZOSReferenceFormat or a RDZReferenceFormat.
It seems that one can figure out this setting using: ((TypeCobol.Compiler.Parser.CodeElementsLine)valueToken.TokensLine).ColumnsLayout), so the value can be CobolReferenceFormat or FreeTextFormat.

PR #1658, opened 2020-02-04
1440 improve level number detection in scanner

### LINKED ISSUE #1440: Completion on half written words

### LINKED ISSUE #1655: Clean up csprojs after #1627
Continuation of #1627, some configurations (`OutputPath`) have been left out in `CSCup.csproj` and `CSCupRuntime.csproj` and should be removed.

fm-117:
Fixes #1440.
Fixes #1655.


TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -1537,8 +1537,54 @@ private Token ScanNumericLiteral(int startIndex)
...
+                                token.CorrectType(TokenType.LevelNumber);
+                            }
+
+                            //This method is here to help recognize LevelNumbers when PeriodSeparator has been forgotten at the end of previous data definition.
+                            bool GuessIfCurrentTokenIsLevelNumber()
+                            {
+                                var lastSignificantToken = tokensLine.ScanState.LastSignificantToken;
+                                var beforeLastSignificantToken = tokensLine.ScanState.BeforeLastSignificantToken;
+
+                                bool currentTokenIsAtBeginningOfNewLine = token.Line == lastSignificantToken.Line + 1;
```
smedilol: What if there is an empty line between   tokenand lastSignificantToken?

fm-117: Good catch ! Fixed in latest commit.

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -1537,8 +1537,54 @@ private Token ScanNumericLiteral(int startIndex)
...
+                                bool currentTokenIsBeforeAreaB = token.Column < 12;
```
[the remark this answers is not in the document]
fm-117: Issue #1666 has been created to deal with this kind of problem globally.

PR #1683, opened 2020-02-26
WI #1621 Display completion when typed parameter of a procedure is pr…

### LINKED ISSUE #1621: Completion doesn't list all possible variables when a typed paremeter of a procedure is prefixed with a program name
**Describe the bug**
Completion doesn't list all possible variables when a typed paremeter of a procedure is prefixed with a program name.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       working-storage section.
       01 Toto typedef strict private pic X.
       01 Var2 type Toto.
       PROCEDURE DIVISION.
           call Foo input
           .
       declare Foo private input var1 type MyPGM::Toto.
      
       end-declare.
      
      
       END PROGRAM MyPGM.
```
Ask completion after "input" keyword.
Varaible "Var2" is not listed.

If the procedure is declared like this:
```cobol
       declare Foo private input var1 type Toto.
      
       end-declare.
```
Completion after "input" keyword now shows "Var2"

**Expected behavior**
Completion after "input" keyword must propose "Var2"

LANOYDO:
…efixed with a program name
Fixes #1621.


TypeCobol.LanguageServer/Completion Factory/CompletionFactory.cs
```diff
@@ -229,8 +229,24 @@ public static IEnumerable<CompletionItem> GetCompletionForProcedureParameter(Pos
...
+                DataType paramDataType = parameterToFill.DataType;
+                string paramQualifiedNameTail = string.Empty;
+                if (paramDataType.Name.Contains("."))
+                {
+                    // parameter has a qualified name 
+                    int pos = paramDataType.Name.IndexOf(".");
+                    paramQualifiedNameTail = paramDataType.Name.Substring(0, pos);
+                    paramDataType = new DataType(paramDataType.Name.Substring(pos + 1), paramDataType.RestrictionLevel, paramDataType.CobolLanguageLevel);
+                }
+                List<DataDefinition> variablesForCompletion = node.SymbolTable.GetVariablesByType(paramDataType, potentialVariablesForCompletion, SymbolTable.Scope.Program);
```
smedilol: It works but, this code still rely on DataType.
The DataType the exact text written in the TypeCobol source to references a Type.
The property TypeDefinition is a link to a TypeDefinition node if the node is typed.
Property TypeDefinition is more accurate than DataType because you are sure that the link is correct.

So method SymbolTable.GetVariablesByType should be rewritten to use a TypeDefinition instead of a DataType.

LANOYDO: done
https://github.com/TypeCobolTeam/TypeCobol/pull/1683/commits/ea9a7fea4aeb71af6fb64cc6e4b4a87a5e9108c8

PR #1701, opened 2020-03-24
1260 dataflow domain refactoring 3

fm-117:
Included in this PR :
- Renaming
  - AbstractScope to ScopeSymbol
  - Domain to Container
  - Scope to Domain
- Some methods/abstractions reworked
  - SymbolTableBuilder deleted
  - Storage of Built-in types
  - Symbol releasing to forget symbols from the Root

To do in next PR :
- Properties to jump to EnclosingProgram/Function, TopProgram and Root
- Resolve* methods factorization

To discuss :
- Role of the `MayExpand`property


TypeCobol/Compiler/Symbols/NamespaceSymbol.cs
```diff
@@ -122,5 +122,23 @@ public override Container<ScopeSymbol>.Entry ResolveScope(RootSymbolTable root,
...
+
+        internal override void ReleaseSymbols()
+        {
+            var root = TopParent(Kinds.Root) as RootSymbolTable;
+            System.Diagnostics.Debug.Assert(root != null);
+
+            //Release all programs
+            foreach (var program in Programs)
+            {
+                root.Forget(program);
```
mayanje: Why not also use the same term: Release rather Forget?

fm-117: I used two distinct words because it is slightly different :
- on a ScopeSymbol you can ask to 'release' the symbols which means they are kept in the Scope but removed from the Root
- on a root you ask to 'forget' a symbol which means the root does not know anything about it after the call (and the symbol is effectively removed from the Root)

mayanje: So this, is a litle disturbing me, replacing Add by Register and Remove by Forget, because the inverse of Register is Unregister. So in my opinion Add and Remove were suitable, ok for replacing Clear by Release.

fm-117: After discussing with @smedilol, we agreed on Add/Remove on RootSymbolTable and RemoveSymbolsFromRoot on ScopeSymbol.

fm-117: `Register/Unregister` on RootSymbolTable, `UnregisterSymbolsFromRoot` on ScopeSymbol seems the clearest for me.
I'd like to keep different words than Add/Remove because i'd like to emphasize the difference between actually adding/removing a symbol in a set (like on a Namespace when calling RemoveProgram or on a Program when calling Add(variable)) and the action off declaring a symbol in the Root for further use. Also this would mean that AddToUniverse and RemoveFromUniverse would become Register/Unregister but for variables. This unifies vocabulary for the same action on different types of symbols.

mayanje: I was also thinking to the couple Put/Retrieve, or Store/Retrieve because
I am not sure that this is a registration mechanism in the RootSymbolTable with the couple Register/Unregister.

fm-117: See how it looks with Register/Unregister (7380daa05426b99dae866cb7c5e04f219fab061a).
'Retrieve' doesn't work because it still has the idea to add something into the table instead of making the table be aware of something.

mayanje: Exactly RootSymbolTable is a table,into which we are storing symbols, so the inverse is Extract, so the couple Store/Extract is in my opinion suitable because Register/Unregister means that The RootSymbolTable is a set of registers, it is a set of Scopes and other things.

fm-117: 'Store' sounds good to me. Instead of 'Extract', I suggest 'Discard' because in my opinion the word 'extract' can be interpreted as 'to extract a value from' (as a synonym for 'to read'). 'discard' is clearer on the fact that the symbol won't be accessible anymore from the table. So would you agree on Store/Discard/DiscardSymbols ?

mayanje: Ok, Discard means "Jeter" un french this not exactly what I mean, that is we store the symbol in the universe of the RootSymbolTable, and then I think we can erase it. so with your last comments I suggest the couple Store/Erase

fm-117: The symbol is not erased, it may still be defined in a Scope and still has an owner. It is simply not known anymore in the root table, that's why I think 'discard" is more precise in this context. 'discard' only means that the symbol is thrown away from the root table but it is not destroyed or otherwise modified. 'erase' sounds like the symbol may vanish which is not the case.

mayanje: Ok, so we Store it and We Take It Off (Store/TakeOff)

fm-117: 'take off' means 'décoller' (for a plane 😄). 'to discard' is the proper antonym for 'to store'.

mayanje: It's true Take Off for a plane means "décoller", but for a thing it can means "enlever, ôter, retirer, supprimer" (Harrap's Compact Dictionary). But we can stop this discussion now, let's go for Store/Discard.

fm-117: ok

fm-117: Done in ac2679e48d07cf35b69f9898bdaf25d2cf0fe2ef. Awaiting approval.

PR #1717, opened 2020-04-14
WI 1625 Address change on input parameter

### LINKED ISSUE #1625: Allow for address change of pointer on input parameter in procedure
**Description**

The following TypeCobol instruction should be accepted inside a procedure (where myInput is declared as input) :
```cob
Set address of myInput to address of var1
```
because no change on input variable (myInput) memory location is made.

rooksdo:
Rebase of #1716 

Fixes #1625

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
@@ -600,9 +600,16 @@ namespace TypeCobol.Compiler.Diagnostics
                 if (!isReadStorageArea && node.SymbolTable.CurrentScope == SymbolTable.Scope.Function)
                 {
                     var paramDesc = (dataDefinitionPath?.CurrentDataDefinition ?? dataDefinitionFound) as ParameterDescription;
+                    //Check if we're dealing with an input parameter
                     if (paramDesc?.PassingType == ParameterDescription.PassingTypes.Input)
                     {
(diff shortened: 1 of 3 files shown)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -600,9 +600,16 @@ public static DataDefinition CheckVariable(Node node, StorageArea storageArea, b
...
                 {
                     var paramDesc = (dataDefinitionPath?.CurrentDataDefinition ?? dataDefinitionFound) as ParameterDescription;
+                    //Check if we're dealing with an input parameter
                     if (paramDesc?.PassingType == ParameterDescription.PassingTypes.Input)
                     {
-                        DiagnosticUtils.AddError(node, "Input variable '" + paramDesc.Name + "' is modified by an instruction", area.SymbolReference);
+                        var specialRegister = storageArea as StorageAreaPropertySpecialRegister;
+                        //Unless this is a format 5 set statement, we have an error. So we're checking we're not in the following format  :
+                        //set (address of)? identifier(pointer) TO (address of)? identifier | NULL
+                        if (specialRegister?.SpecialRegisterName.TokenType != TokenType.ADDRESS || !specialRegister.IsWrittenTo)
```
fm-117, line 609: Is use of property IsWrittenTo required ? You already have isReadStorageArea in the method, is it the same ? Note that you already are in an if condition stating that the storage area is written to (see line 600 just above).

rooksdo: Correct. Done.

PR #1729, opened 2020-05-27
WI 969 Forbid Sync Clause for Strict Types

### LINKED ISSUE #960: Forbird SYNC clause with Strict Type

rooksdo:
Closes #960

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
@@ -389,6 +389,49 @@ namespace TypeCobol.Compiler.Diagnostics
 
             }
 
+            //Check if variable is Sync inside Strict Type definition
+            if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.InsideTypedef) &&
+                dataDefinition.SemanticData.HasFlag(Symbol.Flags.Sync))
+            {
+                //Typedef instruction => check if it's marked Strict
+                if (dataDefinition.SemanticData.Kind == Symbol.Kinds.Typedef && dataDefinition.SemanticData.HasFlag(Symbol.Flags.Strict))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"Cannot declare Type definition {dataDefinition.Name} with Sync clause because it is Strict.");
+                }
+                //Variable inside Typedef => check if parent typedef is marked Strict
+                else if (dataDefinition.SemanticData.NearestParent(Symbol.Kinds.Typedef).HasFlag(Symbol.Flags.Strict))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} cannot be declared with Sync clause because it is part of Strict Type definition {dataDefinition.ParentTypeDefinition?.Name}.");
+                }
+
+            }
+            
+            //Variable is of a user defined Strict Type
+            if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.HasATypedefType) && dataDefinition.DataType.RestrictionLevel == RestrictionLevel.STRICT)
+            {
+                //Sync clause not allowed on variable declaration because its type is Strict
+                if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.Sync))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} cannot be declared with Sync clause because its Type definition {dataDefinition.DataType.Name} is Strict.");
+                }
+                //Check whether variable is under a sync parent
+                else if (dataDefinition.CodeElement?.LevelNumber?.Value > 1)
+                {
+                    //Retrieve higher parent
+                    Symbol parent = dataDefinition.SemanticData.Owner;
+                    //Iterate through parent variables
+                    while (parent != null && parent.Kind == Symbol.Kinds.Variable)
+                    {
+                        //Check if parent is synchronized
+                        if (parent.HasFlag(Symbol.Flags.Sync))
+                        {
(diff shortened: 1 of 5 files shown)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -389,6 +389,49 @@ public override bool Visit(DataDefinition dataDefinition)
...
+                //Variable inside Typedef => check if parent typedef is marked Strict
+                else if (dataDefinition.SemanticData.NearestParent(Symbol.Kinds.Typedef).HasFlag(Symbol.Flags.Strict))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} cannot be declared with Sync clause because it is part of Strict Type definition {dataDefinition.ParentTypeDefinition?.Name}.");
+                }
+
+            }
+            
+            //Variable is of a user defined Strict Type
+            if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.HasATypedefType) && dataDefinition.DataType.RestrictionLevel == RestrictionLevel.STRICT)
```
smedilol, line 410: DataType is the old type mechanism that must be replaced by new Type class.
New Type class doesn't have the resolved TypeDefinition yet.
You can use dataDefinition.TypeDefinition?.RestrictionLevel in the meantime.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -389,6 +389,49 @@ public override bool Visit(DataDefinition dataDefinition)
...
+            {
+                //Sync clause not allowed on variable declaration because its type is Strict
+                if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.Sync))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} cannot be declared with Sync clause because its Type definition {dataDefinition.DataType.Name} is Strict.");
+                }
+                //Check whether variable is under a sync parent
+                else if (dataDefinition.CodeElement?.LevelNumber?.Value > 1)
```
fm-117, line 418: dataDefinition.SemanticData can be converted to VariableSymbol which has a Level property. Start by casting dataDefinition.SemanticData so you can use it all the way in the checker.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -389,6 +389,49 @@ public override bool Visit(DataDefinition dataDefinition)
...
+                {
+                    //Retrieve higher parent
+                    Symbol parent = dataDefinition.SemanticData.Owner;
```
smedilol, line 421: I thought that sync flag were inherited.
Anyway, can you replace this code by an inheritance of Sync flag?
This way the `if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.Sync))` will be enough to test this rule.

fm-117: Could you update your branch with latest commits from 1721_SemanticDomainLite please ? 558b8e75295dd0dab6d8c8fcd330bdc4e8b194f3 should give you inherited SYNC flags.

rooksdo: Ok now sync flag is inherited, but the information that it is inherited is not present. So those two cases are indistinguishable using only Symbol class:
```cob
01 MyGroup sync.
    05 MyTypeVar type MyTypeStrict.
```
and
```cob
01 MyGroup.
    05 MyTypeVar type MyTypeStrict sync.
```

smedilol:
Beware, there is a mix of issue ID in commit : 960 and 969.
The char `#` is also missing before issue ID in commit.

rooksdo:
See rebase PR #1733

PR #1733, opened 2020-06-02
WI 960 ForbidSyncClauseWithStrictType

### LINKED ISSUE #960: Forbird SYNC clause with Strict Type

rooksdo:
Rebase of PR #1729.

Closes #960

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
@@ -389,6 +389,29 @@ namespace TypeCobol.Compiler.Diagnostics
 
             }
 
+            //Check if Strict Typedef declaration uses Sync clause
+            if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.InsideTypedef) &&
+                dataDefinition.SemanticData.HasFlag(Symbol.Flags.Sync))
+            {
+                //Typedef instruction => check if it's marked Strict
+                if (dataDefinition.SemanticData.Kind == Symbol.Kinds.Typedef && dataDefinition.SemanticData.HasFlag(Symbol.Flags.Strict))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"Cannot declare Type definition {dataDefinition.Name} with Sync clause because it is Strict.");
+                }
+                //Variable inside Typedef => check if parent typedef is marked Strict
+                else if (dataDefinition.SemanticData.NearestParent(Symbol.Kinds.Typedef).HasFlag(Symbol.Flags.Strict))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} is part of a declaration using Sync clause in Strict Type definition {dataDefinition.ParentTypeDefinition?.Name}.");
+                }
+
+            }
+            //Check if variable of user defined Strict Type is declared or has a parent declared with Sync clause (flag is inherited so no need to iterate through parents)
+            else if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.HasATypedefType) && 
+                dataDefinition.TypeDefinition?.RestrictionLevel == RestrictionLevel.STRICT &&
+                dataDefinition.SemanticData.HasFlag(Symbol.Flags.Sync))
+            {
+                DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} cannot be declared or have a parent declared with Sync clause because its Type definition {dataDefinition.DataType.Name} is Strict.");
+            }
 
             if (HasChildrenThatDeclareData(dataDefinition))
             {
--- a/TypeCobol/Compiler/Domain/ProgramSymbolTableBuilder.cs
+++ b/TypeCobol/Compiler/Domain/ProgramSymbolTableBuilder.cs
@@ -140,10 +140,10 @@ namespace TypeCobol.Compiler.Domain
 
             //Create the program
             var program = new ProgramSymbol(programIdentification.ProgramName.Name)
-                          {
-                              //Set empty ScopeType for now, the parameters and return variable will be resolved later
-                              Type = new ScopeType(new List<VariableSymbol>(), null)
-                          };
+            {
+                //Set empty ScopeType for now, the parameters and return variable will be resolved later
+                Type = new ScopeType(new List<VariableSymbol>(), null)
+            };
             if (this.CurrentProgram == null)
             {
                 //This is the main program or a stacked program.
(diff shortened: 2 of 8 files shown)


smedilol:
Commit message is not correct, it should be something like:
`WI #960 Forbid sync clause with strict type`

TypeCobol/Compiler/Domain/ProgramSymbolTableBuilder.cs
```diff
@@ -246,6 +246,74 @@ public override void EndLinkageSection()
             CurrentDataDivisionSection = null;
         }
 
+        /// <summary>
+        /// Resolve in the current program linkage section, the given parameter.
+        /// </summary>
+        /// <param name="p">The parameter to resolve.</param>
+        /// <returns>The resolved variable if any, null otherwise</returns>
+        private VariableSymbol ResolveUsingParameter(CallTargetParameter p)
```
smedilol, line 254: This code has been removed from develop branch. It must be removed as weel here.

TypeCobol/Compiler/Domain/ProgramSymbolTableBuilder.cs
```diff
@@ -246,6 +246,74 @@ public override void EndLinkageSection()
...
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        /// Starting a Program PROCEDURE DIVISION => Collect all parameters.
+        /// </summary>
+        /// <param name="header"></param>
+        public override void StartProcedureDivision(ProcedureDivisionHeader header)
```
smedilol, line 289: Same remarks here, this method must be removed.

TypeCobol/Compiler/Symbols/Builtins.cs
```diff
@@ -58,18 +59,22 @@ static Builtins()
 
             Boolean = new TypedefSymbol(string.Intern("Bool"));
             Boolean.Type = new TypedefType(Boolean, new Type(Type.Tags.Boolean));
+            Boolean.Type.SetFlag(Symbol.Flags.Strong, true, false);
```
smedilol, line 62: For me these types (Bool, Date, Currency and String) are pure TypeCobol types and therefore are Strict.

fm-117: They are marked as Strong in their TypeDefinition counterparts, that's why we applied the Strong flag here.

TypeCobol/Compiler/Symbols/Builtins.cs
```diff
@@ -1,4 +1,5 @@
-﻿using TypeCobol.Compiler.Types;
+﻿using TypeCobol.Compiler.CodeElements;
```
fm-117, line 1: This first using is useless, it may have been added involuntarily. Please remove it.

TypeCobol/Compiler/Domain/ProgramSymbolTableBuilder.cs
rooksdo: Done

TypeCobol/Compiler/Domain/ProgramSymbolTableBuilder.cs
rooksdo: Done

TypeCobol/Compiler/Symbols/Builtins.cs
rooksdo: I agree with @fm-117

TypeCobol/Compiler/Symbols/Builtins.cs
rooksdo: Done.

TypeCobol/Compiler/Symbols/Builtins.cs
smedilol: Ok

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -389,6 +389,29 @@ public override bool Visit(DataDefinition dataDefinition)
...
+                //Variable inside Typedef => check if parent typedef is marked Strict
+                else if (dataDefinition.SemanticData.NearestParent(Symbol.Kinds.Typedef).HasFlag(Symbol.Flags.Strict))
+                {
+                    DiagnosticUtils.AddError(dataDefinition, $"{dataDefinition.Name} is part of a declaration using Sync clause in Strict Type definition {dataDefinition.ParentTypeDefinition?.Name}.");
+                }
+
+            }
+            //Check if variable of user defined Strict Type is declared or has a parent declared with Sync clause (flag is inherited so no need to iterate through parents)
+            else if (dataDefinition.SemanticData.HasFlag(Symbol.Flags.HasATypedefType) && 
+                dataDefinition.TypeDefinition?.RestrictionLevel == RestrictionLevel.STRICT &&
```
fm-117, line 410: Actually we could use SemanticDomain here too : `dataDefinition.TypeDefinition.SemanticData.HasFlag(Strict)`

PR #1493, opened 2019-10-09
WI #1449 Add diagnostic warning when a STOP RUN or GOBACK is not the …

LANOYDO:
…last statement in paragraph or section

Issue #1449, opened 2019-09-05
New diagnostic : "STOP RUN" or "GOBACK" should be the last statement o


smedilol:
**Is your feature request related to a problem? Please describe.**
New diagnostic : "STOP RUN" or "GOBACK" should be the last statement o

If a `stop run` or `goback` statement is encountered, then it must be the last statement of the paragraph/section.
Generate an error if it's not the case.

Cobol example:
```cobol
        INIT.
       *Ok
            goback
            .
        INIT2.
       *Ko
            goback
            display "Hello"
            .
```

smedilol:
See comment on #1493

fm-117:
This overlaps with unreachable code detection (see #2016 and #2023). When an instruction follows a `STOP RUN` or a `GOBACK`, a warning is created to signal that it won't be reached.
Moreover, we also have a specific check in our private code-quality analyzer.

PR #1735, opened 2020-06-05
1260 cfg semantic domain lite

fm-117:
This is a draft PR to discuss all code enhancements/refactorings.


fm-117:
General code improvements :
- use more explicit names for generic type arguments
- generics
  - instruction type parameterization may be removed as we always use `Node` to describe instructions
  - the data type is used only for blocks, instead of having the data type in the builder we could specify the type of blocks to be used or even better provide a factory for blocks. The builder would only use a common abstraction of blocks instead of having to deal with parameterized blocks
- accessibility of members
  - use stricter visibility when possible to avoid unproper usages
- ControlFlowGraphBuilder does too much
  - Some methods could be moved into other classes to be closer to the data they manipulate
  - for example: grafting and relocate process could be attached to the BasicBlockForNodeGroup

fm-117:
ControlFlowGraphBuilder possible improvements
- split into 2 classes
  - one should be responsible for the jumps between builders when entering or leaving program/function
    - would inherit from `ProgramClassBuilderNodeListener`
    - would override Enter/Exit and forward calls to the other class (calling EnterXXX/LeaveXXX methods with typed node)
  - the other would be responsible for the actual building of a graph for a single program/function
    - contains the EnterXXX/LeaveXXX methods
    - can be the CFG graph itself or a new class
- big switches on CodeElementType are dangerous
  - because some cases are missing (new Cobol V6 statements must be added)
  - try to avoid those big switches
    - factorize into one switch intead of 3
    - use only the cases that are actually handled by the builder and leave the default case with a default behavior that should not break the graph
- make better use of the context
  - introduce an abstraction for `DeclarativesContext` and `MultiBranchContext` ?
  - turn the context properties into a Strategy pattern ?

mayanje:
- generics 

    - The Generic implementation should not be removed even if there only one instantiation using Node.

- The logic of CFG Building belongs to the ControlFlowGraphBuilder not to BasiCBlock instance or Basic Block Group instance. This shall not be changed.

**Be carefull Refactoring does not means redesign.**

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.BasicBlockForNode.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// BasicBlock which instruction are Nodes.
+        /// </summary>
+        /// <typeparam name="D"></typeparam>
+        public class BasicBlockForNode : BasicBlock<Node, D>
+        {
+            /// <summary>
+            /// Any tag associated to this Node.
+            /// </summary>
+            public object Tag
```
fm-117, line 18: This is used to memorize the section or paragraph name in which this block appears, so this can be turned into `string ProcedureName`.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgParagraphSymbol.cs
```diff
@@ -0,0 +1,41 @@
...
+using TypeCobol.Compiler.Symbols;
+
+namespace TypeCobol.Analysis.Cfg
+{
+    public partial class ControlFlowGraphBuilder<D>
+    {
+        /// <summary>
+        /// A Paragraph symbol used by a CFG, it contains Sentence symbols.
+        /// </summary>
+        internal class CfgParagraphSymbol : ParagraphSymbol
```
fm-117, line 11: Inheriting from `ParagraphSymbol leads to duplictae symbol instanciation as the SemanticDomain currently builds the paragraph symbols too.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSectionSymbol.cs
```diff
@@ -0,0 +1,42 @@
...
+using TypeCobol.Compiler.Symbols;
+
+namespace TypeCobol.Analysis.Cfg
+{
+    public partial class ControlFlowGraphBuilder<D>
+    {
+        /// <summary>
+        /// A Section symbol used by Cfg : it contains Sentences and Paragraphs in order of appearance.
+        /// </summary>
+        internal class CfgSectionSymbol : SectionSymbol
```
fm-117, line 11: Same remark as for `CfgParagraphSymbol`.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSectionSymbol.cs
```diff
@@ -0,0 +1,42 @@
...
+            /// <param name="name">Section's name</param>
+            public CfgSectionSymbol(string name) : base(name)
+            {
+                SentencesParagraphs = new Domain<Symbol>(this);
+            }
+
+            /// <summary>
+            /// All sentences and paragraphs in this section in the order of appearance.
+            /// </summary>
+            public Domain<Symbol> SentencesParagraphs
```
fm-117, line 25: Add a method to yield sentences from a section. Users of the classs would not have to test symbol kind.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSectionSymbol.cs
mayanje: Same remark as for CfgParagraphSymbol.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
```diff
@@ -0,0 +1,77 @@
...
+using TypeCobol.Compiler.Symbols;
+
+namespace TypeCobol.Analysis.Cfg
+{
+    public partial class ControlFlowGraphBuilder<D>
+    {
+        /// <summary>
+        /// A Sentence used for our builder. A Sentence is a special symbol.
+        /// </summary>
+        internal class CfgSentence : Symbol
```
fm-117, line 11: A sentence looks like more a block group than a symbol.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgParagraphSymbol.cs
mayanje: The inheritance is correct it is a specialization, So you should reproduce were what we suggested : Section and Paragraph symbols can be created by CFG if CFG is enabled or by SDLite otherwise.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            /// <summary>
+            /// First Block associated to this sentence.
+            /// </summary>
+            internal BasicBlockForNode Block
```
fm-117, line 16: => Allblocks.First() ?

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                get;
+                set;
+            }
+
+            /// <summary>
+            /// The Block Index Associated to the Block.
+            /// </summary>
+            internal int BlockIndex
```
fm-117, line 25: => Block.Index ?

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    AllBlocks = new LinkedList<BasicBlockForNode>();
+                }
+                AllBlocks.AddLast(block);
+            }
+
+            /// <summary>
+            /// Sentence counter
+            /// </summary>
+            private static int SentenceCounter = 0;
```
fm-117, line 56: This won't work in multiple parsing scenarios unless the counter is reset at some point.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        /// <summary>
+        /// Declarative context class.
+        /// </summary>
+        internal class DeclarativesContext
+        {
+            /// <summary>
+            /// Current Block before all declaratives sections and paragraphs.
+            /// </summary>
+            internal BasicBlockForNode CurrentBlock;
```
fm-117, line 15: Rename as `StartBlock` ?

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
mayanje: This is a sentence from a COBOL perspective, and CfgSectionSymbol and CfgParagraphSymbol contains sentences seen as symbols. It is a convenience to see a Sentence as Symbol so that it can be given an Owner which is its Section or Paragraph symbol.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.MultiBranchContext.cs
```diff
@@ -0,0 +1,183 @@
...
+    {
+        /// <summary>
+        /// The context of a Multi Branch instruction like IF or Evaluate.
+        /// </summary>
+        internal class MultiBranchContext
+        {
+            /// <summary>
+            /// Current Block before multi branches
+            /// </summary>
+            internal BasicBlockForNode CurrentBlock;
```
fm-117, line 17: Same remark, this is the StartBlock

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.MultiBranchContext.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            /// <summary>
+            /// The instruction associated to this context.
+            /// </summary>
+            internal Node Instruction;
+
+            /// <summary>
+            /// Any contextual Data.
+            /// </summary>
+            internal object ContextualData;
```
fm-117, line 47: Always used to hold a `List<Node>` (these are the when alternatives).

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.MultiBranchContext.cs
mayanje: In CFG a Start Block is in fact a Root Block, this is not always the case here. But ok, it is like the SourceBlock. or OriginBlock.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
mayanje: In CFG a Start Block is in fact a Root Block, this is not always the case here. But ok, it is like the SourceBlock. or OriginBlock.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        public ControlFlowGraph<Node, D> Cfg
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// The Current Program symbol being built as a Scope
+        /// </summary>
+        public Program Program
```
fm-117, line 82: Duplicate with CurrentProgram proprerty, this can be safely removed.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        internal CfgSentence CurrentSentence
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// The Section and Paragraph Domain of this program.
+        /// </summary>
+        internal Container<Symbol> SectionsParagraphs
```
fm-117, line 145: If the CFG builds its own Section and Paragraph symbol, maybe we could introduce an abstraction for both (`ProcedureSymbol`), this may simplify the code.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        /// <summary>
+        /// All encountered sentences
+        /// </summary>
+        internal List<CfgSentence> AllSentences;
+
+        public IList<Diagnostic> Diagnostics { get; private set; }
+        /// <summary>
+        /// Empty constructor.
+        /// </summary>
+        public ControlFlowGraphBuilder() : this(null)
```
fm-117, line 192: Can be safely removed.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="parentCfgBuilder">Parent Control Flow Builder for a nested program</param>
+        public ControlFlowGraphBuilder(ControlFlowGraphBuilder<D> parentCfgBuilder = null)
+        {
+            this.ParentProgramCfgBuilder = parentCfgBuilder;
+            this.Diagnostics = new List<Diagnostic>();
+            this.UseEvaluateCascade = true;
+            this.UseSearchCascade = true;
```
fm-117, line 205: `UseEvaluateCascade` and `UseSearchCascade` are always set to true. Should we keep these properties ? If so, they should be added as parameters in the constructor.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+            Mode = CfgMode.Normal;
+        }
+
+        /// <summary>
+        /// Called when a Node has been completely parsed.
+        /// </summary>
+        /// <param name="node"></param>
+        /// <param name="program"></param>
+        public override void OnNode(Node node, Program program)
```
fm-117, line 214: Remark for the entire class: remove empty overrides.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Determines if the given Node is a statement.
+        /// </summary>
+        /// <param name="node">The node to be checked</param>
+        /// <returns>Return true if the node is a statement, false otherwise.</returns>
+        public static bool IsStatement(Node node)
```
fm-117, line 223: Cobol V6.2 new statements missing.

Naming may be a bit confusing because to determine if a Node is a Statement, we already have the `Statement` marker interface. This is more like "Is this node a statement that CFG should handle".

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+            {
+                //A Sentence has at least one block
+                System.Diagnostics.Debug.Assert(sentence.AllBlocks != null);
+                System.Diagnostics.Debug.Assert(sentence.AllBlocks.First.Value == sentence.Block);
+                foreach (var block in sentence.AllBlocks)
+                {//We must clone each block of the sequence and add them to the group.                    
+                    //System.Diagnostics.Debug.Assert(!clonedBlockIndexMap.ContainsKey(block.Index));
+                    if (!clonedBlockIndexMap.ContainsKey(block.Index))
+                    {//If this block has been already add, this mean there are recursive GOTOs
+                        clonedBlockIndexMap[block.Index] = block.Index;
```
fm-117, line 1127: This is the only add operation on the dictionary so the keys are always equal to the values. The Dictionary can be turned into a List.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+
+        /// <summary>
+        /// Relocate the graph generated by a BasicBlockForNodeGroup
+        /// </summary>
+        /// <param name="p">The Perform Procedure node source of the call</param>
+        /// <param name="group">The Group to be relocated</param>
+        /// <param name="clonedBlockIndexMap">The map between the old block indices to their new block indices.</param>
+        /// <returns>true if the relocation is successful, false if the relocation goes beyond the group limit, 
+        /// this often means that de target paragraphs of the PERFORM goes out of the paragraph.</returns>
+        private bool RelocateBasicBlockForNodeGroupGraph(PerformProcedure p, BasicBlockForNodeGroup group, Dictionary<int, int> clonedBlockIndexMap)
```
fm-117, line 1210: I need to review this again. I don't get the meaning of the Diagnostic. Also we are talking about clonedBlock but it is an existing instance not a new one.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        /// Compute the terminal blocks associated to Group
+        /// 
+        /// </summary>
+        /// <param name="group">The group to compute the terminal blocks</param>
+        private void ComputeBasicBlockGroupTerminalBlocks(BasicBlockForNodeGroup group)
+        {
+            if (group.Group.Count > 0)
+            {
+                LinkedListNode<BasicBlock<Node, D>> first = group.Group.First;
+                MultiBranchContext ctx = new MultiBranchContext(this.CurrentProgramCfgBuilder, null);
```
fm-117, line 1290: We should not instantiate a context here. In my opinionThe `GetTerminalSuccessorEdges` method belongs to the `BasicBlockForNode` class. We'll have to add a reference to the graph in the block class though.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+            return block;
+        }
+
+        /// <summary>
+        /// Create a Fresh Control Flow Graph Builder.
+        /// </summary>
+        /// <returns>The fresh Control Flow Graph Builder</returns>
+        protected virtual ControlFlowGraphBuilder<D> CreateFreshControlFlowGraphBuilder(ControlFlowGraphBuilder<D> parentCfgBuilder = null)
+        {
+            return new ControlFlowGraphBuilder<D>();
```
fm-117, line 2639: Bug, the parent is not used. this has been fixed in DFA though.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        /// Initialize the Cfg by a Function.
+        /// </summary>
+        /// <param name="funDecl">The target function declaration of the Cfg</param>
+        protected virtual void InitializeCfg(FunctionDeclaration funDecl)
+        {
+            Cfg = new ControlFlowGraph<Node, D>();
+            Cfg.ProgramNode = funDecl;
+        }
+
+        public static readonly string ROOT_SECTION_NAME = "<< RootSection >>";
```
fm-117, line 2662: Is this required ? I mean, in Semantic Domain and according to the Cobol spec, when a paragraph is not inside a section, it is attached to the Program directly. This is more problematic for sentences as they still need to be attached to something...

TypeCobol.Analysis/Cfg/DefaultControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace TypeCobol.Analysis.Cfg
+{
+    /// <summary>
+    /// Default Control Flow Graph Builder with any object has Data.
+    /// </summary>
+    public class DefaultControlFlowGraphBuilder : ControlFlowGraphBuilder<object>
```
fm-117, line 6: Should be typed otherwise the Data type can not be used. Fixed in DFA.

TypeCobol.Analysis/Graph/CfgDotFileForNodeGenerator.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Get the string representing an instruction.
+        /// </summary>
+        /// <param name="instruction">The instruction to get the string representation.</param>
+        protected override string InstructionToString(Node instruction)
```
fm-117, line 38: Duplicate with the method in BasicBlock itself. See how we can factorize this.

TypeCobol.Analysis/Graph/CfgDotFileForNodeGenerator.cs
```diff
@@ -0,0 +1,128 @@
...
+                    sw.WriteLine("subgraph cluster_" + group.GroupIndex + '{');
+                    sw.WriteLine("color = blue;");
+                    if (group.Group.Count > 0)
+                    {
+                        sw.WriteLine(string.Format("label = \"{0}\";", ((ControlFlowGraphBuilder<D>.BasicBlockForNode)group.Group.First.Value).Tag));
+                        CfgDotFileForNodeGenerator<D> cfgDot = new CfgDotFileForNodeGenerator<D>(cfg);
+                        cfgDot.EmittedGroupIndices = EmittedGroupIndices;
+                        cfgDot.FullInstruction = this.FullInstruction;
+                        cfgDot.Writer = sw;
+                        cfgDot.DigraphBuilder = new StringBuilder();
```
fm-117, line 105: Why creating a StringBuilder if we already have a TextWriter ? It would be more efficient to continue writing subgraph on the writer.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
```diff
@@ -0,0 +1,170 @@
...
+        public bool HasFlag(Flags flag)
+        {
+            return (this.Flag & flag) != 0;
+        }
+
+        /// <summary>
+        /// Root blocks. Usually it has a single item which is the first block in the program.
+        /// But Exception handlers are also stored as root blocks.
+        /// </summary>
+        public List<BasicBlock<N, D>> RootBlocks
```
fm-117, line 65: So far the list contains the first block of a Procedure Division, how are the exception blocks handled ?

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        /// <summary>
+        /// DFS Depth First Search implementation.
+        /// </summary>
+        /// <param name="callback">CallBack function</param>
+        public void DFS(BasicBlockCallback callback)
+        {
+            foreach(var root in RootBlocks)
+            { 
+                DFS(root, callback);
```
fm-117, line 166: Call the most precise overload to avoid multiple BitArray instanciation.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.BasicBlockForNode.cs
mayanje: This is a general tag having object type, I don't want to give it as meaning ProcedureName out of any context.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSectionSymbol.cs
mayanje: Ok but the same thing must be done for CfgParagraphSymbol, as you suggest later it will be necessay to have a common base class for CfgSectionSymbol and CfgParagraphSymbol and use code in method 
```C#
private IEnumerable<CfgSentence> YieldSectionOrParagraphSentences(Symbol sectionOrParagraphSymbol)
```
The problem is that CfgSectionSymbol must ihnerits SectionSymbol from SDLite and CfgParagraphSymbol must inherits ParagraphSymbol from SDLite, thus we should create an Interface like ISentenceContainer that implements such contract.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
mayanje: No it is the block of the sentence, in fact in COBOL the basic block unit is the Sentence, (in other words it is the dot character separator).

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
mayanje: Ok, this the index in the Cfg.SuccessorEdges list of the Sentence's block. In other words the previous block before the sentence's block will have in its successor list a value which is the BlockIndex. So I agree the name of this field should be something like 'BlockIndexInSuccessorList'.
This value is also used for resolving target successor index for GOTOs

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
mayanje: No in fact no this just an id for the sentence used to give a distinct name to a sentence. it's a static field we don't care about concurrency here. an **int** has 214748364750 values, it can be changed to an **uint** to have 4294967295 values, don't worry.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.MultiBranchContext.cs
mayanje: Yes, at the time I didn't know if there can be another type of contextual data.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: yes

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: See your post about adding a method in CfgSectionSymbol so that it can yield its sentences. I answered that CfgSectionSymbol must ihnerits from SectionSymbol and CfgParagraphSymbol inherits from ParagraphSymbol, so if you want to introduce this base class this shall be done in SDLite.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yes, because of default argument value.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: The first implementation was using a CFG graph construction with basic block in sequence in sequence for WHENs and WHENOTHER alternatives, but the best way followed by IBM compiler when the code is generated is to build corresponding basic blocks in waterfall (Waterfall IF THEN ELSE construct). So these options are left there in order to be able to test both styles.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yes, cleaning code

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yes, this is the purpose of the static method: _"Is this node that CFG can handle"_ Because for instance it exists Node without CodeElement, CFG cannot take in account such nodes (like Then, WhenGroup, ...).

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
```diff
@@ -0,0 +1,97 @@
...
+                System.Diagnostics.Debug.Assert(nextBlock != null);
+
+                //First Current block is linked to the next block.
+                int nbIndex = Builder.Cfg.SuccessorEdges.Count;
+                Builder.Cfg.SuccessorEdges.Add(nextBlock);
+                CurrentBlock.SuccessorEdges.Add(nbIndex);
+
+                //For each section, link the current block to the first block of the section.
+                bool bFirstsection = true;
+                foreach (var section in Sections)
```
fm-117, line 67: Iterate over `Sections.Skip(1)` instead.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    var sentences = Builder.YieldSectionOrParagraphSentences(section);
+                    foreach (var sentence in sentences)
```
fm-117, line 70: Use `First()` LINQ method instead.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: No, looking in the list as a complexity of O(n) where n is the count of element in the list, but looking in a Dictionary as a complexity of O(1) method _ContainsKeys But List Contains method as a complexcity of O(n)

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
```diff
@@ -0,0 +1,1181 @@
...
+        public void IfThenTest()
+        {
+            string path = Path.Combine(Directory.GetCurrentDirectory(), "BasicCfgInstrs", "IfThen0.cbl");
+            var document = TypeCobol.Parser.Parse(path, /*format*/ DocumentFormat.RDZReferenceFormat, /*autoRemarks*/
+                false, /*copies*/ null);
+            //Assert.IsTrue(document.Results.PrgSymbolTblBuilder.Programs.Count == 1);
+            string expectedPath = Path.Combine(Directory.GetCurrentDirectory(), "DotOutput", "IfThen0.dot");
+
+            Assert.IsTrue(CfgBuilder.AllCfgBuilder.Count == 1);
+            Assert.IsNotNull(CfgBuilder.AllCfgBuilder);
```
smedilol, line 171:
```suggestion
            Assert.IsTrue(CfgBuilder.AllCfgBuilder?.Count == 1);
```

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
```diff
@@ -0,0 +1,1181 @@
...
+
+            CfgTestUtils.GenDotCfgAndCompare(CfgBuilder.Cfg, path, expectedPath);
```
smedilol, line 173: We agreed that CFG can run even if there are error.
But our tests check how many errors there are.
I guess most of CFG tests have no errors.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yes, at the time I though that blocks shall be cloned, but here the word "clone" shall be changed to "referenced". About diagnostics we shall use an option to emit or not diagnostics inside CFG, here it is a real CFG diagnostic

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: Ok then we should use `HashSet<int>` which has a `Contains` method with a complexity of O(1).

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        {
+            System.Diagnostics.Debug.Assert(this.CurrentProgramCfgBuilder.MultiBranchContextStack != null);
+            System.Diagnostics.Debug.Assert(this.CurrentProgramCfgBuilder.MultiBranchContextStack.Count > 0);
+            MultiBranchContext ctx = this.CurrentProgramCfgBuilder.MultiBranchContextStack.Pop();
+            System.Diagnostics.Debug.Assert(ctx.Branches != null);
+
+            if (UseEvaluateCascade)
+            {   //Pop each MultiBranchContextStack instance till to the EVALUATE one
+                //and close each one.
+                while (ctx.Instruction == null)
```
fm-117, line 1715: Shouldn't it be `while (ctx.Instruction != evaluate)` ?

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        /// All pending Next Sentence instructions that will be handled at the end of the Procedure Division.
+        /// </summary>
+        internal LinkedList<Tuple<NextSentence, BasicBlockForNode, CfgSentence>> PendingNextSentences;
+
+        /// <summary>
+        /// All encountered sentences
+        /// </summary>
+        internal List<CfgSentence> AllSentences;
+
+        public IList<Diagnostic> Diagnostics { get; private set; }
```
smedilol, line 188: Diagnostics are never read.
I think these Diagnostics should be returned by CompilationUnit.AllDiagnostics.
So if you activate CFG, you can keep your current code to check Diagnostics.
I know CompilationUnit is in TypeCobol project but we can find a way to "plug" CFG diagnostics into it.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Create a basic block for a node.
+        /// </summary>
+        /// <param name="node">The leading node of the block</param>
+        /// <param name="addToCurrentSentence">true if the block must be added to the current Sentence, false otherwise.</param>
+        /// <returns>The new Block</returns>
+        internal BasicBlockForNode CreateBlock(Node node, bool addToCurrentSentence)
```
fm-117, line 2579: Second parameter can be removed, set to true in every usages.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+        }
+
+        /// <summary>
+        /// Enter a paragraph
+        /// </summary>
+        /// <param name="p">The paragraph to be entered</param>
+        protected virtual void EnterParagraph(Paragraph p)
+        {
+            string name = p.Name;
+            CfgParagraphSymbol sym = new CfgParagraphSymbol(name);
```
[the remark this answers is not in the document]
fm-117: I think there are two ways to see this:
- either we consider the CFG is dependent from SD and then the control flow graph building is some sort of extension of the symbol building process
  - `ControlFlowGraphBuilder` has to know about the `ProgramSymbolTableBuilder` so it can do its own stuff while symbols are built
- or the two processes are independent of each other which means `ControlFlowGraphBuilder` only knows about Node building and never uses symbols
  - CFG builder uses its own representation of section and paragraphs which are not related to symbols

I personnally think the second way is cleaner because it would be simpler to change the SD or the CFG when they are independent (no need to worry about consequences in one feature when modifying the other)

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: It is possible to make the method GetTerminalSuccessorEdges a static, by adding a new parameter the target Cfg instance (ControlFlowGraph<Node, D> Cfg). But this method is really used in a MultiBranchContext logic associated to a Cfg construction not a Basic Block.

TypeCobol/Compiler/Symbols/Symbol.cs
```diff
@@ -76,6 +77,8 @@ public enum Flags : ulong
             InsideTypedef = 0x01L << 35,//Flag of any symbol inside a Typedef definition.
             BuiltinSymbol = 0x01L << 36, //This is a Builtin symbol.
             
+            Declaratives = 0x01L << 37,
```
fm-117, line 80: Actually, this flag is only set, never tested for existence. So it can be removed.

TypeCobol.Analysis/Graph/BasicBlock.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        /// <summary>
+        /// Flag on a basic block.
+        /// </summary>
+        [Flags]
+        public enum Flags : uint
+        {
+            Ending = 0x01 << 0,         //Flag if this basic block is an ending block.
+            Default = 0x01 << 1,        //Flag if this basic block is  default block for instance a WhenOther block.
+            Declaratives = 0x01 << 2,   //Flag if this basic block is inside a declaratives section.
```
fm-117, line 57: Only set.

TypeCobol.Analysis/Graph/BasicBlock.cs
```diff
@@ -0,0 +1,132 @@
...
+            Start = 0x01 << 3,          //Flag for a start node.
+            End = 0x01 << 4,            //Flag for a end node.
```
fm-117, line 59: Only tested for existence (!)

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
smedilol: The second option seems fine for me.
I think ControlFlowGraphBuilder ~should~ must not know about ProgramSymbolTableBuilder.
Builders are internal implementation and are not "public".

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: There is another bug: see : 
https://github.com/TypeCobolTeam/TypeCobol/commit/68c7c477d08e8c5d86fed521be63a3f4ef290c7f

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yes, this for a CFG construction point of view, we need to introduce a kind of virtual  ROOT Section, with a start section and a Start Basic Block. See it as Root entities of the graph.

TypeCobol.Analysis/Cfg/DefaultControlFlowGraphBuilder.cs
mayanje: Yes, but by default to instantiate the TypeVariable  (D) to something object is used for CFG as CFG does not use by default custom DATA, DFA uses its own data.

TypeCobol.Analysis/Graph/CfgDotFileForNodeGenerator.cs
mayanje: This is a overloading for Node instance which uses CodeElement. And thus this is a specialization.
BasicBlock is Generic.

TypeCobol.Analysis/Graph/CfgDotFileForNodeGenerator.cs
mayanje: We are creating a Dot SubGraph for BasicBlockForNodeGroup, to be merge din the main the parent Dot gaph. The Writer is reused (cfgDot.Writer = sw;), but the instruction for directed edges are written at tail construction of the graph in cfgDot.Writer = sw;).

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// <returns>true if ok, false otherwise</returns>
+        public delegate bool BasicBlockCallback(BasicBlock<N, D> block, ControlFlowGraph<N, D> cfg);
+
+        /// <summary>
+        /// Flag on a Cfg.
+        /// </summary>
+        [Flags]
+        public enum Flags : uint
+        {
+            Compound = 0x01 << 0,       //Flag to indicate that this Cfg graph has subgraphs.
```
fm-117, line 28: Turn this unique flag into a boolean property `HasSubgraphs`.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
```diff
@@ -0,0 +1,170 @@
...
+        public List<BasicBlock<N, D>> AllBlocks
+        {
+            get;
+            internal set;
+        }
+
+        /// <summary>
+        /// A map from Node to corresponding basic block.
+        /// </summary>
+        public Dictionary<N, BasicBlock<N,D>> BlockFor
```
fm-117, line 83: This dictionary has its values set but never read. It can be used in the various `ResolvePending*` methods to get the block back from a node, this would allow us to store only nodes to be resolved instead of Tuples.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
mayanje: In fact I though that maybe Cobol declaratives could be seen as separted root,s but in fact not, so in the case of COBOL language this list contains only one element  the Root Basic Block.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
mayanje: I don't understand what you mean.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
mayanje: Happy for _LinkQers_

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
mayanje: idem

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Resolve a section or a paragraph symbol reference
+        /// </summary>
+        /// <param name="symRef">The Symbol Reference instance to a section or a paragraph.</param>
+        /// <returns>The scope of symbols found</returns>
+        internal Container<Symbol>.Entry ResolveSectionOrParagraphSymbol(SymbolReference symRef)
```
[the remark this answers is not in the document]
mayanje: This a new Issue, because current CrossCheck does not support what CFG has implemented. This is not a refactorization, but a port to CrossChecker to use either what CFG does or what SDLite will implement to support also qualified Section and Paragraph symbol resolution. So that CFG can move to the port or SDLite implementation, because CFG has every thing to do that.

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
```diff
@@ -0,0 +1,1181 @@
...
+    {
+        public NodeListenerFactory CfgBuilderNodeListenerFactory = null;
+
+        public static DefaultControlFlowGraphBuilder CfgBuilder;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            //Allocate a static Default Control Flow Graph Builder
+            CfgBuilderNodeListenerFactory = () =>
```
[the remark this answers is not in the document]
mayanje: ok

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
```diff
@@ -0,0 +1,1181 @@
...
+            string expectedPath = Path.Combine(Directory.GetCurrentDirectory(), "DotOutput", "IfThenAfter0.dot");
+
+            Assert.IsTrue(CfgBuilder.AllCfgBuilder.Count == 1);
+            Assert.IsNotNull(CfgBuilder.AllCfgBuilder);
+
+            CfgTestUtils.GenDotCfgAndCompare(CfgBuilder.Cfg, path, expectedPath);
+        }
+
+        [TestMethod]
+        public void IfThenElseTest()
```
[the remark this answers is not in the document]
mayanje: hum... factorize what all tests have in common......

TypeCobol.Analysis.Test/BasicCfgInstrs/Alter0.cbl
```diff
@@ -0,0 +1,24 @@
+﻿       IDENTIFICATION DIVISION.
```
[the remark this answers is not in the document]
mayanje: hum I followed the current way of doing of many TypeCobol Tests, but I tried to give the same base name than the input test name.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Check that a Section or a Paragraph is resolvable
+        /// </summary>
+        /// <param name="node">A reference node for the check, used as position.</param>
+        /// <param name="symRef">The Symbol Reference to the Section os Paragraph</param>
+        /// <returns>The Symbol of the section or Paragraph if resolved, null otherwise.</returns>
+        private Symbol CheckSectionOrParagraph(Node node, SymbolReference symRef)
```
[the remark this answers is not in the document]
mayanje: Already answered

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
mayanje: :-)

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -0,0 +1,2695 @@
...
+                Symbol targetSymbol = null;
+                IEnumerable<CfgSentence> sentences = ResolveSectionOrParagraphSentences(@goto, sref, out targetSymbol);
+                if (targetSymbol == null)
+                {
+                    Diagnostic d = new Diagnostic(MessageCode.SemanticTCErrorInParser,
+                        @goto.CodeElement.Column,
+                        @goto.CodeElement.Column,
+                        @goto.CodeElement.Line,
+                        string.Format(Resource.UnknownSectionOrParagraph, sref.ToString()));
+                    Diagnostics.Add(d);
```
fm-117, line 1489: Duplicate diagnostic with the one directly created in `CheckSectionOrParagraph` method.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
smedilol: I think we can keep Flags enum, even if there is only 1 value.
Has we already have a flag mechanism in a lot of class, it will be consistent.
And it'll be there if we need to add other values.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
smedilol: All things considered, I prefer the second option:

> or the two processes are independent of each other which means ControlFlowGraphBuilder only knows about Node building and never uses symbols
CFG builder uses its own representation of section and paragraphs which are not related to symbols

CfgParagrahSymbol and CfgSectionSymbol should not inherent from Symbol classes but rather "attach" information to Symbol.
I see 2 ways to do this:
- Manage a Dictionary in CFG to map a Symbol to additional info
- Add a Dictionary<string, object> into Symbol class. string = free identifier, object = information to attach. This dictionary can be use by any class to "attach" information to a Symbol.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: ok

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
fm-117: So are you ok to turn it into a single field `RootBlock` instead of a List ?

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
fm-117: Here we are calling inside a `foreach` the overload of the method which instantiate the BitArray `discovered`. The array has the size of Allblocks which means it does not depend on where the traversal start, it accomodates with the whole graph no matter what. So to avoid creating it multiple times, I suggested to call `DFS(root, discovered, callback)` and create the discovered array outside of the foreach. This is not required though if we switch to a single RootBlock, this method will simply disappear.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
fm-117: Yes I like Linq 😄 , but is is also a matter of readability. It took me a bit of time to understand we  are iterating over all sections but first and using only first sentence of each.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.DeclarativesContext.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.MultiBranchContext.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.MultiBranchContext.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: Removed the 'clone' wording.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: In fact we have a skype meeting about the explnation, here MultiBranchContext instance are those created by WhenConditionClause corresponding to When nodes, and here are not associated to a When node that has no CodeElement. Thus when the ctx.Instruction will be different than null, it will be the ctx corresponding to the evaluate.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yes this still a point to clarify how CFG diagnostics are reported or accessed by callers. So We discussed to have a CfgContext, which can be given a DiagnosticsDispatcher interface instance.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: hum...

TypeCobol/Compiler/Symbols/Symbol.cs
mayanje: No the initial purpose was to flag BasicBlock entry  of Declaratives, I thought this was done, but It seems that I forgot to do that. So one could check if the instruction of the block is declarative, but I think it's better to flag such blocks. So it still to be done.

TypeCobol.Analysis/Graph/BasicBlock.cs
mayanje: Still to be used.

TypeCobol.Analysis/Graph/BasicBlock.cs
mayanje: Still to used, for instance the End flag is used to detect terminal block that contains a GOTO instruction. Default flag is used in whenCondBlock. No these flags are usefull.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
mayanje: Sure

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
mayanje: Hum.. block for node is usefull for instance for:

https://github.com/TypeCobolTeam/TypeCobol/pull/1493
https://github.com/TypeCobolTeam/TypeCobol/issues/1449

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: ok

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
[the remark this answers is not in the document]
fm-117: done

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
fm-117: Fixed through code factorization.

TypeCobol.Analysis/Graph/BasicBlock.cs
fm-117: Ok but it has to be applied somewhere. How do you define an End block ? And how is it different from the `Ending` flag ?

TypeCobol.Analysis.Test/BasicControlFlowInstructionTest.cs
fm-117: Added diagnostics comparison in latest commits.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgParagraphSymbol.cs
fm-117: Solved with PR #1738, CFG does not use Symbols anymore.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSectionSymbol.cs
fm-117: See PR #1738.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSectionSymbol.cs
fm-117: See PR #1738.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
fm-117: See PR #1738, Sentence keeps its list of blocks.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
fm-117: See PR #1738.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
fm-117: See PR #1738. You're right, it's not the index of the block but its index in the successors list.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.CfgSentence.cs
fm-117: See PR #1738. As Sentence is not a Symbol anymore, this counter has been removed.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: Done in PR #1738.

TypeCobol.Analysis.Test/BasicCfgInstrs/Alter0.cbl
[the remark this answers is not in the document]
fm-117: Done

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
fm-117: CFG now uses its own classes for sentences, sections and paragraphs and they are not related to SemanticDomain.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: After PR #1738, the parameter is now used once.

TypeCobol/Compiler/Symbols/Symbol.cs
fm-117: See PR #1738, Symbol.cs reverted to its state in develop branch.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
fm-117: Done thanks to #1736, CFG and CrossCheck now use the same method which also supports qualified name resolution.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
fm-117: And done 😄

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
fm-117: Diagnostic has been deleted.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
fm-117: As discussed together, the array still needs to be instantiated because we need to reset the discovered nodes each time we start a new traversal from a specific root block.

TypeCobol.Analysis/Graph/BasicBlock.cs
fm-117: As discussed, the `End` flag is meant to identify a single end node for the whole graph. This is not implemented yet but an option could be added to build an additional node in the graph to which all `Ending` nodes would point.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.BasicBlockForNode.cs
fm-117: Well actually, the `Tag` property is defined for the derived class `BasicBlockForNode` so it already acknowledges a Cobol (or TypeCobol) context because it uses the `Node` class to describe instructions. In Cobol context I think it makes sense to describe a block with the section or paragraph in which it appears.

TypeCobol.Analysis/Graph/BasicBlock.cs
mayanje: yes, for instance if we need to represent taht in an IDE.

mayanje (approved):

fm-117:
See rebase version: #1806.

PR #1736, opened 2020-06-09
WI 1710 Check duplicate paragraph names with semantic data owner

rooksdo:
Closes #1710

PR #1738, opened 2020-06-16
1260 cfg semantic domain lite avoid using symbols

fm-117:
First refactoring : avoid using symbols in CFG
- Paragraph/Section/Sentence won't inherit from symbol anymore
- Custom CFG-only objects to represent Paragraph/Section/Sentence
- Temporary resolution mechanism before sharing with CrossChecker (dependent on PR #1736)


TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Procedure.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace TypeCobol.Analysis.Cfg
+{
+    public partial class ControlFlowGraphBuilder<D>
+    {
+        /// <summary>
+        /// Base class for Paragraphs and Sections.
+        /// A procedure is a target of a GOTO or PERFORM, it has a name and holds sentences.
+        /// </summary>
+        private abstract class Procedure : ProcedureDivisionPartition
```
mayanje, line 9: I think Procedure is not the good term Section and Paragraph are not only for PERFORM procedure but also for GOTO label. So I think this is a _Label_ or a _Part (Partition)_, maybe _PartitionLabel_.

mayanje: I think another term could be ProcedureDivisionLabel, thus these are Labels, but if you keep the term  Partition then the class Procedure should but named Part.

fm-117: I took the term 'Procedure' from the specifications. See page 260 of Language Reference (v6.2)

> Within the PROCEDURE DIVISION, a procedure consists of a section or a group of
sections, and a paragraph or group of paragraphs.
A procedure-name is a user-defined name that identifies a section or a paragraph.

smedilol: +1 to include the terme `label` in the name.
I prefer ProcedureDivisionLabel because it's more explicit.
`Partition` can be related to a machine partition.

mayanje: So I think we shoud rename the class Procedure to ProcedureName.

fm-117: Ok but we are talking about two different things.
- the `ProcedureDivisionPartition` class is an abstraction for paragraphs, sections and sentences. Sentences don't have any name or label so `ProcedureDivisionLabel` doesn't fit. Other alternatives were `ProcedureDivisionPart`, `ProcedureDivisionRegion` and `ProcedureDivisionZone`
- the `Procedure` class is an abstraction for paragraphs and sections. It does not refer to the name of the section or paragraph but the section or paragraph _itself_ so `ProcedureName` is not suitable.

mayanje: Ok so another term is also PrecedureDivisionArea.

fm-117:
> Why the refactoring of Cfg Tests is not in this PR? I really appreciate what you have done.

What do you mean ? The refactoring of CFG tests is already commited in main branch 1260_CFG_SemanticDomainLite. I've created this sub-PR to focus the discussion on the changes in the builder related to the abandon of symbols.

mayanje:
Ok, good idea for CFG tests refactoring.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Procedure.cs
smedilol: I didn't see the comment of fm-117 when I commented before.
So Cobol specs wins.
I prefer to keep `Procedure` rather to rename to `ProcedureName` because the class represents the whole procedure content (name and its paragraph/sentences).

fm-117:
> Ok, good idea for CFG tests refactoring.

Thanks 😄

mayanje:
The fact is 1260_CFG_SemanticDomainLite is still to be validated, before this one.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Procedure.cs
mayanje: Hum.. don't include sentence, as I said sentence cannot be the target of a procedure, cobol spec does not include in the case sentences.

fm-117: Yep It doesn't include sentences, here is hierarchy:
- ProcedureDivisionRegion
  - Procedure
    - Paragraph
    - Section
  - Sentence

mayanje: Thus the class Procedure could be renamed ProcedureName according to  Cobol spec..

mayanje: But you are right Procedure is the instance, Section and Paragrpah have a ProcedureName property.

PR #1741, opened 2020-07-02
1260 cfg semantic domain lite integrate as analyzer

fm-117:
This PR focuses on actual integration of CFG/DFA as a new kind of analyzer instead of using `NodeDispatcher` listeners (which are now totally removed).

Diagnostics need to be improved
- add test for recursive performs
- add test for "out-of-bounds performs"
- add test for incorrect order of paragraphs in a PERFORM THRU
- add test for invalid ALTER statements
- review diagnostics emitted for IX105A et SG102A samples
  - duplicates ?
  - how to improve messages ?


TypeCobol.Analysis.Test/BasicCfgInstrs/PerformGoesOutOfbounds0.dot
```diff
@@ -0,0 +1,36 @@
...
+Block4 -> Block6
+
+}
+Block1 -> Block4 [style=dashed]
+
+Block2 [
+label = "{Block2|    goback\l}"
+]
+Block0 -> Block1
+Block1 -> Block2
```
smedilol, line 34: The code contains an error:
`Line 4[12,12] <30, Error, Semantics> - Semantic error: Statement '    go to b' located at line 10, column 12 prevents this PERFORM statement to reach its exit.`

IBM Compiler only reports a warning, because the code can still be executed.
Execution will produce an abend U4038, because there  is no goback at the end of paragraph `b`. But if there is a goback, the execution is ok.
So shall we create an error or a warning? I think it should be a warning to behave like IBM compiler.
We can then ask our users if they want an error. In this case the error could be set as parameter of CLI/LSP (like parameter `diagnostic.checkEndAlignment`)


For the CFG : the goback statement can never be reached, but the dot file shows otherwise.
Line Block1 -> Block2 should not appear in the dot file.

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformThru2.dot
```diff
@@ -0,0 +1,97 @@
...
+Block2 -> Block3
+Block3 -> Block4
+Block4 -> Block5
+Block5 -> Block6
+Block6 -> Block7
+Block7 -> Block8
+Block8 -> Block9
+Block9 -> Block10
+
+}
```
rooksdo, line 97: Is it normal that there's no Block11 declared?

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformGoesOutOfbounds0.dot
fm-117: Ok to turn this diagnostic into a Warning.

About the Block1 to Block2 edge, this edge is created while building the graph. It is only after resolving all GOTO/ALTER/PERFORM that we could find out which part of the code is unreachable or not.
The graph is built as to contain all and every possible paths but we need a post-build analysis to actually find which parts are unreachable. Do you confirm @mayanje ?

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformThru2.dot
fm-117: I don't get the meaning of your question 😕 
During parsing, the CFG builder will create new block each time a new sentence is encountered and also when branching is required. After building, blocks may be cloned if they participate in a PERFORM target group. When generating the dot file however, only blocks connected to the root are considered. Some blocks created during parsing may finally not appear in the graph.
Do you think the resulting graph is wrong/missing some elements ?

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformGoesOutOfbounds0.dot
mayanje: In my opinion this case the error is that the perfom goes out of paragraph a and thus the goback will never be executed, but this a consequence of the **abend U4038** and not the execution flow. So you are right to stay iso with IBM Compiler it is a warging (a Severe Warning).
The unreachability of the goback instruction can be detected by an abstract interpretation in this case, because we must detect that finally we will face an **abend U4038**.

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformThru2.dot
rooksdo: I was checking the tests from the .dot files directly, I was wondering about the numbering as I couldn't see any block11 when block10 and block12 were declared.

TypeCobol.Analysis.Test/BasicCfgInstrs/IX105A.diag
```diff
@@ -12,18 +12,38 @@ Line 421[12,47] <37, Warning, General> - Warning: "end-if" is missing
 Line 423[12,46] <37, Warning, General> - Warning: "end-if" is missing
 Line 436[12,47] <37, Warning, General> - Warning: "end-if" is missing
 Line 437[12,42] <37, Warning, General> - Warning: "end-if" is missing
+Line 455[12,12] <37, Warning, General> - Warning: Statement '       GO TO WRITE-TEST-GF-01-FAIL' located at line 485, column 15 prevents this PERFORM statement to reach its exit.
```
smedilol, line 15: IBM Compiler doesn't report for this GOTO.
What I suspect based on test:
Englobing section `SECT-RC-02-001` is called and not the paragrah `WRITE-TEST-GF-01` which contains the GO TO.
The GO TO is inside the same section so the execution flow will continue until it reaches the end of the section.
After that the next instruction after the call to the section will be called.
Same rules applies for perform THRU, if the GO TO doesn't jump after the "THRU paragraph/section ", then it's ok.


Note that in this case, there is no perform to section `SECT-RC-02-001`. All section/paragraphs of the program are called in sequence.

fm-117: This is in fact a limitation of IBM compiler in this case because the incriminated `GO TO` is inside a `WRITE` condition. IBM compiler does not report premature `PERFORM` exit whereas our CFG implementation does.

smedilol: Yes, indeed. It works as expected.

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformProcedure0.dot
```diff
@@ -35,72 +35,72 @@ label = "{Block4|    PERFORM B-PROCESS\l}"
...
-Block14 -> Block15
+Block31 [
+label = "{Block31|}"
+]
+Block26 -> Block27
+Block26 -> Block28
+Block27 -> Block28
+Block28 -> Block29
+Block29 -> Block30
+Block30 -> Block31
```
rooksdo, line 101: Graph doesn't show loop on block30 for Perform until (as for Perform varying in PerfromNested0)

TypeCobol.Analysis.Test/BasicCfgInstrs/SearchPerform0.dot
```diff
@@ -0,0 +1,90 @@
...
+label = "{Block1|    PERFORM SMPL-SEARCH\l}"
+]
+subgraph cluster_1{
+color = blue;
+label = "SMPL-SEARCH";
+Block16 [
+label = "{SMPL-SEARCH. Block16|    DISPLAY 'SEARCH EXAMPLE....'\l    SET SEQ TO 1\l}"
+]
+Block17 [
+label = "{Block17|    SEARCH SUBJECT\l}"
```
rooksdo, line 23: Why Block17 not directly included in  Block16 (as in Start0 Block15 contains the START instruction) ?

fm-117: Block17 has to be independent so as to have the returning edge from Block18. This is specific to SEARCH statement which acts as a loop.

PR #1755, opened 2020-08-07
WI #1747 Use TypeDefinition to improve overload resolution

### LINKED ISSUE #1747: Function overload fails with nested type in dependancy file
### Description
Function overload seems to fail when call instruction references a typed variable whose type is nested in another type declared in a dependency file.

Consider the following dependency declaring two function overloads, the second referencing type Type01 nested in type GroupType
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DPDCY01.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 Type01 typedef strict public.
          05 myPic       pic X.
       01 GroupType typedef strict public.
          05 subType     type Type01.
       PROCEDURE DIVISION.
       declare procedure FunTest public
                                 input myInput pic X.
       data division.
       procedure division.
                continue.
       end-declare.
       declare procedure FunTest public
                                 input myInput type Type01.
       data division.
       procedure division.
                continue.
       end-declare.
       end program DPDCY01.
```
Now consider the following program trying to use a call instruction on both versions of `FunTest` function
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. PGMTEST.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 alpha     pic X.
       01 myVar     type DPDCY01::GroupType.
       PROCEDURE DIVISION.
       INIT-LIBRARY.
           Continue.
*     OK this signature is recognized
       call DPDCY01::FunTest input alpha.
*     Fail the call instruction is expecting an alphanumeric value and not a Type01 variable
       call DPDCY01::FunTest input myVar::subType.
       end program PGMTEST.
```
### Expected behavior
The second call instruction `call DPDCY01::FunTest input myVar::subType.` should also be considered valid by the parser.

fm-117:
Fixes #1747.
The fix uses the TypeDefinition and DataType (instead of DataType only) to improve overload resolution.
Also include some renaming.

### DIFF
--- a/TypeCobol/Compiler/CodeModel/SymbolTable.cs
+++ b/TypeCobol/Compiler/CodeModel/SymbolTable.cs
@@ -1036,12 +1036,7 @@ namespace TypeCobol.Compiler.CodeModel
             var argumentType = argumentTypeInfo.TypeDefinition;
             if (argumentType == null)
             {
-                var argumentTypes = GetType(argumentTypeInfo.DataType);
-                if (argumentTypes.Count > 1)
-                {
-                    return false;
-                }
-                argumentType = argumentTypes.FirstOrDefault();
+                return parameterType == null && parameter.DataType == argumentTypeInfo.DataType;
             }
 
             return parameterType == argumentType;
--- a/TypeCobol/Compiler/CodeModel/SymbolTable.cs
+++ b/TypeCobol/Compiler/CodeModel/SymbolTable.cs
@@ -938,12 +938,12 @@ namespace TypeCobol.Compiler.CodeModel
             Add(Functions, function);
         }
 
-        public List<FunctionDeclaration> GetFunction(StorageArea storageArea, ParameterList profile = null)
+        public List<FunctionDeclaration> GetFunction(StorageArea storageArea, IProfile profile = null)
         {
             return GetFunction(storageArea.SymbolReference, profile);
         }
 
-        private List<FunctionDeclaration> GetFunction(SymbolReference symbolReference, ParameterList profile = null)
+        private List<FunctionDeclaration> GetFunction(SymbolReference symbolReference, IProfile profile = null)
         {
             var uri = new URI(symbolReference.Name);
             return GetFunction(uri, profile);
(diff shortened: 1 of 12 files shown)


TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -999,46 +999,52 @@ public List<FunctionDeclaration> GetFunction(QualifiedName name, ParameterList p
...
-            var p2Types = this.GetType(p2);
 
-            if (p1Types.Count > 1 || p2Types.Count > 1)
-                return false; //Means that a type is declare many times. Case already handle by checker.
-            var p1Type = p1Types.FirstOrDefault();
-            var p2Type = p2Types.FirstOrDefault();
+            var parameterType = parameter.TypeDefinition;
+            if (parameterType == null)
+            {
+                var parameterTypes = parameter.Name.Contains(".") ? this.GetType(parameter) : this.GetType(parameter.DataType, pgmName);
```
smedilol, line 1028: Is the call to `GetType` still necessary here?

TypeCobol/Compiler/CodeModel/SymbolTable.cs
```diff
@@ -999,46 +999,52 @@ public List<FunctionDeclaration> GetFunction(QualifiedName name, ParameterList p
...
+                }
+                parameterType = parameterTypes.FirstOrDefault();
+            }
 
-            if (p1Type != p2Type)
-                return false;
+            var argumentType = argumentTypeInfo.TypeDefinition;
+            if (argumentType == null)
+            {
+                var argumentTypes = GetType(argumentTypeInfo.DataType);
```
smedilol, line 1039: Is the call to GetType still necessary here? In my other comment, GetType is called for the parameter but here it's for the argument.
Arguments should always be checked by TypeCobolLinker.

TypeCobol/Compiler/CodeModel/SymbolTable.cs
fm-117: This is required for ParameterDescription coming from a dependency. At this step, the TypeDefinition may not be resolved yet.

TypeCobol/Compiler/CodeModel/SymbolTable.cs
fm-117: See the latest commit, in fact the argumentType may still be null if it is not a TC typed variable but a primitive DataType. In that case, instead of comparing the TypeDefs, I use the DataTypes. Seems to work fine except for one test where a diagnostics disappear. To discuss.

smedilol: One diagnostic about a duplicate function disappear but it's ok. 
The second function declaration is not valid. The error about the duplicate it's not totally accurate, because the second bad declaration is like an overload.

PR #1765, opened 2020-09-01
Fix for iterative and recursive PERFORMs

fm-117:
Based on 62a1c96823295193eb44b8293f2e17f9a1e1f785 and adapted to this branch using SemanticDomain Lite.


fm-117:
A bug has been identified in the clonedPerforms handling. This sample produces 2 recursivity diagnostics although the code isn't recursive.

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZZMFT1.
       DATA DIVISION.
       PROCEDURE DIVISION.
       main.
           PERFORM a
           GOBACK
           .
       a.
           PERFORM b
           PERFORM c
           .
       b.
           PERFORM d
           .
       c.
           PERFORM d
           .
       d.
           DISPLAY "end !"
           .
       END PROGRAM DVZZMFT1.
```

mayanje (commented):
Thus to validate this PR, I prefer wating for the last issue to be fixed, using the fix provided in branch:
https://github.com/TypeCobolTeam/TypeCobol/tree/1260_TypeCobolCFG

fm-117:
Now includes latest fixes from [1260_TypeCobolCFG](https://github.com/TypeCobolTeam/TypeCobol/tree/1260_TypeCobolCFG)
- avoid multiple outgoing edges
- fix recursion detection

TypeCobol.Analysis/Graph/CfgDotFileForNodeGenerator.cs
```diff
@@ -144,6 +132,7 @@ protected override bool EmitBasicBlock(BasicBlock<Node, D> block, int incomingEd
                 {
                     sw.WriteLine(string.Format("Block{0} -> \"\" [style=dashed]", block.Index));
```
mayanje, line 133: @fm-117 So as we are always using explicit iterative perform procedure style, we can display a plain arrow style. This can be detected using IsExplicitIterativeGroup property on the group.

fm-117: Actually this is not used. The flag `GroupGrafted` is used in Extended mode and we keep the dashed style in Normal mode so I've removed property `IsExplicitIterativeGroup`. Dot files stay the same and are in-sync with those from [1260_TypeCobolCFG](https://github.com/TypeCobolTeam/TypeCobol/tree/1260_TypeCobolCFG).

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformProcIterativeAfterRecursive0.dot
```diff
@@ -0,0 +1,78 @@
...
+}
+Block1 -> Block10 [style=dashed]
+
+Block2 [
+label = "{Block2|    goback\l}"
+]
+Block14 [
+label = "{Block14|    else\l       display 'not done'\l}"
+]
+Block0 -> Block10
```
smedilol, line 68: Here Block0 should point to block1 `perform pararec test after varying n from 1 by 1 until n > 3`.

smedilol (changes requested):

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformProcedure2.dot
```diff
@@ -0,0 +1,133 @@
...
+Block30 -> Block41 [style=dashed]
+
+Block31 [
+label = "{Block31|}"
+]
+Block26 -> Block27
+Block26 -> Block28
+Block27 -> Block28
+Block28 -> Block29
+Block29 -> Block41
```
smedilol, line 91: Same problem here, the block29 should point to block30

smedilol (changes requested):

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformProcIterativeAfterRecursive1.dot
```diff
@@ -0,0 +1,78 @@
...
+]
+Block15 [
+label = "{Block15|    display 'terminate'\l}"
+]
+Block14 [
+label = "{Block14|    else\l       display 'not done'\l}"
+]
+Block10 -> Block11
+Block10 -> Block14
+Block11 -> Block21
```
rooksdo, line 58: Following the same logic as @smedilol's remarks, we should have block11->block12. So, the instruction TEST AFTER should not affect the end graph.

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformProcedure2.dot
fm-117: This is not bug but a feature 😆 ! The difference between PerformProcedure0 and PerformProcedure2 is that the `PERFORM B-LOOP` instruction uses a `TEST AFTER` clause. This PR introduces specifically some change to represent performs having a TEST AFTER clause.

TypeCobol.Analysis.Test/BasicCfgInstrs/PerformProcIterativeAfterRecursive0.dot
fm-117: So after our discussion, this should be adressed through a separate issue (#1783)

fm-117:
About the last two commits :
- in GEFDOT, an empty subgraph leads to a rectangle drawn around all blocks of the graph (whereas in Graphviz, nothing is drawn for empty subgraphs). So we now check the content of a subgraph before writing it to the output file.
- in DOT (in general), if an edge belongs to a subgraph, then its target block will be included into this subgraph. This would result in PERFORM blocks being forcingly relocated into subgraphs. The fix here consists in writing edges only at the root level. As a conssequence, subgraphs now never have edges, only blocks.

This is not perfect though, there is still a bug with those blue rectangles : for example in `PerformProcIterativeAfterRecursive2.dot`, no rectangle is drawn at all because no subgraph is generated. This is related to the current handling of AFTER clauses. So I think it is better to leave it as it is and change that with #1783 altogether.

rooksdo (approved):
Ok with @fm-117, the known limitations of graphic representation should be addressed in #1783. This PR is ok as it is.

mayanje (approved):

smedilol (approved):

PR #1764, opened 2020-08-26
WI #1762 Remember index hash on StorageArea and use it during code ge…

fm-117:
…neration
Fixes #1762.
This is more a hack than a fix because it involves adding a Codegen-specific field into `DataOrConditionStorageArea` class.

Issue #1762, opened 2020-08-21
Call instruction with input parameter referencing list index doesn't generate hash


rooksdo:
### Description
This is the follow up of #1718 :
- Call instruction with input parameter referencing list index doesn't generate hash

```cob
       identification division.
       program-id. pgm.
       data division.
       global-storage section.
       01 MyList                        type TypeA.
       working-storage section.
       01 TypeA  typedef strict private.
          05 filler occurs 10 indexed by idxA.
             10 flag          type Bool.
      *-----------------------------------------
       procedure division.

      *--> KO 'MyList.idxA' generated instead of hash
           call myproc input MyList::flag(MyList::idxA)

      *--> KO 'idxA' generated instead of hash
           call myproc input MyList::flag(idxA)
           .

       declare procedure myproc private
                   input myBool        type bool.
           procedure division.
              goback.
       end-declare.

       end program pgm.
```
Here's the corresponding generated code :
```cob
      *--> KO 'MyList.idxA' generated instead of hash
      *    call myproc input MyList::flag(MyList::idxA)
           CALL 'ba67825b' USING
                                 flag-value IN MyList(MyList.idxA)
           end-call                        

      *--> KO 'idxA' generated instead of hash
      *    call myproc input MyList::flag(idxA)
           CALL 'ba67825b' USING
                                 flag-value IN MyList(idxA)
           end-call
```

fm-117:
Fixed in 1.4.3, need to be copied to develop as well.

PR #1771, opened 2020-09-28
WI #1762 Fix index hash in CALL arguments

rooksdo:
Remember index hash on StorageArea and use it during code generation (PR #1764)


rooksdo:
Draft PR to check differences with develop.

PR #1779, opened 2020-10-07
WI #1778 Fix numeric literal detection in scanner after THROUGH/THRU …

### LINKED ISSUE #1778: Range value in condition levels are not parsed correctly
**Describe the bug**
Some value declarations are not parsed correctly and lead to an exception.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 item PIC 9(4).
          88 ranges VALUE 0001 THRU
                          0005.
       procedure division.
           goback
           .
       END PROGRAM MyPGM.
```

**Expected behavior**
No error.

Check this code, the keywords THROUGH and THRU should be added to distinguish between literals and level number in that case.
https://github.com/TypeCobolTeam/TypeCobol/blob/a7793a0f925ebbf0cb5353ca9238adaac45c28d1/TypeCobol/Compiler/Scanner/Scanner.cs#L1563-L1569

fm-117:
…keyword

Fixes #1778.

### DIFF
--- a/TypeCobol/Compiler/Scanner/Scanner.cs
+++ b/TypeCobol/Compiler/Scanner/Scanner.cs
@@ -3,16 +3,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
-using Antlr4.Runtime.Atn;
 using JetBrains.Annotations;
 using TypeCobol.Compiler.Concurrency;
 using TypeCobol.Compiler.Diagnostics;
 using TypeCobol.Compiler.Directives;
 using TypeCobol.Compiler.File;
-using TypeCobol.Compiler.Nodes;
-using TypeCobol.Compiler.Parser.Generated;
 using TypeCobol.Compiler.Text;
-using String = System.String;
 
 namespace TypeCobol.Compiler.Scanner
 {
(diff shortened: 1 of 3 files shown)


TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -1561,12 +1557,23 @@ bool GuessIfCurrentTokenIsLevelNumber()
...
+                                    case TokenType.VALUES:
+                                    case TokenType.THROUGH:
+                                    case TokenType.THRU:
+                                        currentTokenIsExpectedToBeALiteral = true;
+                                        break;
+                                    case TokenType.IS:
+                                    case TokenType.ARE:
+                                        currentTokenIsExpectedToBeALiteral =
+                                            beforeLastSignificantToken.TokenType == TokenType.VALUE ||
+                                            beforeLastSignificantToken.TokenType == TokenType.VALUE;
```
smedilol, line 1574: I think a `S` is missing:
```beforeLastSignificantToken.TokenType == TokenType.VALUES```

File: TypeCobol/Compiler/Scanner/Scanner.cs [hunk 2]
```diff
@@ -1561,12 +1557,23 @@ bool GuessIfCurrentTokenIsLevelNumber()
...
+                                {
+                                    case TokenType.OCCURS:
+                                    case TokenType.VALUE:
+                                    case TokenType.VALUES:
+                                    case TokenType.THROUGH:
+                                    case TokenType.THRU:
+                                        currentTokenIsExpectedToBeALiteral = true;
+                                        break;
+                                    case TokenType.IS:
+                                    case TokenType.ARE:
```
smedilol, line 1571: This PR now allow to use syntax `VALUE ARE` and `VALUES IS`.
Can you crate an issue about it and add an unit test?

smedilol (changes requested):

TypeCobol/Compiler/Scanner/Scanner.cs
fm-117: done

TypeCobol/Compiler/Scanner/Scanner.cs
fm-117: See issue #1782. Reverted to original code to separate between this PR and the "VALUES IS"/"VALUE ARE" issue.

rooksdo (approved):

smedilol (approved):

Issue #1752, opened 2020-08-04
Handle copy replacing with ==::==


smedilol:
**Describe the bug**
Our parser cannot use standard Cobol Copy with a replacing clause containing string `::`.
```cobol
       01  xxxENT. COPY YxxxENT REPLACING ==::== BY ====. 
```
YxxxENT contains:
```cobol
           05  xxxENT::-FCT01.
               10 xxxENT::-FCT01-Var1            PIC X(05).
```

The parser interpret `::` as TypeCobol operator but inside a replacing clause (Pseudo text), this operator should be considered as plain text.

**Expected behavior**
No error.
The expanded source is:
```cobol
       01  xxxENT. 
           05  xxxENT-FCT01.
               10 xxxENT-FCT01-Var1            PIC X(05).
```

**Technical**
See Scanner class and use property `IsInsidePseudoText` of the ScanState to inhibit creation of token `::` inside pseudo text.

fm-117:
Testing for `IsInsidePseudoText` won't be enough, we'll have to allow the creation of a PartialCobolWrod for text `::` and double colon must be replaced correctly while parsing the copy (as for now `::` will be materialized as a QualifiedNameSeparator token).
It means that interpretation of the text `::` inside a copy depends on the active REPLACING clause.
See #428 for other out-of-ordinary REPLACING clauses.
One way to tackle this could be to avoid creating original tokens of a copy but only create the replaced ones. This is a rather large change of the Scanner though...

rooksdo:
To be in conformity with the specifications, replacing clause should handle any placeholder between ==[placeholder]==

smedilol:
@fm-117 @mayanje Goal for this issue: In the scanner:
Inside a copy, do not recognize the `::` operator. Only recognize two `:`  in a row.

fm-117:
Reopening because current solution does not allow direct parsing of a copy without initial ScanState.

fm-117:
So I have created [this branch](https://github.com/TypeCobolTeam/TypeCobol/tree/1752_ImproveInitialScanStateHandling) to change when the initial scan state is created (see 809508738e7947cae2b3924f5ed7d38463b0cb5e) but all this will certainly change again with support of direct copy parsing, so no PR for the moment.

fm-117:
See #1870, the handling of initial scan state will be done with direct copy parsing.

PR #1787, opened 2020-10-23
WI #1752 Implements Replace :: by two characters :.

mayanje:
**What have been done.**

- A new class GroupToken to represent for instance the token :: as a group of two tokens ':'
- At REPLACE directive creating time, the GoupToken is replaced by its sub-tokens
- The ImportedTokensDocument class that is used as token iterator for handling the REPLACE, now create a preprocessed text fragment which is then scanned again to get the new set of tokens to be iterated.
- The error message of an empty pseudo text of comparison, which was generated by the Scanner is now generated by the CompilerDirectiveBuilder class.


TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -54,11 +54,23 @@ public class CompilationDocument
         public List<RemarksDirective.TextNameVariation> CopyTextNamesVariations { get; set; }
 
         public List<CopyDirective> MissingCopies { get; set; }
+        /// <summary>
+        /// Is this Compilation Document for a Copy.
+        /// </summary>
+        internal bool IsForCopy
+        {
+            get;
+            set;
```
fm-117, line 63: `CompilationUnit` is for programs so this property can be made readonly and set at construction time (default true, false when called from CompilationUnit constructor)

TypeCobol/Compiler/CupPreprocessor/CompilerDirectiveBuilder.cs
```diff
@@ -172,13 +183,42 @@ public virtual void EnterCopyCompilerStatement(TypeCobol.Compiler.Scanner.Token
...
+                    else
+                    {
+                        flatList.Add(t);
+                    }                   
+                }
+                return flatList;
+            }
+            return list;
+        }
+        private bool BuildReplaceOperation(IList<ReplaceOperation> replaceOperations, ref Token comparisonToken, ref Token[] followingComparisonTokens, ref Token replacementToken, ref Token[] replacementTokens, bool replaceTokens, List<Token> operandTokens)
```
fm-117, line 214: The ANTLR-related duplicate method in `TypeCobol.Compiler.Preprocessor.CompilerDirectiveBuilder` has not been updated.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -32,6 +32,14 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
             set { _useEuroInformationLegacyReplacingSyntax = value; }
         }
+        /// <summary>
+        /// Are these Options used for parsing a Copy.
+        /// </summary>
+        public bool AreForCopyParsing
+        {
+            get;
+            set;
+        }
```
fm-117, line 42: I'm not sure if it is a new property in `TypeCobolOptions`, why not introduce an additional parameter in scanning method ?

TypeCobol/Compiler/Preprocessor/ImportedTokensDocument.cs
```diff
@@ -44,17 +47,67 @@ public ImportedTokensDocument(CopyDirective copyDirective, ProcessedTokensDocume
...
+                        line = t.Line;
+                        tokenLine = t.TokensLine;
+                        sb.Append(t.Text);
+                    }
+                    string preprocessedFRagment = sb.ToString();
+
+                    //Now reparse the preprocessed fragment
+                    ITextDocument initialTextDocumentLines = new ReadOnlyTextDocument(SourceDocument.TextSourceInfo.Name, DocumentFormat.RDZReferenceFormat.Encoding, DocumentFormat.RDZReferenceFormat.ColumnsLayout, preprocessedFRagment);
+                    TypeCobolOptions tcOptions = new TypeCobolOptions();                    
+                    tcOptions.AreForCopyParsing = true;//Doing that any "::" will be treated as two tokens
```
fm-117, line 97: `tcOptions` is not used.

TypeCobol/Compiler/Preprocessor/ImportedTokensDocument.cs
```diff
@@ -44,17 +47,67 @@ public ImportedTokensDocument(CopyDirective copyDirective, ProcessedTokensDocume
...
+                    FileCompiler fileCompiler = new FileCompiler(initialTextDocumentLines, null, null, new TypeCobolOptions(), null, false, null);
+                    fileCompiler.CompilationResultsForProgram.InitialScanStateForCopy = state;
+                    fileCompiler.CompilationResultsForProgram.UpdateTokensLines();
```
fm-117, line 100: Do we really need a whole FileCompiler ? Using `Scanner` only would be cleaner.

TypeCobol/Compiler/Preprocessor/ProcessedTokensDocument.cs
```diff
@@ -77,9 +77,14 @@ public IEnumerable<Token> ProcessedTokensSource
         /// - COPY directives text imports
         /// - REPLACE directive token remplacements
         /// </summary>
-        public static ITokensLinesIterator GetProcessedTokensIterator(TextSourceInfo textSourceInfo, ISearchableReadOnlyList<IProcessedTokensLine> lines, TypeCobolOptions compilerOptions)
+        /// <param name="textSourceInfo"></param>
+        /// <param name="lines"></param>
+        /// <param name="compilerOptions"></param>
+        /// <param name="allowWhitespaceTokens">This parameters is used to force whitspaces tokens to be also returned.</param>
+        /// <returns></returns>
+        public static ITokensLinesIterator GetProcessedTokensIterator(TextSourceInfo textSourceInfo, ISearchableReadOnlyList<IProcessedTokensLine> lines, TypeCobolOptions compilerOptions, bool allowWhitespaceTokens)
```
fm-117, line 85: The new `allowWhitespaceTokens` parameter is always `false` except for one usage, we could use a default value.

TypeCobol/Compiler/Scanner/GroupToken.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TypeCobol.Compiler.Scanner
+{
+    /// <summary>
+    /// A class that represents a Group Of Tokens.
+    /// </summary>
+    public class GroupToken : Token
```
fm-117, line 12: `TokenGroup`  class already exists. Can we use it instead (with adaptations if need be) ?

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -1144,10 +1148,10 @@ public Token GetNextToken()
...
 
                             delimiterToken = new Token(TokenType.PseudoTextDelimiter, startIndex, startIndex + 1, usesVirtualSpaceAtEndOfLine, tokensLine);
-                            if (!(followingChar == ' ' || followingChar == ',' || followingChar == ';' || followingChar == '.'))
-                            {
-                                tokensLine.AddDiagnostic(MessageCode.InvalidCharAfterPseudoTextDelimiter, delimiterToken);
-                            }
+                            //if (!(followingChar == ' ' || followingChar == ',' || followingChar == ';' || followingChar == '.'))
+                            //{
+                            //    tokensLine.AddDiagnostic(MessageCode.InvalidCharAfterPseudoTextDelimiter, delimiterToken);
+                            //}
```
fm-117, line 1154: Commented code may be confusing for future readers, better remove it.

fm-117 (changes requested):

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -959,11 +959,15 @@ public Token GetNextToken()
...
-                    if (currentIndex < lastIndex && line[currentIndex + 1] == ':')
+                    if (currentIndex < lastIndex && line[currentIndex + 1] == ':' && (this.compilerOptions == null || !this.compilerOptions.AreForCopyParsing))
                     {
-                        // consume two :: chars
-                        currentIndex += 2;
-                        return new Token(TokenType.QualifiedNameSeparator, startIndex, startIndex + 1, tokensLine);
+                        // consume two :: chars in token1 and token2
+                        var token1 = ScanOneChar(startIndex, TokenType.ColonSeparator);
+                        var token2 = ScanOneChar(startIndex + 1, TokenType.ColonSeparator);
+                        GroupToken group = new GroupToken(TokenType.QualifiedNameSeparator, startIndex, startIndex + 1, tokensLine);
```
smedilol, line 967: Why use a GroupToken for token `::` now ?

mayanje: Group of Tokens is needed here in order to still returning one token when GetNextToken() is called, The REPLACE parsing will flat the group into it's groups tokens, and I just wanted to still returning TokenType.QualifiedNameSeparator.
But In fact maybe Group of Token can be avoid if we consider the flag  this.compilerOptions.AreForCopyParsing for a Pure Cobol Treatment not TypeCobol.

TypeCobol/Compiler/CupPreprocessor/CompilerDirectiveBuilder.cs
smedilol: I think we can remove the Antlr Preprocessor.
I've created #1788 about that.

smedilol (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/CopyReplace4ColonErr.rdz.cbl
```diff
@@ -0,0 +1,10 @@
+﻿       IDENTIFICATION DIVISION.
+       PROGRAM-ID. CPYRPL4C.
+      
+       ENVIRONMENT DIVISION.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01  xxxENT. COPY YxxxENT REPLACING ==== BY ====.
+       PROCEDURE DIVISION.
+           MOVE 'A' TO xxxENT-FCT01-Var1.
```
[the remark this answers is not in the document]
mayanje: Yes, I need also further test.

TypeCobol/Compiler/Scanner/Scanner.cs
mayanje: So this code is commented because, I still needed to know how to emit this diagnostics in  a right context.

TypeCobol/Compiler/Scanner/GroupToken.cs
mayanje: Yep... yes you are right I must test it.

TypeCobol/Compiler/Preprocessor/ProcessedTokensDocument.cs
mayanje: I do that in order to not miss same places where it is called, but at the end, I think yes.

TypeCobol/Compiler/Preprocessor/ImportedTokensDocument.cs
mayanje: This was a  question that I asked to myself, the answer is that it is FileCompiler that connect all things welll.

TypeCobol/Compiler/Preprocessor/ImportedTokensDocument.cs
mayanje: Yes it must be replaced by the new TypeCobolOptions()

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -959,11 +959,15 @@ public Token GetNextToken()
                 case ':':
                     // -- TypeCobol specific syntax --
                     // QualifiedNameSeparator => qualifierName::qualifiedName
-                    if (currentIndex < lastIndex && line[currentIndex + 1] == ':')
+                    if (currentIndex < lastIndex && line[currentIndex + 1] == ':' && (this.compilerOptions == null || !this.compilerOptions.AreForCopyParsing))
                     {
```
smedilol, line 963: Can you try the following solution:
Modify this condition:
```C#
if (!currentState.InsidePseudoText //No QualifiedNameSeparator allowed in pseudoText
                        && (we are not in a COPY) //No QualifiedNameSeparator allowed in COPY 
                        && currentIndex< lastIndex && line[currentIndex + 1] == ':')
                    {
```

Delete line 2307 in original Scanner.cs, because now we allow empty pseudoText `==::==`
```C#
            // no legal cobol word chars found 
            if (index == startIndex + 1 && !CobolChar.IsCobolWordChar(line[index])) return false;
```
Keep your fix in TypeCobol/Compiler/Scanner/Token.cs
Still return a bool from method BuildReplaceOperation in TypeCobol/Compiler/CupPreprocessor/CompilerDirectiveBuilder.cs to report Diagnostic : "REPLACE Empty Comparison Pseudo Text." and "REPLACE Empty Pseudo Text Delimiter"

Otherwise all of the rest of the C# code can be deleted.

I didn't test Codegen, but it worked on Parser part.

fm-117:
Although original issue focuses on COPY - REPLACING, could you add a test with a REPLACE ? I think REPLACING and REPLACE should have the same behavior.

smedilol (changes requested):

smedilol (changes requested):

TypeCobol/Compiler/Scanner/Scanner.cs
smedilol: After more investigation, characters :: that need to be replaced, must be generated as a partial Cobol Word Token.
E.g. :
`move W-Var1:: to W-Var2`
`W-Var1::` must be single a single Token of type Partial Cobol Word Token. Otherwise the replacing won't occurs.

TypeCobol/Compiler/Scanner/Scanner.cs
mayanje: Yes it works like that!

mayanje: I am going to create a new PR and give up this one. I didn't know Partial Cobol Word pattern and how it works :-|

mayanje:
@smedilol So The last requested test fails with smedilol method but succeed with the method of this PR, I am looking why.
The one with Copy YDVZOSM.

mayanje:
As a new solution has been suggested by @smedilol a new pull request has been created:
https://github.com/TypeCobolTeam/TypeCobol/pull/1794

TypeCobol/Compiler/Preprocessor/CopyTokensLinesIterator.cs
```diff
@@ -50,17 +50,18 @@ private struct CopyTokensLinesIteratorPosition
...
         private object snapshotPosition;
-        private bool AllowWhitespaceTokens;
+        // Allow the caller to Filter some tokens.
+        private Func<Token, bool> TokenFilterCallback;
 
         /// <summary>
         /// Set the initial position of the iterator with startToken.
         /// </summary>
-        public CopyTokensLinesIterator(string sourceFileName, IReadOnlyList<IProcessedTokensLine> tokensLines, int channelFilter, bool allowWhitespaceTokens)
+        public CopyTokensLinesIterator(string sourceFileName, IReadOnlyList<IProcessedTokensLine> tokensLines, int channelFilter, Func<Token, bool> tokenFilterCallback)
```
fm-117, line 59: Both `channelFilter` and `tokenFilterCallback` are used to select/discard tokens during iteration so they should be merged together, adding a default value (filtering based on channel like before the change).

mayanje:
Replaced by PR #1794

mayanje:
Handle by PR #1794

PR #1793, opened 2020-10-28
WI #1789 Allow modification on input indices in procedures

### LINKED ISSUE #1789: Error on input table index SET statement
Given the following typedef
```cob
01 Tab typedef strict public.
      Item occurs 200 indexed by itemIdx.
```
The following assignments should be allowed inside a procedure declaration
```cob
declare procedure Foo public
            input MyInput type Tab.
*This statement should be allowed but is KO
   Set itemIdx          to 1
*This statement should be allowed but is KO
   Perform varying MyInput::itemIdx from 1 by 1 until MyInput::itemIdx > 200
      display "hi"
   End-Perform
end-declare
```
To avoid having these statements in error the user needs to shift to an in-out argument instead of input.

fm-117:
Fixes #1789


smedilol:
I let @rooksdo check if everything is ok from a Cobol point of view.

rooksdo (approved):

mayanje (approved):

Issue #599, opened 2017-07-06
Replaced Token doesn't work inside DataDivision


smedilol:
Consider this example:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOZCHKP.
       data division.
       working-storage section.
       REPLACE                                                          
           ==:LONGUEUR-IDOFUS:==      BY ==0900==.                                                     
       01  var1         PIC X(:LONGUEUR-IDOFUS:).                                                                           
      
       PROCEDURE DIVISION.
           move :LONGUEUR-IDOFUS: to var1
           .
      
       END PROGRAM TCOZCHKP.
```

ConsumedToken for CodeElement of Var1 contains `:LONGUEUR-IDOFUS:`
ConsumedToken for CodeElement of move contains `0900`

It seems Token are not replaced correctly when inside DataDivision.

smedilol:
@mayanje I assigned you because it seems to be the same bug as #1753

rooksdo:
Fixed by #1798

PR #1797, opened 2020-11-02
WI #599 Use a MultilineScanStateSnapshot

mayanje:
**What have been done.**

- A new class MultilineScanStateSnapshot have been introduced to capture a Scan state associated to a PartialCobolWord. A Snapshot is used to capture values of calculated properties.

- The MultilineScanStateSnapshot instance is used in : ReplaceTokensLinesIterator::CreateReplacedTokens(...) method and the call to the Scanner.Scanner.ScanIsolatedTokenInDefaultContext(...) method.

- New comparison results use Symbol's type Dump

### DIFF
--- a/TypeCobol.Test/TestCollection.cs
+++ b/TypeCobol.Test/TestCollection.cs
@@ -11,12 +11,14 @@ using TypeCobol.Test.Parser.Scanner;
 using TypeCobol.Test.Parser.Text;
 using TypeCobol.Test.Utils;
 
-namespace TypeCobol.Test {
+namespace TypeCobol.Test
+{
 
     [TestClass]
     public class TestCollection
     {
         static readonly string root = PlatformUtils.GetPathForProjectFile("Parser" + Path.DirectorySeparatorChar + "Programs");
+        static readonly string rootNew = PlatformUtils.GetPathForProjectFile("Parser" + Path.DirectorySeparatorChar + "ProgramsNew");
 
         [TestMethod]
         [TestProperty("Time", "fast")]
@@ -126,7 +128,8 @@ namespace TypeCobol.Test {
         [TestMethod]
         [TestCategory("Parsing")]
         [TestProperty("Time", "fast")]
-        public void CheckCodeElements() {
+        public void CheckCodeElements()
+        {
             // Test the recognition of potentially ambiguous CodeElements which begin with the same first Token
             TestCodeElements.Check_DISPLAYCodeElements();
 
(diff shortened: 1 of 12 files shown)


TypeCobol.Test/TestCollection.cs
```diff
@@ -11,12 +11,14 @@
...
 
-namespace TypeCobol.Test {
+namespace TypeCobol.Test
+{
 
     [TestClass]
     public class TestCollection
     {
         static readonly string root = PlatformUtils.GetPathForProjectFile("Parser" + Path.DirectorySeparatorChar + "Programs");
+        static readonly string rootNew = PlatformUtils.GetPathForProjectFile("Parser" + Path.DirectorySeparatorChar + "ProgramsNew");
```
smedilol, line 21: Why create a new folder for programs ?

mayanje: Because I use a new kind of dump that uses new Semantic Domain Symbol Type ToString. Thus based of this new folder I can use this new dump style, and let current dump style as it is.

TypeCobol/Compiler/Scanner/MultilineScanStateSnapshot.cs
```diff
@@ -0,0 +1,119 @@
...
+            _AfterFUNCTION = currentState.AfterFUNCTION;
+            _AtBeginningOfSentence = currentState.AtBeginningOfSentence;
+        }
+
+        private bool _AfterExecSql;
+        public override bool AfterExecSql
+        {
+            get
+            {
+                return AfterExecSql;
```
fm-117, line 33: Typo, beware the property is now recursive !

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -540,22 +540,23 @@ public static void ScanTokensLineContinuationGroup(IList<TokensLine> continuatio
...
+        /// Scan an isolated token given a a Scan State.
+        /// If the given scan state is null then Scan in performed in the following "default" context :
         /// - insideDataDivision = true
         /// - decimalPointIsComma = false
         /// - withDebuggingMode = false
         /// - encodingForAlphanumericLiterals = IBM 1147
         /// - default compiler options
         /// </summary>
-        public static Token ScanIsolatedTokenInDefaultContext(string tokenText, out Diagnostic error)
+        public static Token ScanIsolatedTokenInDefaultContext(MultilineScanState scanState, string tokenText, out Diagnostic error)
```
fm-117, line 551: This does not fit with the name of the method, the purpose of `ScanIsolatedTokenInDefaultContext` is to provide a default context so the caller does not have to consider the scan state. So maybe it is better to create a new method `ScanIsolatedToken` which accepts a scan state parameter (as a bonus, ScanIsolatedTokenInDefaultContext will be able to call this new method providing its own default scan state).

TypeCobol/Compiler/Scanner/MultilineScanStateSnapshot.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TypeCobol.Compiler.Scanner
+{
+    /// <summary>
+    /// Snapshot of a  MultilineScanState
+    /// </summary>
+    public class MultilineScanStateSnapshot : MultilineScanState
```
fm-117, line 12: So basically `MultilineScanStateSnapshot` is a readonly `MultilineScanState`. I don't think we need this new class, you can set the scan state of a partial cobol word with a clone of the current relevant scan state. this is more or less what we are already doing with the `InitialScanState` property in `TokensLine`.

TypeCobol/Compiler/Scanner/Token.cs
```diff
@@ -295,8 +295,13 @@ public virtual string Text
         /// </summary>
         public LiteralTokenValue LiteralValue { get; set; }
 
+        /// <summary>
+        /// ScanState associated to this token if any, null otherwise.
+        /// </summary>
+        public MultilineScanState ScanStateSnapshot { get; set; }
```
fm-117, line 301: Better make it internal as it is used during parsing but not necessarily a result to be exposed to clients. Besides only PartialCobolWord tokens will get this property set so it is confusing for callers.

fm-117 (changes requested):

TypeCobol/Compiler/Scanner/MultilineScanStateSnapshot.cs
mayanje: I will try a clone, I use property values because most of them are calculated properties.

TypeCobol/Compiler/Scanner/MultilineScanStateSnapshot.cs
mayanje: Yep, yes

TypeCobol.Test/TestCollection.cs
smedilol: In this case, you must use a new Comparator instead of a folder.
As stated by @fm-117, the SYM comparator seems to be what you need.

smedilol (changes requested):

TypeCobol/Compiler/Scanner/Scanner.cs
smedilol: I agree too, but for me after this PR, ScanIsolatedTokenInDefaultContext must not be used anymore.
As the problem is that we always need a context to scan token.

fm-117: @smedilol is right, this is the only usage of the method, so renaming it is enough.

mayanje: So the idea here is to create a private method, and use two methods with distinct names that call the private one. But the call to call the methods is performed at only one place, thus I prefer rename it: for instance ScanIsolatedTokenInGivenContextorDefault

mayanje:
Replaced by #1798

PR #1809, opened 2020-11-23
1808 report missing copies all levels

### LINKED ISSUE #1808: Complete the missing copies list with all levels of inclusion
**Only the first level of missing copies is displayed**
When a program references a copy that contains another copy, if the embedded copy is missing it won't display in the missing copy list.

**How to reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZZMFT0.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       COPY COPY1.
       PROCEDURE DIVISION.
           DISPLAY final-var OF group3 OF group2 OF group1
           GOBACK
           .
       END PROGRAM DVZZMFT0.    
```
COPY1
```cob
 01 group1.
       COPY COPY2.
```
COPY2
```cob
    05 group2.
          COPY COPY3.
```
COPY3
```cob
       10 group3.
                15 final-var PIC X.
```

**Expected result**
Missing copys at all levels of inclusion should be listed in the Missing Copies list.

fm-117:
Fixes #1808 
But does not implement full error message with indication of including copy. I could not find a simple way to keep track of the including copy so I pushed what I got so far.


CLI/test/ressources/checkMissingCopies_2/output_expected/error.txt
```diff
@@ -0,0 +1,4 @@
+ReturnCode=MissingCopy_
+2 errors in "input\MissingCopy.rdz.tcbl":
+Line 5[8,18] <46, Error, Directives> - Dependent COPY COPY3 not found
```
smedilol, line 3: There is no indication that it's **COPY2** which need COPY3 and COPY4?

rooksdo (commented):
Seems good so far. Should we put aside finding the parent copy or is it manageable to find it?

CLI/test/ressources/checkMissingCopies_2/output_expected/error.txt
fm-117: As I explained it in this PR description, I could not find a simple way to keep the info.
When we first detect it, we know it's COPY2 that needs COPY3 (and COPY4) but with the current code this info is lost for upper levels. To keep it, we would have to change the `List<CopyDirective> missingCopies` into something that holds the link between the missing copy and its including copy (if any). this has impacts on several classes/methods.

fm-117: You know have the choice with either this solution or #1815 which supports parent copy in message.

smedilol:
As suggested by @rooksdo we can close this PR as the new PR #1815 works better.

Issue #1811, opened 2020-11-26
LSP Signature Help Request on the first line raise exception


mayanje:
**Description**
In the TypeCobol Plugin Invoking a signature help on the first line of a Document raise an unexpected exception from the LSP server.

**How to reproduce**

- Open a document in the TypeCobol RDz plugin

- Enter an empty first line in the document

- Put the caret in the first line of the document

- Perform Ctrl+Shift+Space action
The following popup will be raised.
![image](https://user-images.githubusercontent.com/19149269/100375634-68271800-300e-11eb-8eef-5cd2f8e1b6ba.png)

PR #1812, opened 2020-11-26
WI #1622 Add unhandled exception handler.

### LINKED ISSUE #1622: Send notification for all logged exceptions
**Description**
Follow up to #518. 

Send a notification to TCOB taskforce for all logged exceptions. This generalizes the present behavior with CLI exceptions.
Thus all uncaught exceptions must be notified by LSP notification and by telemetry, from the LSP server.

mayanje:
A handler on Unhandled Exception has been added.
This feature will be testable using the following scenario.

- Issue https://github.com/TypeCobolTeam/TypeCobol/issues/1811 can be used for that purpose.
- In TypeCobol notification mail are only sent if the TypeCobol related version is a release version not a debug one.
- The current TypeCobol RDz Plugin does not send -t (Use statistics option) to the LSP server thus TFS issue 94607 must be fixed first for instance in develop branch and test RDz plugin develop version.

Fixes #1622


smedilol (approved):

TypeCobol.LanguageServer/JsonRPC/IRPCServer.cs
```diff
@@ -29,5 +29,10 @@ public interface IRPCServer
         /// Write a trace in the server log file
         /// </summary>
         void WriteServerLog(string trace);
+
+        /// <summary>
+        /// An Unhandled Exception handler
+        /// </summary>
+        UnhandledExceptionEventHandler Handler { get; set; }
```
fm-117, line 36: Maybe use an event instead of a delegate property ? It works but an event is more descriptive of what you want to achieve.
```csharp
event UnhandledExceptionEventHandler UnhandledException;
```

TypeCobol.LanguageServer/VsCodeProtocol/LanguageServer.cs
```diff
@@ -41,15 +41,42 @@ public LanguageServer(IRPCServer rpcServer)
...
+        }
+
+        /// <summary>
+        /// Destructor.
+        /// </summary>
+        ~LanguageServer()
+        {
+            this.RpcServer.Handler -= UnhandledExceptionHandler;
+            AppDomain currentDomain = AppDomain.CurrentDomain;
+            currentDomain.UnhandledException -= UnhandledExceptionHandler;
```
fm-117, line 59: I don't think we need to use the `AppDomain.UnhandledException` event if we aready have our own logic for handling exceptions in the server object but we do have to ensure that this last chance handling is always safe so add a try-catch block around the `NotifyException` call in the handler.
Also why attempt to unregister the handler in a finalizer, the application is shutting down anyway.

fm-117 (changes requested):

TypeCobol.LanguageServer/JsonRPC/IRPCServer.cs
mayanje: event is interresting if we sent events here we just use and handler for propagating already sent exceptions.

TypeCobol.LanguageServer/VsCodeProtocol/LanguageServer.cs
mayanje: AppDomain.UnhandledException is interresting to capture all exceptions that we don't know yet.

mayanje: The destructor here to ensure safe programming.

TypeCobol.LanguageServer/JsonRPC/IRPCServer.cs
fm-117: Ok so the question is: which object shuld be responsible of the handling of those unhandled exceptions ?
If it is the IRPCServer instance itself then a method `void OnUnhandledException(Exception e)` in the interface is appropriate. If the handling should be done externally (and is optional) then an event is best.

mayanje: If you mean that another one than LanguageServer can register itself to listen unhandled exception ok event mechanism shall be used, otherwise no use, only one handler shall be set.

fm-117: No I mean a class different from the implementor of IRPCServer, which is the case here. LanguageServer handles the exceptions of JsonRPCServer.

mayanje: First the fact is that class LanguageServer receives in its constructor an IRPCServer interface instance.
Declaring and even in the interface in IRPCServer, will make JsonRPCServer implementation provides add and remove method and then finally use a UnhandledExceptionEventHandler delegate instance.

So in this case event is the equivalent of the delegate UnhandledExceptionEventHandler, with less code using delegate.

TypeCobol.LanguageServer/VsCodeProtocol/LanguageServer.cs
fm-117: Ok I think I see now, we have 2 kinds of scenarios: some exceptions are already handled but not properly notified (those are the ones in the various catch block of JsonRPCServer and we also have unpredictable exceptions (those can be handled with the AppDomain.UnhandledException event as a last resort. 

It would be best to remove the Handler property from IRPCServer interface and add a new parameter in JsonRPCServer constructor but it cannot currently easily be done. So ok to keep the property.

As for the destructor, I maintain my previous remark. There is no need to detach handling methods because in our current architecture, if LanguageServer dies, it means the whole process is exiting.

smedilol (approved):
For the desctructor, yes I agree that it's unuseful, but anyway it's working.

fm-117 (approved):

rooksdo (approved):

PR #1815, opened 2020-11-30
WI #1808 Include missing copys from all levels

fm-117:
This is an alternate solution for #1808.
Unlike #1809, this modification allows to track including document so as to report a more descriptive diagnostic.


rooksdo (dismissed):

rooksdo:
@smedilol 
This solution seems to conform with expected result. Should we close #1809

smedilol (dismissed):

mayanje (approved):

smedilol (approved):

rooksdo (approved):

TypeCobol/Compiler/Diagnostics/MessageCode.cs
```diff
@@ -54,6 +54,7 @@ public enum MessageCode
         MultiFormalizedCommentIndicatorMisused = 44,
         Info = 45,
         AnalyzerFailure = 46,
-        QualityRuleViolation = 47 //MessageCode 47 is reserved for violations coming from any analyzer
+        //MessageCode number 47 is reserved for violations coming from a Quality analyzer, it has no corresponding DiagnosticMessage object.
```
smedilol, line 57: I don't think this change is related to #1808.
Can you revert it ?
Same remarks on Violation.cs

smedilol (changes requested):

TypeCobol/Compiler/Diagnostics/MessageCode.cs
fm-117: This is required. Initially it worked because the reserved code was last in the csv file, now with the new MessageCode/MessageDiagnostic we need to use a dictionary and account for the reserved code.
If I revert this latest commit, some unit tests will break. I can change the commit message to attach it to #1837 though.

fm-117:
See rebase version #1855.

Issue #1816, opened 2020-12-02
IAnalyser and new execution step for quality rules


smedilol:
### The need
We plan to develop more quality rules with our parser.

These quality rules will be done on a private repository, so we need a way to "plug" quality rules inside the parser.
A quality rule must be executed when all symbol have been resolved.
These quality rules could become slow. Code completion for LSP must remains as fast as possible.
As Completion within LSP requires  CrossCheck execution step to be complete, we need a new execution step to be sure that quality rules will not be executed.

Diagnostics of quality rules and all others diagnostic must be sent by LSP 500ms after no update of texte have been received by LSP.
When completion is requested, only diagnostics up to CrossCheck step will be sent by LSP.

### Raw solution
Extend the IAnalyser to provide a way to create quality rules.
This IAnalyser will be executed in a new step after CrossCheck.

### Out of perimeter of this issue:
- Load quality rules from an external DLL. See #1825.
- Define how a quality rule return information. Fow now we use Diagnostics.

fm-117:
- [ ] Introduce new compilation step `QualityCheckStep`
- [ ] Introduce new analyzer type `IASTAnalyzer` (extending both `IAnalyzer` and `IASTVisitor`) and update analyzer provider mechanism
- [ ] On LSP, modify `RefreshSyntaxTree` method
  - add new boolean parameter `refreshCodeAnalysis`
  - when set to true, call the new execution step on the FileCompiler instance
  - this new boolean will be set to true for timer event and node refresh action
  - for the other usages, this would depend of the current request (see usages of `GetDocumentContextFromStringUri`)

Issue #1818, opened 2020-12-03
TC LSP create cfg_dfa_info custom notification.


mayanje:
**Purpose**
For displaying a Control Flow Graph with Data Flow Information, we need a custom LSP notification from the server to the client.
Such information can be:

- The Generated Dot File Path
- For each Basic Block the range in the target document of each instruction
- Any DFA information to be reported

This Issue must:

- Implements the custom LSP notification's type and params

- Add an option to enable or disable CFG/DFA information to be calculated.

- Send the notification to the client, after each successfull semantic analysis on a document, if option enabled.

PR #1819, opened 2020-12-04
WI #1816 Add new ExecutionStep for code quality checking

fm-117:
Work in progress for issue #1816.


TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -189,7 +189,7 @@ protected virtual void OnDidReceiveSignatureHelpContext(SignatureHelpContextPara
         /// <param name="bForced">Force the server to send the program OutlineNodes</param>
         protected virtual void OnDidReceiveRefreshOutline(string uri, bool bForced)
         {
-            var context = GetDocumentContextFromStringUri(uri, false);
+            var context = GetDocumentContextFromStringUri(uri, Workspace.SyntaxTreeRefreshLevel.RebuildNodes);
```
smedilol, line 192: Why not 
```suggestion
            var context = GetDocumentContextFromStringUri(uri, Workspace.SyntaxTreeRefreshLevel.NoRefresh);
```
?

fm-117: I thought it was a bug because the outline needs Nodes to be displayed correctly, no ?

smedilol: Yes indeed, I thought about the case when the server send an update of the Outline to the client.
But as I unterdstand it, this case is when the client want to refresh the whole Outline.

fm-117: Well i'm not sure now 🤨, because the `OnDidReceiveRefreshOutline` method is called in two spots:
- as an event handler when the ProgramClassDocument has changed
- as part of `OnDidSaveTextDocument` method

So it is not a notification/request method but an internal mechanism to refresh the outline data and then sends the data to the client. But it is still unclear for me if we need to rebuild nodes... Also why call this method in the DidSaveTextDocument process if we already have a ProgramClass refresh scheduled ?

TypeCobol/Compiler/Parser/InspectedProgramClassDocument.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Snapshot of the fully parsed program that was quality checked
+        /// </summary>
+        public ProgramClassDocument PreviousStepSnapshot { get; }
+
+        /// <summary>
+        /// List of rules violations returned by code analysis
+        /// </summary>
+        public List<Diagnostic> Violations { get; }
```
smedilol, line 26: After discussion, can you rename Violations to Diagnostics and remove:
`public List<Diagnostic> Diagnostics => PreviousStepSnapshot.Diagnostics;`

Currently each Document return only its own Diagnostics.

We need to discuss what a Violation is. We can start by using Diagnostics to return violation of quality rules.

smedilol (changes requested):

TypeCobol/Compiler/Parser/InspectedProgramClassDocument.cs
fm-117: Done with 0c68ca479d644a16680a4f146e5a39c2de6dd933.

TypeCobol/Compiler/CompilationUnit.cs
```diff
@@ -287,9 +288,77 @@ public void ProduceTemporarySemanticDocument()
...
+        public void RefreshCodeAnalysisDocumentSnapshot()
+        {
+            lock (lockObjectForCodeAnalysisDocumentSnapshot)
+            {
+                var programClassDocument = ProgramClassDocumentSnapshot;
+                if (programClassDocument != null && CodeAnalysisDocumentNeedsUpdate())
+                {
+                    PerfStatsForCodeQualityCheck.OnStartRefresh();
+
+                    List<Diagnostic> violations = new List<Diagnostic>();
```
smedilol, line 304: Can you rename violations to diagnostics to be consistent with the rest ?

TypeCobol/Compiler/Parser/InspectedProgramClassDocument.cs
```diff
@@ -0,0 +1,33 @@
...
+using TypeCobol.Compiler.Nodes;
+
+namespace TypeCobol.Compiler.Parser
+{
+    /// <summary>
+    /// Capture results of code analysis.
+    /// </summary>
+    public class InspectedProgramClassDocument
+    {
+        public InspectedProgramClassDocument(ProgramClassDocument programClassDocument, List<Diagnostic> violations)
```
smedilol, line 12: Can you rename violations to diagnostics to be consistent with the rest ?

smedilol (changes requested):

smedilol (approved):

rooksdo (dismissed):

smedilol (approved):

rooksdo (approved):

smedilol (approved):

rooksdo (approved):

mayanje (approved):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
fm-117: Will have to be fixed along with #1818.

fm-117:
See rebased version #1848.

Issue #1750, opened 2020-07-30
Integrate DFA in CFG_SemanticDomainLite


mayanje:
**Description**
The purpose of this issue is to integrate DFA algorithms based on the integration of the CFG in the SemanticDomainLite, this according to the work done in following issues:

#1260 Implement TypeCobol Data Flow Analysis

# Data Flow Analysis (DFA) #
## Data structures and Graphs (DFG) (A06) ##
- #1270 Increase the CFG graph with data structures and fields for DFA. (A06.T01).

## DFG and DFA Development. (A07) ##
- #1271 Identify TypeCobol instructions targeted by DFA. (A07.T01) 
- #1272 Apply the DFA algorithms. (A07.T02) 

## DFG/GraphViz Visualization and Tests. (A08). ##
- #1273 Generate the .dot file of a DFG graph. (A08.T01) 
- #1274 Test DFG dot graphs and visualization. (A08.T02)

# DFA in Practice #
## Identification of the target modules of a CALL. (A09) ##
- #1275 Identify the modules called by a CALL statement. (A09.T01) 

## CFG/DFA TypeCobol Tests. (A10) ##
- #1276 Write CFG and DFA test sets for TypeCobol. (A10.T01)

rooksdo:
Fixed by #1751

PR #1820, opened 2020-12-08
WI #1750 Add DataDefinition nodes having initial value to START Block.

mayanje:
**Purpose**
This Pull request is based on branch 
https://github.com/TypeCobolTeam/TypeCobol/compare/1260_CFG_DFA_SemanticDomainLite_IntegrateAsAnalyzer

To allow Data Definition having values and declared in the DataDivision of a Program to be put as instruction nodes in the START block of the corresponding PROCEDURE DIVISION.
So that their intial values are correctly propagated by DFA algorithms from the START block.

### DIFF
--- a/TypeCobol.Analysis.Test/DfaCallPgmReport.cs
+++ b/TypeCobol.Analysis.Test/DfaCallPgmReport.cs
@@ -59,11 +59,11 @@ namespace TypeCobol.Analysis.Test
         public void ProcCallPgmReportTest()
         {
             string path = Path.Combine(CfgTestUtils.Report, "ProcCallPgm.cbl");
-            var cfgs = ParseCompareDiagnostics<DfaBasicBlockInfo<VariableSymbol>>(path, CfgBuildingMode.WithDfa);
-            Assert.IsTrue(cfgs.Count == 1);
+            DfaTestResults dfaresult = ParseCompareDiagnosticsWithDfa(path);
+            Assert.IsTrue(dfaresult.Graphs.Count == 1);
 
             //Create the report file.
-            ZCallPgmReport reporter = new ZCallPgmReport(cfgs);
+            ZCallPgmReport reporter = new ZCallPgmReport(dfaresult.Graphs);
             using (StringWriter sw = new StringWriter())
             {
                 reporter.Report(sw);
--- a/TypeCobol.Analysis.Test/DfaCallPgmReport.cs
+++ b/TypeCobol.Analysis.Test/DfaCallPgmReport.cs
@@ -54,5 +54,26 @@ namespace TypeCobol.Analysis.Test
                 TypeCobol.Test.TestUtils.compareLines(path, result, expected, output);
             }
         }
+
+        [TestMethod]
+        public void ProcCallPgmReportTest()
+        {
+            string path = Path.Combine(CfgTestUtils.Report, "ProcCallPgm.cbl");
+            var cfgs = ParseCompareDiagnostics<DfaBasicBlockInfo<VariableSymbol>>(path, CfgBuildingMode.WithDfa);
+            Assert.IsTrue(cfgs.Count == 1);
+
+            //Create the report file.
+            ZCallPgmReport reporter = new ZCallPgmReport(cfgs);
+            using (StringWriter sw = new StringWriter())
+            {
+                reporter.Report(sw);
+                // compare with expected result
+                string result = sw.ToString();
+                string output = Path.Combine(CfgTestUtils.Report, "ProcCallPgm.csv");
+                string expected = File.ReadAllText(output, DocumentFormat.RDZReferenceFormat.Encoding);
+                TypeCobol.Test.TestUtils.compareLines(path, result, expected, output);
+            }
+        }
+
     }
 }
--- a/TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
+++ b/TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
@@ -44,16 +44,20 @@ namespace TypeCobol.Analysis.Dfa
                 var dataDivision = this.Cfg.ProcedureDivisionNode.Parent.GetChildren<DataDivision>().FirstOrDefault();
                 if (dataDivision != null)
                 {
-                    foreach(var dataDef in Collect(dataDivision))
+                    var firstInstr = root.Instructions.First;
+                    foreach (var dataDef in Collect(dataDivision))
                     {
-                        root.Instructions.AddFirst(dataDef);
+                        if (firstInstr == null)
+                            root.Instructions.AddLast(dataDef);
+                        else
+                            root.Instructions.AddBefore(firstInstr, dataDef);
                     }
                 }
                 IEnumerable<DataDefinition> Collect(Node node)
                 {
-                    if (node.SemanticData is VariableSymbol variable && variable.Value != null)
+                    if (node.SemanticData is VariableSymbol variable && variable.Value is Value)
                     {
-                        yield return node as DataDefinition;
+                        yield return (DataDefinition)node;
                     }
 
                     if (node.ChildrenCount > 0)
@@ -88,7 +92,7 @@ namespace TypeCobol.Analysis.Dfa
         public override HashSet<VariableSymbol> GetDefVariables(Node node)
         {
             System.Diagnostics.Debug.Assert(node != null);
-            return node is DataDefinition ? new HashSet<VariableSymbol>() { node.SemanticData as VariableSymbol} : GetSymbols(node.StorageAreaWritesSymbol);
+            return node is DataDefinition ? new HashSet<VariableSymbol>() { (VariableSymbol)node.SemanticData} : GetSymbols(node.StorageAreaWritesSymbol);
         }
     }
 }
--- a/TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
+++ b/TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
@@ -4,6 +4,7 @@ using TypeCobol.Compiler.CodeElements;
 using TypeCobol.Compiler.Nodes;
 using TypeCobol.Compiler.Symbols;
 using TypeCobol.Compiler.Types;
+using System.Linq;
 
 namespace TypeCobol.Analysis.Dfa
 {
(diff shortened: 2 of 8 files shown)


TypeCobol.Analysis.Test/DfaCallPgmReport.cs
```diff
@@ -54,5 +54,26 @@ public void InBulkCallPgm88SetReportTest()
                 TypeCobol.Test.TestUtils.compareLines(path, result, expected, output);
             }
         }
+
+        [TestMethod]
+        public void ProcCallPgmReportTest()
+        {
+            string path = Path.Combine(CfgTestUtils.Report, "ProcCallPgm.cbl");
+            var cfgs = ParseCompareDiagnostics<DfaBasicBlockInfo<VariableSymbol>>(path, CfgBuildingMode.WithDfa);
```
fm-117, line 62: You can use ParseCompareDiagnosticsWithDfa as a shortcut

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
```diff
@@ -11,6 +12,40 @@ namespace TypeCobol.Analysis.Dfa
     /// </summary>
     public class DefaultDataFlowGraphBuilder : DataFlowGraphBuilder<Node, VariableSymbol>
     {
+        /// <summary>
+        /// Collector of DataDefinition Nodes in the list of Instructions of a given block.
+        /// </summary>
+        public class DataDefinitionCollector : AbstractSymbolAndTypeVisitor<BasicBlock<Node, DfaBasicBlockInfo<VariableSymbol>> , BasicBlock<Node, DfaBasicBlockInfo<VariableSymbol>> >
```
fm-117, line 18: the `AbstractSymbolAndTypeVisitor` is not required. You can collect nodes using a recursive collector method and the `Node.Children` property.

Something like:

```csharp
public override HashSet<VariableSymbol> GetDeclarations()
{
    var dataDivision = this.Cfg.ProcedureDivisionNode.Parent.GetChildren<DataDivision>().SingleOrDefault();
    return new HashSet<VariableSymbol>(Collect(dataDivision));

    IEnumerable<VariableSymbol> Collect(Node node)
    {
        if (node.SemanticData is VariableSymbol variable && variable.Value != null)
        {
            yield return variable;
        }

        if (node.ChildrenCount > 0)
        {
            foreach (var child in node.Children)
            {
                foreach (var childVariable in Collect(child))
                {
                    yield return childVariable;
                }
            }
        }
    }
}
```

This requires some adaptation to add the node into rootBlock instruction list but you get the idea 😄.

TypeCobol.Analysis/Dfa/ValueOrigin.cs
```diff
@@ -137,6 +136,30 @@ void ProcessSetStatement(SetStatement set)
...
                     }
                 }
+
+                void ProcessDataDefinition(DataDefinition dataDef)
+                {
+                    VariableSymbol vs = (VariableSymbol)dataDef.SemanticData;
+                    if (vs.Value != null)
+                    { //(2.1.1.3.1) Take the initial value and add it to the origins path.
+                        if (vs.Value is Value value)
+                        {
```
fm-117, line 146: I think we can safely assume that dataDef.SemanticData is a non-null VariableSymbol with a non-null Value at that point. No need to test it in an `if`, a `Debug.Assert` is more appropriate.

fm-117 (changes requested):

TypeCobol.Analysis/Dfa/ValueOrigin.cs
mayanje: Hum... this is not for testing dataDef.SemanticData but the initial value.

fm-117: Yes, I know, what I say is that you can call `(Value) ((VariableSymbol) dataDef.SemanticData).Value` without any prior check and get a non-null value because the DataDefinition we'll get here have been previously collected with that criterion. So an Assertion is enough, no need for more `if` testing.

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
mayanje: Ok

TypeCobol.Analysis/Dfa/ValueOrigin.cs
mayanje: Done in e9f226e

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
mayanje: Ok thanks, done in e9f226e

TypeCobol.Analysis.Test/DfaCallPgmReport.cs
mayanje: done in e9f226e

TypeCobol.Analysis/Dfa/ValueOrigin.cs
```diff
@@ -137,6 +136,29 @@ void ProcessSetStatement(SetStatement set)
...
+                        result.Origins.Add(origin);
+                    }
+                    else if (vs.Value is Value[] values && values.Length > 0)
+                    {
+                        var origin = new ValueOrigin(vs)
+                        {
+                            Value = values[0]
+                        };
+                        result.Origins.Add(origin);
+                    };
```
fm-117, line 160: I am not so sure about this part because I've tested this source:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOMFL06.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 somecondition PIC X(5).
          88 cond1  VALUE 'A'.
          88 cond2 VALUES 'B' 'C' 'D'.
          88 cond3  VALUE 'E'.
       PROCEDURE DIVISION.
           DISPLAY "somecondition: " somecondition
           evaluate true
           when cond1 display "cond1"
           when cond2 display "cond2"
           when cond3 display "cond3"
           when other display "other"
           end-evaluate
           GOBACK
           .
       END PROGRAM TCOMFL06.
```

and the output is:
```
somecondition:      
other

```

So it looks like a data condition has no initial value. I think we should remove this `else if` block and also remove the `case CodeElementType.DataConditionEntry` in parent switch.

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
```diff
@@ -32,9 +34,41 @@ private static HashSet<VariableSymbol> GetSymbols(Dictionary<StorageArea, Variab
...
+                    foreach(var dataDef in Collect(dataDivision))
+                    {
+                        root.Instructions.AddFirst(dataDef);
+                    }
+                }
+                IEnumerable<DataDefinition> Collect(Node node)
+                {
+                    if (node.SemanticData is VariableSymbol variable && variable.Value != null)
+                    {
+                        yield return node as DataDefinition;
```
fm-117, line 56: static cast is ok in fact, so `yield return (DataDefinition) node`.

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
```diff
@@ -54,7 +88,7 @@ public override HashSet<VariableSymbol> GetUseVariables(Node node)
         public override HashSet<VariableSymbol> GetDefVariables(Node node)
         {
             System.Diagnostics.Debug.Assert(node != null);
-            return GetSymbols(node.StorageAreaWritesSymbol);
+            return node is DataDefinition ? new HashSet<VariableSymbol>() { node.SemanticData as VariableSymbol} : GetSymbols(node.StorageAreaWritesSymbol);
```
fm-117, line 91: Same here, you can use static cast.

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
```diff
@@ -32,9 +34,41 @@ private static HashSet<VariableSymbol> GetSymbols(Dictionary<StorageArea, Variab
...
+        {
+            if (Cfg != null && Cfg.IsInitialized)
+            {
+                var root = Cfg.RootBlock;
+                var dataDivision = this.Cfg.ProcedureDivisionNode.Parent.GetChildren<DataDivision>().FirstOrDefault();
+                if (dataDivision != null)
+                {
+                    foreach(var dataDef in Collect(dataDivision))
+                    {
+                        root.Instructions.AddFirst(dataDef);
```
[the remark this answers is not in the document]
mayanje: I agree the order has no sens, but  in this case if there are already nodes put in the START block, it is logic to think that definitions came first.

TypeCobol.Analysis/Dfa/ValueOrigin.cs
mayanje: @fm-117 Hum... I had also a doubt about conditions, but as Conditions are also DataDefinition instances, a treat them  with the same logic, but I am curious to know the context of your test: whar is the test which give you such result as output?
Anyway ZcallPgmReport does a special treatment on level 88 so in what context do you use your test?

mayanje: Ok, if as you have tested it is a nonsense, to consider 88 value as read/write value ok.

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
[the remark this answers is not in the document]
fm-117: Yeah ok, in that case you can use `AddBefore` method from `LinkedList` to ensure that the DataDefinitions are added before any instructions of the START block and also maintaining the order of declarations from original source text.

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
mayanje: done in 37ed136

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
mayanje: done in 37ed136

TypeCobol.Analysis/Dfa/ValueOrigin.cs
mayanje: done in 37ed136

TypeCobol.Analysis/Dfa/DefaultDataFlowGraphBuilder.cs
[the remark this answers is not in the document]
mayanje: done in 37ed136

fm-117 (approved):

smedilol (approved):

PR #1826, opened 2020-12-17
WI #1825 Find and load analyzers from external assemblies

### LINKED ISSUE #1825: Support external analyzers
This is a follow up on #1816 .

Current API does not fully support custom analyzers that could be injected into the compilation process.
The goal of this issue is to add a mechanism to load custom code analyzers from external sources.

fm-117:
Fixes #1825.
No error handling yet but it works...

### DIFF
--- a/CLI/src/CLI.cs
+++ b/CLI/src/CLI.cs
@@ -112,6 +112,9 @@ namespace TypeCobol.Server
             var analyzerProvider = new CompositeAnalyzerProvider();
             var reports = RegisterAnalyzers(analyzerProvider);
 
+            //Add external analyzers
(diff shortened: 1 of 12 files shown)


smedilol:
As you say "No error handling yet", shall we validate or do you itendent to complete this PR ?

fm-117:
In fact any error happening during the load of external analyzer will be caught by the default error handling mechanism:
- CLI will be interrupted, message and stacktrace will be reported
- LanguageServer will report the exception as usual unhandled exception
So I think it is ok to leave as it is.

smedilol (approved):

rooksdo (changes requested):
All seems good to me but I wonder, is it possible to have an example of usage in this PR?

rooksdo (approved):

PR #1855, opened 2021-01-13
1808 report missing copies all levels ptd rebase

### LINKED ISSUE #1808: Complete the missing copies list with all levels of inclusion

fm-117:
Fixes #1808.
Rebase version of #1815.

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/DiagnosticMessage.cs
+++ b/TypeCobol/Compiler/Diagnostics/DiagnosticMessage.cs
@@ -52,12 +52,16 @@ namespace TypeCobol.Compiler.Diagnostics
 
         // Load messages table DiagnosticMessages.csv in memory at startup
 
-        public static IList<DiagnosticMessage> GetFromCode { get; private set; }
+        private static readonly Dictionary<MessageCode, DiagnosticMessage> _predefinedMessages;
+        public static DiagnosticMessage GetFromCode(MessageCode code)
+        {
+            System.Diagnostics.Debug.Assert(_predefinedMessages.ContainsKey(code));
+            return _predefinedMessages[code];
+        }
 
         static DiagnosticMessage()
         {
-            GetFromCode = new List<DiagnosticMessage>();
-            GetFromCode.Add(null); // to start message codes at index 1
+            _predefinedMessages = new Dictionary<MessageCode, DiagnosticMessage>();
 
             using (StreamReader tableReader = new StreamReader(GetStreamForProjectFile("Compiler/Diagnostics/Resources/DiagnosticMessages.csv")))
             {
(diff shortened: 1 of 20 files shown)


smedilol (approved):

TypeCobol/Compiler/Diagnostics/DiagnosticMessage.cs
```diff
@@ -52,12 +52,16 @@ internal DiagnosticMessage(Category category, int code, Severity severity, strin
...
 
-        public static IList<DiagnosticMessage> GetFromCode { get; private set; }
+        private static readonly Dictionary<MessageCode, DiagnosticMessage> _predefinedMessages;
+        public static DiagnosticMessage GetFromCode(MessageCode code)
+        {
+            System.Diagnostics.Debug.Assert(_predefinedMessages.ContainsKey(code));
+            return _predefinedMessages[code];
+        }
 
         static DiagnosticMessage()
```
mayanje, line 62: I think is time to make this static constructor thread safe, because I think that the first thread that will create a diagnostic will execute this code, but if will filling _predefinedMessages another thread creates a diagnostic there will be a concurrent access here, for instance LSR and Analyzers can be in differents threads, but's ok if it is same thread that run analyzers in the QualityCheck step. No other thread should create a diagnostics outside the compiler thread steps.

fm-117: Well, we're in luck then because static constructors are thread-safe in .NET. I could not find the doc reference but this is discussed [here](https://github.com/dotnet/docs/issues/10243). Basically first thread that needs to use the type will run the static constructor and other threads are on hold until the type is properly initialized.

mayanje: Ok so that's good.

mayanje (approved):

Issue #428, opened 2017-02-16
Tokens replaced need the MultilineScanState to determine their type


smedilol:
Number scanned after a REPLACE has been evaluated are considered as Token of type LevelNumber.

```cobol
000000 IDENTIFICATION DIVISION.
000000 PROGRAM-ID. TestReplace.
000000 DATA DIVISION.       
000000 working-storage section.
000000 replace ==:MAX:==  by ==0020== .
000000 01 Array.
000000    05 Elt1  pic X  Occurs :MAX:.                                       
000000 PROCEDURE DIVISION.
000000  .
```
Here `:C-NBR-MAX:`  will be replaced by `0020`but will be considered as LevelNumber instead of Number.

The cause is method `Scanner. ScanIsolatedTokenInDefaultContext` create a new `MultilineScanState` with property `InsideDataDivision` to true.
Then properties `LastSignificantToken` and `BeforeLastSignificantToken` are lost.
Further the Scanner will change the Token type because it thinks he's at the beginning of a sentence.
`(if(tokensLine.ScanState.InsideDataDivision && tokensLine.ScanState.AtBeginningOfSentence)).`

After discussion with @laurentprudhon  : 

> Quotes from him:

> 
> When scanning replacement tokens with REPLACE directive, we are by definition out of context.
> 
> This replacement tokens can be theoretically inserted anywhere in the soure code.
> The same replacement token whose text is `01` can be inserted as a LevelNumber (`01 TBL-TAUCCAU` or a numeric literal `TBL-TAUCCAU VALUE 01`).

> To handle all this cases, we should in ReplaceTokensLinesIterator :
> -	Compare the text of token to replace and not their type
> -	Inject text of replacement tokens in the line (or lines) et scan again all tokens in the context. Currently we only use the remplacement token
> -	Do the same on all successive lines until we the scanner state is synchronized again
> 
> Implement such a behavior can be very bad for performance, especially if we imagine incremental compilation in the future.
> 
> To be more pragmatic, this behavior is only necessary if we replace a token whose type is affected by the scanner state.
> It's the case only with the following replacement :
> -	On keywords : DATA, DECLARATIVES, END, FILE, ID, NEXT, PROCEDURE, SERVICE, WHEN
> -	levelNumber > IntegerLiteral, paragraphNameDefinition > UserDefinedWord, sectionNameDefinition > UserDefinedWord
> ```cobol
> REPLACE == PROCEDURE == BY == MOVE == ?
[...]

PR #1857, opened 2021-01-13
WI #428 Handle token scan state dependent in DataDivision.

mayanje:
This fix handle the case when a replace occurs inside a DataDivision and the replaced token is scan state dependent:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOZCHKP.
       data division.
       working-storage section.
       REPLACE
                ==C-NBRPCB==      BY ==6==.  
       01  W-IDTPCB OCCURS C-NBRPCB    PIC X(8).      
      
       END PROGRAM TCOZCHKP.
```

### DIFF
--- a/TypeCobol/Compiler/Scanner/Token.cs
+++ b/TypeCobol/Compiler/Scanner/Token.cs
@@ -78,7 +78,8 @@ namespace TypeCobol.Compiler.Scanner
             //Scan Dependent Tokens Inside DataDivision must have their scan state. see #428
             if (tokensLine is TokensLine tl && tl.ScanState != null && (tokenType == TokenType.PartialCobolWord || 
                     tl.ScanState.InsideDataDivision && MultilineScanState.IsScanStateDependent(this)))
-                ScanStateSnapshot = tl.ScanState.Clone();
+                scanStateSnapshot = tl.ScanState.Clone();
+
         }
 
         /// <summary>
@@ -302,7 +303,7 @@ namespace TypeCobol.Compiler.Scanner
         /// </summary>
         public LiteralTokenValue LiteralValue { get; set; }
 
-        private MultilineScanState scanStateSnapshot;
+        private readonly MultilineScanState scanStateSnapshot;
         /// <summary>
         /// ScanState associated to this token if any, null otherwise.
         /// This property is used to allow PartialCobolWords proper reconstruction.
@@ -312,10 +313,6 @@ namespace TypeCobol.Compiler.Scanner
             get
             {
                 return scanStateSnapshot ?? (this.TokensLine is TokensLine tl ? tl.ScanState : null);
-            }
-            private set
-            {
-                scanStateSnapshot = value;
             }
         }
 
--- a/TypeCobol/Compiler/Scanner/Token.cs
+++ b/TypeCobol/Compiler/Scanner/Token.cs
@@ -76,9 +76,9 @@ namespace TypeCobol.Compiler.Scanner
             UsesVirtualSpaceAtEndOfLine = usesVirtualSpaceAtEndOfLine;
 
             //Scan Dependent Tokens Inside DataDivision must have their scan state. see #428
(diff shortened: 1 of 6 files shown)


fm-117:
Ok I don't fully understand the fix but it looks like we were previously lacking the `ScanStateSnapshot` for some tokens and then the replace operation would produce inconsistent tokens because of that.
@mayanje Correct me if I'm wrong but the fix consists in making a snapshot of the ScanState when required at construction time which is good. Here are my suggestions for improvement :
- Make `ScanStateSnapshot` property readonly and initialize it at construction time
  - for tokens inside data division that are scan state dependent
  - for PartialCobolWords
- According to current usages, the `ScanStateSnapshot` does not need to be modified after the token has been instanciated (the current usages of the setter are close to the constructor call so it is possible to remove them and let the constructor do the job)

Some additional remarks :
- `FigureOutScanState` logic could be moved into `ScanStateSnapshot` property getter as it uses only data from token
- `ScanState` property of `TokensLine` could be directly defined in `ITokensLine` interface as all implementations are in fact children of `TokensLine` and thus have the property. This would remove the need to type-check before accessing the ScanState
`

mayanje:
I agree all the remarks, but concerning ITokensLine and adding a getter ScanState, this is a litle bit an another issue, because here ScanStateSnapshot is before all a pseudo Token specific property, I mean  ITokensLine is a general interface all ITokensLine implementations may not need a ScanState and ScanState is specific to TokensLine, otherwise we can simply say that Token's tokensLine property a TokensLine instance.
ok CorrectType is for another PR.

rooksdo (dismissed):

TypeCobol/Compiler/Scanner/Token.cs
```diff
@@ -297,11 +302,22 @@ public virtual string Text
...
         /// This property is used to allow PartialCobolWords proper reconstruction.
         /// </summary>
-        public MultilineScanState ScanStateSnapshot { get; set; }
+        public MultilineScanState ScanStateSnapshot
+        {
+            get
+            {
+                return scanStateSnapshot ?? (this.TokensLine is TokensLine tl ? tl.ScanState : null);
+            }
+            private set
```
fm-117, line 316: You can remove this setter, make the field `scanStateSnapshot` readonly and set the field value (instead of calling the setter) in constructor.

fm-117 (dismissed):

TypeCobol/Compiler/Scanner/Token.cs
mayanje: So this means that this value must be always initialized to (this.TokensLine is TokensLine tl ? tl.ScanState : null); by default.
ok

fm-117: No this does not change the initial value. Keep the property getter and replace the setter call with field assignment. The initial value and logic stays the same.

mayanje: ok done in 5e55727

TypeCobol/Compiler/Scanner/Token.cs
```diff
@@ -74,8 +74,14 @@ internal Token(TokenType tokenType, int startIndex, int stopIndex, bool usesVirt
             HasClosingDelimiter = false;
 
             UsesVirtualSpaceAtEndOfLine = usesVirtualSpaceAtEndOfLine;
-        }  
-      
+
+            //Scan Dependent Tokens Inside DataDivision must have their scan state. see #428
+            if (tokensLine is TokensLine tl && tl.ScanState != null && (tokenType == TokenType.PartialCobolWord ||
```
smedilol, line 79: ScanState could be exposed in ITokensLine to avoid this cast.

smedilol (changes requested):

TypeCobol/Compiler/Scanner/Token.cs
mayanje: ok done in 
64eb1a1

smedilol (approved):

fm-117 (approved):

rooksdo (approved):

PR #1861, opened 2021-01-15
1818 create cfgdfainfo lsp notification

### LINKED ISSUE #1818: TC LSP create cfg_dfa_info custom notification.

mayanje:
**What have been done:**

- Code clean up regarding -ol (Outlineview) option
- Add a new server notification to the client for CFG/DFA information
- Cfg data builder to be sent by the notification
- Dynamic update of -ol and -cfg option from the client when it's configuration changes.
- Merge from develop

**Hum...**
A Next commit is coming with a LSR test.

Resolves #1818.


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -53,8 +53,16 @@ public class Workspace
...
+        /// <summary>
+        /// Event When the client configuration has changed in reception of a DidChangeConfiguration notification
+        /// </summary>
+        public EventHandler<IEnumerable<string>> ClientConfigurationChangedEvent { get; set; }
         public Queue<MessageActionWrapper> MessagesActionsQueue { get; private set; }
         private Func<string, Uri, bool> _Logger;
+        /// <summary>
+        /// The Analyzer Provider.
+        /// </summary>
+        public AnalyzerProvider AnalyzerProvider { get; private set; }
```
fm-117, line 65: No need for a new property, you already have access to the `IAnalyzerProvider` of the current `CompilationProject`.

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -153,14 +157,16 @@ public Workspace(string rootDirectoryFullName, string workspaceName, Queue<Messa
...
+            this.AnalyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer(TypeCobolLanguageServer.lspcfgId, CfgBuildingMode.Standard));
 
             this._rootDirectoryFullName = rootDirectoryFullName;
             this._workspaceName = workspaceName;
 
             var defaultDocumentFormat = new DocumentFormat(Encoding.GetEncoding("iso-8859-1"), EndOfLineDelimiter.CrLfCharacters, 80, ColumnsLayout.CobolReferenceFormat);
             this.CompilationProject = new CompilationProject(
                 _workspaceName, _rootDirectoryFullName, Helpers.DEFAULT_EXTENSIONS, defaultDocumentFormat,
-                new TypeCobolOptions(), null); //Initialize a default CompilationProject - has to be recreated after ConfigurationChange Notification
+                new TypeCobolOptions(), this.AnalyzerProvider); //Initialize a default CompilationProject - has to be recreated after ConfigurationChange Notification
```
fm-117, line 169: I thought the initial `CompilationProject` would not require any analyzerProvider because it will be recreated anyway when we get the client configuration (including the cfg/dfa options). So better keep it to null initially.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
```diff
@@ -278,6 +282,42 @@ public void WholeDocumentChanged(object sender, EventArgs eventArgs)
             UpdateTokensLines(compilationDocument);
         }
 
+        /// <summary>
+        /// LSR Cfg Analyzer ID.
+        /// </summary>
+        public const string lspcfgId = "lsp-cfg";
```
fm-117, line 288: `public const` is evil because of compile-time copy of the constant. Use `internal const` instead.

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -506,12 +512,15 @@ public void DidChangeConfigurationParams(IEnumerable<string> arguments)
 
             var typeCobolOptions = new TypeCobolOptions(Configuration);
 
+            if (ClientConfigurationChangedEvent != null)
+                ClientConfigurationChangedEvent(this, arguments);
+
             //Configure CFG/DFA analyzer + external analyzers if any
             var analyzerProvider = new CompositeAnalyzerProvider();
-            analyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer("cfg-dfa", Configuration.CfgBuildingMode));
+            this.AnalyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer(TypeCobolLanguageServer.lspcfgId, CfgBuildingMode.Standard));
```
fm-117, line 520: There is a problem here, the `Configuration.CfgBuildingMode` is lost so this means that we currently do not support use of DFA features along with CFG graph view. Also the `UseCfgDfaRefresh` option is not used here but it should.

We need to think about the configuration and the way graphs are used. If we keep our current CFG/DFA implementation, we'll have to create 2 analyzers if the user want to visualize graphs and perform DFA analysis and this is decided here when we register activators.

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
mayanje: to qualify it as evil it's very severe, but it's a lesson for me, ok for internal.

CLI/test/CLI.Test.csproj
```diff
@@ -60,7 +60,7 @@
     <None Include="ressources\**">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
-  </ItemGroup>
+  </ItemGroup>
```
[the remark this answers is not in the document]
mayanje: Some merge effect are really strange

TypeCobol.Analysis.Test/DfaBuildUseAndDefListTest.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System.IO;
+using System.IO;
```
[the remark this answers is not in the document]
mayanje: Same remark as above

TypeCobol.Analysis/Graph/CfgDotFileForNodeGenerator.cs
```diff
@@ -50,6 +50,10 @@ private static string Escape(string text)
         private readonly StringBuilder _blocksBuffer;
         private readonly StringBuilder _edgesBuffer;
         private ControlFlowGraph<Node, D> _cfg;
+        private int ClusterIndex = -1;
+
+        public delegate void BlockEmitted(BasicBlock<Node, D> block, int clusterIndex);
+        public event BlockEmitted BlockEmittedEvent;
```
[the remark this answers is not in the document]
mayanje: It's ok like that....

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/Cfg Dfa Notification/CfgDfaParams.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        /// <summary>
+        /// Empty Constructor
+        /// </summary>
+        /// <param name="textDocument">The Target document identifier</param>
+        public CfgDfaParams(VsCodeProtocol.TextDocumentIdentifier textDocument)
+            : this(textDocument, null, null, null)
+        {            
+        }
```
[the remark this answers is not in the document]
mayanje: Both constructors are used you can see in what cases

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/Cfg Dfa Notification/CfgDfaParamsBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+    public class CfgDfaParamsBuilder
+    {
+        private TextDocumentIdentifier textDocumentId;
+        private string dotFilePath;
+        private List<Position> instructionPositions;
+        private List<BasicBlockInfo> basicBlockInfos;
+        /// <summary>
+        /// Resulting CfgDfaParams instance.
+        /// </summary>
+        public CfgDfaParams Params => new CfgDfaParams(textDocumentId, dotFilePath, instructionPositions.ToArray(), basicBlockInfos.ToArray());
```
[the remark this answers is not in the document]
mayanje: Property does not mean immutable.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -23,29 +24,35 @@ public TypeCobolCustomLanguageServer(IRPCServer rpcServer, Queue<MessageActionWr
...
+            return result;
         }
 
-        protected override void OnDidOpenTextDocument(DidOpenTextDocumentParams parameters)
+        /// <summary>
+        /// Handle -ol and -cfg options from the client configuration change notification.
+        /// </summary>
+        /// <param name="sender">Should be the Workspace instance</param>
+        /// <param name="options">Client's Options</param>
+        private void OnClientOptionsChanged(object sender, IEnumerable<string> options)
```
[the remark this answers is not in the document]
mayanje: Yes it is for properties that come from the Client plugin an that are not TypeCobol.CLI properties, like -ol (Outline) these are TypeCobol Custom and TypeCobolCustomLanguageServer set some flags depending on it.

smedilol (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/Cfg Dfa Notification/CfgDfaParams.cs
```diff
@@ -0,0 +1,73 @@
...
+            /// </summary>
+            public int id { get; set; }
+            /// <summary>         
+            /// Any Sub graph id to which this block belongs to.
+            /// </summary>
+            public int subgraphId { get; set; }
+            /// <summary>
+            /// Index of the first instruction in the instruction range array
+            /// </summary>
+            public int fstInstIdx { get; set; }
```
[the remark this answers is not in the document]
mayanje: I thought so, but for gaining space in the JSON message, and after the suggestion of fm-117 to gain size here, I decided to reduce this name. This is a gain of 11 characaters. What do you thing about that?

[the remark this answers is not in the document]
smedilol: If it's a performance constraint, then it's ok.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/Cfg Dfa Notification/CfgDfaParamsBuilder.cs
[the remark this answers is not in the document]
smedilol: A property can works, but a method is better when you want to show that there is a side effect when you call it.
See guidelines:
https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN

TypeCobol.LanguageServer/Workspace.cs
fm-117: Also we need the `CompositeAnalyzerProvider` to add external quality analyzer.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
```diff
@@ -278,6 +282,48 @@ public void WholeDocumentChanged(object sender, EventArgs eventArgs)
...
+        {
+            CfgDfaParams result = null;
+            docContext.FileCompiler.CompilationResultsForProgram.TryGetAnalyzerResult(lspcfgId, out IList<ControlFlowGraph<Node, object>> cfgs);
+            if (cfgs != null)
+            {                
+                //Create a temporary dot file.
+                string tempFile = Path.GetTempFileName();
+                using (TextWriter writer = lsrMode ? (TextWriter)new StringWriter() : File.CreateText(tempFile))
+                {
+                    CfgDfaParamsBuilder builder = new CfgDfaParamsBuilder(new TextDocumentIdentifier(docContext.TextDocument.uri), lsrMode ? null : tempFile);
```
fm-117, line 306: Maybe we could keep a builder instance and add a `Clear` or `Reset` method to it instead of creating a new builder each time the notification is called.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -23,29 +24,46 @@ public TypeCobolCustomLanguageServer(IRPCServer rpcServer, Queue<MessageActionWr
...
-        {
-            if (UseOutlineRefresh && this.Workspace.DocumentModifiedEvent == null)
+        /// <summary>
+        /// Handle -ol and -cfg options from the client configuration change notification.
+        /// </summary>
+        /// <param name="sender">Should be the Workspace instance</param>
+        /// <param name="options">Client's Options</param>
+        private void OnClientOptionsChanged(object sender, IEnumerable<string> options)
+        {            
+            if (!this.InRobotLsrTestMode)
```
fm-117, line 55: Can you explain why in Robot mode we should not update the options ?

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/Cfg Dfa Notification/CfgDfaParamsBuilder.cs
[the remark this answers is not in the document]
mayanje: Ok it's true, but en fact this property shall be call only once the building is finished, that's why I make the resulting value as a property. There will be no need to call it multiple time.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -23,29 +24,35 @@ public TypeCobolCustomLanguageServer(IRPCServer rpcServer, Queue<MessageActionWr
...
+        /// <param name="sender">Should be the Workspace instance</param>
+        /// <param name="options">Client's Options</param>
+        private void OnClientOptionsChanged(object sender, IEnumerable<string> options)
         {
-            if (UseOutlineRefresh && this.Workspace.DocumentModifiedEvent == null)
-            {
-                this.Workspace.DocumentModifiedEvent += DocumentModified;
-            }
-            base.OnDidOpenTextDocument(parameters);
+            //this.UseOutlineRefresh = options.Contains("-ol");
```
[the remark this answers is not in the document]
mayanje: Ok it is uncommented in  commit "WI #1818 Add LSR CfgDfaParams test c8f5286 ", because otherwise all tests that used -ol were in failure, but now the flags is not taken in account if the server is excuted in LSR test mode.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -23,29 +24,35 @@ public TypeCobolCustomLanguageServer(IRPCServer rpcServer, Queue<MessageActionWr
...
-            base.OnDidCloseTextDocument(parameters);
-            if (UseOutlineRefresh && this.Workspace.DocumentModifiedEvent != null && this.Workspace.IsEmpty)
-                this.Workspace.DocumentModifiedEvent -= DocumentModified;
-
+            this.Workspace.DocumentModifiedEvent -= DocumentModified;
+            this.Workspace.ClientConfigurationChangedEvent -= OnClientOptionsChanged;
+            base.OnShutdown();
         }
 
         protected override void OnDidSaveTextDocument(DidSaveTextDocumentParams parameters)
```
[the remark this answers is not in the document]
mayanje: hum.. it is safe to do that if the server is shuting down then OnDidReceiveRefreshOutline and UpdateCfgDfaInformation willnot be called, this the behavior in this PR.

TypeCobol.LanguageServer/Workspace.cs
mayanje: Ok to remove it.

TypeCobol.LanguageServer/Workspace.cs
mayanje: Right, and you make me discover multi analyzer instantiation.

TypeCobol.LanguageServer/Workspace.cs
mayanje: In LSP mode for the moment only CFG Standard mode is used for the CFGViewer, but you are right if DFA analysis is used we need a second analyzer. I remenber we were thinking a bout a solution to compute the DFA (Extended) Graph from a standard graph, by applying a transformation.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
mayanje: Hum... At this time I don't know if we are thread safe here

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
fm-117: I don't get it, if the server is shuting down, there is no need to send Outline of CFG/DFA refresh.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: Because each LSR test are configured differently, and -ol and -cfg options come as command line options not as configuration changed options. LSR tests create a custom Initialize and Configuration Change notification. Maybe we should review this aspect.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
mayanje: Yes, because there can still be pending parsing in process, so refresh event can come afterwards. Do you understand?

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
fm-117: I don't think we need that. In `TypeCobolCustomLanguageServer` class, you can override `OnDidChangeConfiguration` method and do the same job. So no need for this event to get config data back from the Workspace.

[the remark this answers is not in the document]
mayanje: It is necessary because -ol and -cfg options also come from the command line, and are set via TypeCobolCustomLanguageServer when no Workspace is instantiate yet. So it is TypeCobolCustomLanguageServer which host UseOutlineRefresh  and UseCfgDfaDataRefresh.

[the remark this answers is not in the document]
mayanje: But you are right for -ol and -cfg options inside LSRTest, I am going to parameterized these options in the custom config.json generated.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: But you are right for -ol and -cfg options inside LSRTest, I am going to parameterized these options in the custom config.json generated.

fm-117 (changes requested):

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -506,12 +506,18 @@ public void DidChangeConfigurationParams(IEnumerable<string> arguments)
...
 
+            if (ClientConfigurationChangedEvent != null)
+                ClientConfigurationChangedEvent(this, arguments);
+
             //Configure CFG/DFA analyzer + external analyzers if any
-            var analyzerProvider = new CompositeAnalyzerProvider();
-            analyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer("cfg-dfa", Configuration.CfgBuildingMode));
-            analyzerProvider.AddCustomProviders(Configuration.CustomAnalyzerFiles);
+            var compositeAnalyzerProvider = new CompositeAnalyzerProvider();
+            if (Configuration.CfgBuildingMode != CfgBuildingMode.None)
```
fm-117, line 514: no need to test here, the `CreateCfgAnalyzer` method returns null if building mode is set to None and the analyzer provider discards null values.

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -23,29 +24,42 @@ public TypeCobolCustomLanguageServer(IRPCServer rpcServer, Queue<MessageActionWr
...
-            if (UseOutlineRefresh && this.Workspace.DocumentModifiedEvent != null)
-                this.Workspace.DocumentModifiedEvent -= DocumentModified;
+        /// <summary>
+        /// Are we using the CFG view in the client.    
+        /// </summary>
+        public bool UseCfgDfaDataRefresh { get; set; }
+        /// <summary>
+        /// True if The server is executed in LSR TestMode
+        /// </summary>
+        public bool InRobotLsrTestMode { get; set; }
```
fm-117, line 34: Le flag peut désormais être supprimé, on utilise que `InLsrClientMode`.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
fm-117: Ok so the scenario is that we already unsubscribed from DocumentModified event (because of shutdown) and we have a pending refresh. I agree that the outline and cfg data will not be sent to client but
- why should the data be transmitted if we are terminating the server ?
- how is it related to OnDidSaveTextDocument ?

My idea is that `TypeCobolCustomLanguageServer` does not need to send any data about outline/cfg on a save operation because it is handled asynchronously with the refresh and consequently this override should be deleted.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
fm-117: You're right

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
```diff
@@ -278,6 +282,48 @@ public void WholeDocumentChanged(object sender, EventArgs eventArgs)
...
+        /// </summary>
+        internal const string lspcfgId = "lsp-cfg";
+
+        /// <summary>
+        /// Method to update CFG/DFA information.
+        /// </summary>
+        /// <param name="fileCompiler">The underlying File Compiler</param>
+        /// <param name="lsrMode)">True if this action is executed in a LSR Mode</param>
+        /// <returns>CFG/DFA Data information</returns>
+        public CfgDfaParams UpdateCfgDfaInformation(DocumentContext docContext, bool lsrMode)
```
fm-117, line 296: About the name `lsrMode` for the boolean parameter, I think it is more something like `writeToFile` because the purpose of this method is to turn CFG/DFA results into a `CfgDfaParams` object. The method is not concerned by the mode (LSR or regular) but rather accepts an option on how to create its result. The parameter should be named from method perspective not caller perspective.

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
smedilol: I don't remember the whole mechanism, but the save action is a way to resynchronize all data properly because sometimes incremental mode is out of sync.

Anyway, as the method `OnDidSaveTextDocument` is not in the scope of this PR, I suggest to discuss about it in another issue.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
smedilol: +1

smedilol (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
fm-117: I thought this would be included in this PR because @mayanje reviewed the outline mechanism along with the new cfg/dfa notification.

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -256,17 +256,22 @@ protected override void OnShutdown()
...
-                IEnumerable<string> arguments = array.Select(t => t.ToString());
-                this.Workspace.DidChangeConfigurationParams(arguments);
+                IEnumerable<string> argsEnum = array.Select(t => t.ToString());
+                string[] arguments = argsEnum.ToArray<string>();
+                OnDidChangeConfiguration(arguments);
             }
             else
             {
-                this.Workspace.DidChangeConfigurationParams(parameters.settings.ToString());
+                OnDidChangeConfiguration(parameters.settings.ToString().Split(' '));
```
fm-117, line 274: Unwanted whitespace at the end of the line

fm-117 (changes requested):

TypeCobol.Analysis.Test/CfgDfaBuildingTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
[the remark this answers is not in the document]
mayanje: From the commit history it seems that no.

TypeCobol.Analysis.Test/DfaBuildUseAndDefListTest.cs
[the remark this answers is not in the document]
mayanje: From the commit history it seems that no.

TypeCobol.Analysis.Test/CfgDfaBuildingTests.cs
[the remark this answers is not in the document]
fm-117: On develop the file is UTF8 with BOM, on your branch it is Win-1252. Although I don't know where the change come from, you can switch back to previous version file by re-encoding the file to UTF8 with BOM (using Notepad++ for example)

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: Now both can be used.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
mayanje: The main thing is that DocumentModified event comes from another thread.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
fm-117: Originally, we only had `TypeCobolServerHost.LsrMode` which indicate whether we are using the robot or not and it was limited to `TypeCobolServerHost` class. Now we have to know that in `TypeCobolCustomLanguageServer` to select the correct output type for CFG data. So why not just add a new property `LsrMode` in `TypeCobolCustomLanguageServer` and initialize it with the value from `TypeCobolServerHost.LsrMode` ?

In my opinion, we don't need a new option `lsrclient`, we just need to let the server know that the robot is used and we already have the option for that.

mayanje: My point of view is that it can be  interresting to know if the server is called from for instance CobolEditEI or it is called from the Test execution mode. LsrMode was and is the InRobotLsrTestMode option, and was only set  by test execution, not from the client point.

TypeCobol.LanguageServer/TypeCobolServer.cs
mayanje: Done in d7cc3ff

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
mayanje: Done in d7cc3ff

TypeCobol.Analysis.Test/DfaBuildUseAndDefListTest.cs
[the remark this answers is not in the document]
mayanje: Done in d7cc3ff

TypeCobol.Analysis.Test/CfgDfaBuildingTests.cs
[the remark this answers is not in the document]
mayanje: Done in d7cc3ff

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
```diff
@@ -278,6 +282,49 @@ public void WholeDocumentChanged(object sender, EventArgs eventArgs)
...
+                {
+                    CfgDfaParamsBuilder builder = new CfgDfaParamsBuilder(new TextDocumentIdentifier(docContext.TextDocument.uri), writeToFile ? tempFile : null);
+                    CfgDotFileForNodeGenerator<object> gen = new CfgDotFileForNodeGenerator<object>(cfgs[0]);
+                    gen.FullInstruction = true;
+                    gen.BlockEmittedEvent += (block, subgraph) => builder.AddBlock<object>(block, subgraph);
+                    gen.Report(writer);
+                    result = builder.GetParams();
+                    if (!writeToFile)
+                    {
+                        writer.Flush();
```
fm-117, line 315: `writer.Flush()` is not required. `StringWriter` has empty implementation for it.

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -256,17 +256,22 @@ protected override void OnShutdown()
             base.OnShutdown();
         }
 
+        protected virtual void OnDidChangeConfiguration(string[] arguments)
+        {
+            this.Workspace.DidChangeConfigurationParams(arguments);
+        }
         protected override void OnDidChangeConfiguration(DidChangeConfigurationParams parameters)
         {
             if (parameters.settings is Newtonsoft.Json.Linq.JArray)
```
fm-117, line 265: you can merge type check and cast here

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
fm-117: So if the client uses the `-r` option, we would know that we are in robot test mode but we could not distinguish between unit test and real client. Is that a problem ? I mean the code should never know that it is running in unit test context.

smedilol: I don't understand either why we need 2 options.
`InRobotLsrTestMode` and `InLsrClientMode` are exactly in the same way in TypeCobolCustomLanguageServer#UpdateCfgDfaInformation.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -189,22 +203,46 @@ protected virtual void OnDidReceiveSignatureHelpContext(SignatureHelpContextPara
...
+        /// </summary>
+        /// <param name="uri">Uri of the document to update Cfg/Dfa Informations</param>
+        private void UpdateCfgDfaInformation(string uri)
+        {
+            if (this.UseCfgDfaDataRefresh)
+            {
+                var context = GetDocumentContextFromStringUri(uri, Workspace.SyntaxTreeRefreshLevel.NoRefresh);
+                if (context != null && context.FileCompiler != null)
+                {
+                    var cfgDfaParams = context.LanguageServer.UpdateCfgDfaInformation(context, !(InRobotLsrTestMode || InLsrClientMode));
```
smedilol, line 232: It'd better to have a dedicated option to de/activate the dotFile.
Because a client without a mean to visualize the dot file don't care about it.

```suggestion
                    var cfgDfaParams = context.LanguageServer.UpdateCfgDfaInformation(context, writeCfgDfaDotFile);
```

smedilol (changes requested):

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
smedilol: As writer is declared as a `TextWriter` I'd rather keep `writer.Flush();` just in case the implementation change.

fm-117: In fact, my justification was wrong. The next line `result.dotContent = writer.ToString();` assumes that the writer is a `StringWriter` because only `StringWriter` redefines its `ToString()` method to return all characters accumulated so far. So the `writer.Flush();` call is unnecessary because `ToString` will flush all data.

mayanje: It ssems to  that this Flush was necessay tha't s why i Put it.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: This hack is only for LSR Test mode,  and these are options set thru LSR execution mode, it is not a Server option. An this is used for CFGViewer to have a file ready to read. This is a caller's choice.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
mayanje: I think you are right, it seems that internal representation of StringWriter uses a StringBuilder.
https://referencesource.microsoft.com/#mscorlib/system/io/stringwriter.cs

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
fm-117: I agree with @smedilol, the choice here for the client is how the result should be returned (as a separate file or inside JSON response). It is not really related to the use of the robot because we can imagine scenarios where we are monitoring messages between client and server and the client would still use file mode.

mayanje: So what you say is that there is an option in the client to select file or content, and the option shall be applied to LSR mode also, So this is another issue.

fm-117: No I say we forget about LSR mode because it is of no concern here :
- we need a new option `-cfg` to let the client choose whether it wants to have the CFG/DFA notifications
  - this option let also the client choose the nature of results, for example `-cfg AsFile` for temporary dot file and `-cfg AsContent` for inline JSON
- unit tests are not compatible with `-cfg AsFile` because the file path is unpredictable, they shall use `-cfg AsContent`

And no need to change anything about `-r` option and its related properties.

mayanje: Ok, let's create a new issue for this new option, and tests that come with it.
LSR client execution mode will use -cfg AsContent.

smedilol: It's not a new issue, it's a remark about a change in this PR, so it must be done in this PR.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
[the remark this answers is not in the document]
fm-117: New issue #1882 created.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: Ok, it's not a new issue, but it's a new option.
For TypeCobol.CLI Option or for the Language Server Only?,
What is the default value  AsFile or AsContent ?
Why it should not be possible to pass an output file for the AsFile options? because even tests can create their own output file and compare the result as it is done in other tests that generates a file.

fm-117: This is for Language Server only it is an extension of the new `-cfg` option that you just introduced in this PR.

I'm not sure about the default value but you can also make it mandatory if it's easier.

Passing the output file path cannot be done (easily) as an option because there are multiple documents opened at once therefore we need multiple paths and some kind of mapping between source files and dot files. It can be done with a request though. Instead of saying 'I support CFG/DFA notifications', the client would request CFG/DFA data directly and could specify in the request the file in which it expects the result.

mayanje: Ok in this case no output file as option.

TypeCobol.LanguageServer/TypeCobolServerHost.cs
```diff
@@ -210,6 +215,9 @@ static int Main(string[] args)
                 { "dcs|disablecopysuffixing", "Deactictivate Euro Information suffixing", v => UseEuroInformationLegacyReplacingSyntax = false },
                 { "sc|syntaxcolor",  "Syntax Coloring Support.", _ => UseSyntaxColoring = true},
                 { "ol|outlineRefresh",  "Outline Support.", _ => UseOutlineRefresh = true},
+                { "cfg=",  "{dot output mode} Control Flow Graph support and Dot Output mode: No/0, AsFile/1 or AsContent/2.",
+                    (String m) => {TypeCobolCustomLanguageServer.UseCfgMode ucm = TypeCobolCustomLanguageServer.UseCfgMode.No;
+                        Enum.TryParse(m, false, out ucm); UseCfg = ucm; }  },
```
fm-117, line 220: no need to specify the `ignoreCase` to `false`, it is the default behavior.

fm-117 (approved):

TypeCobol.LanguageServer/TypeCobolServerHost.cs
mayanje: Done in 
c771e78

fm-117 (approved):

rooksdo (approved):

smedilol (approved):

PR #1862, opened 2021-01-21
WI #1845 Add .xmldiag format for error output in CLI

### LINKED ISSUE #1845: New diagnostics output format for CLI
Currently 2 formats are supported by CLI :
-  `txt` is the default format, it is designed for humans and describes diagnostics in plain strings
-  `xml` is an alternate format designed for RdZ (as a preprocessor format)

Those 2 formats do not contain all information as opposed to LSP format so the goal of this issue  is to create a third format for CLI.

Here is a proposal, in XML:

```xml
<build return-code="ParsingDiagnostics">
  <filetable>
    <file index="1" path="C:\Users\UserName\Documents\cobol\test.tcbl" />
  </filetable>
  <diagnostics>
    <diagnostic file-index="1">
      <range line-start="4" column-start="20" line-end="4" column-end="23" />
      <message>Semantic error: Symbol var1 is not referenced</message>
      <info severity="Error" code="30" source="" />
    </diagnostic>
  </diagnostics>
</build>
```

- it is a mix between current xml format and data from LSP diagnostic model
- each diagnostic is associated to a file using a `file-index` attribute to support multifile parsing scenarios
- all fields from LSP Diagnostic class are serialized
- for quality rules violations, code is set to `47` and source will contain the rule identifier

fm-117:
Fixes #1845.

### DIFF
--- a/CLI/test/ressources/mass_generation_xmldiagformat/output_expected/error.xmldiag
+++ b/CLI/test/ressources/mass_generation_xmldiagformat/output_expected/error.xmldiag
@@ -1,0 +1,35 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<build return-code="ParsingDiagnostics">
+  <filetable>
+    <file index="1" path="input\DVZZPGM1.tcbl" />
+    <file index="2" path="input\DVZZPGM2.tcbl" />
+    <file index="3" path="input\DVZZPGM3.tcbl" />
+  </filetable>
+  <diagnostics>
+    <diagnostic file-index="1">
+      <range line-start="5" column-start="8" line-end="5" column-end="43" />
+      <message>Semantic error: TYPE 'DVZZPGM2.Type1' is not referenced</message>
+      <info severity="Error" code="30" source="Page" />
+    </diagnostic>
+    <diagnostic file-index="1">
+      <range line-start="7" column-start="33" line-end="7" column-end="36" />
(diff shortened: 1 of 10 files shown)


rooksdo (approved):

smedilol (approved):

CLI/test/ressources/mass_generation_xmldiagformat/output_expected/error.xmldiag
```diff
@@ -0,0 +1,35 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<build return-code="ParsingDiagnostics">
+  <filetable>
+    <file index="1" path="input\DVZZPGM1.tcbl" />
```
mayanje, line 4: Why paths are relative paths and not absolute paths?

fm-117: Good question, to build the filetable I'm using the `Inputs` dictionary from the `AbstractErrorWriter` so it looks like we store relative or absolute paths depending on the original input format passed to CLI.
@smedilol should we change that ?

mayanje: Ok, this also to make unit tests not absolute path dependent.

mayanje (approved):

Issue #1864, opened 2021-01-22
Add performance tests in v1.5 to use CFG/DFA options


mayanje:
**Purpose**
Is to add in TypeCobol.Analysis.Test the same performence tests than in TypeCobol.Test using test specialization, so that they can use and AnalyzerProvider instance for CFG/DFA when instantiating a Parser.

PR #1867, opened 2021-01-22
WI #1864 Add peformance test with CFG/DFA

mayanje:
**What have been done:**
Add two classes in TypeCobol.Analysis.Test

- PerformanceWithCfg
- PerformanceWithCfgDfa

Both inherit from class TypeCobol.Test.Performance

### DIFF
--- a/TypeCobol.Analysis.Test/PerformanceWithCfg.cs
+++ b/TypeCobol.Analysis.Test/PerformanceWithCfg.cs
@@ -20,7 +20,7 @@ namespace TypeCobol.Analysis.Test
         /// AnalyzerProvider for test.
         /// </summary>
         /// <returns></returns>
-        private static CompositeAnalyzerProvider MyAnalyzerProvider()
+        protected override CompositeAnalyzerProvider CreateAnalyzerProvider()
         {
             //Add analyzers
             var analyzerProvider = new CompositeAnalyzerProvider();
@@ -28,27 +28,6 @@ namespace TypeCobol.Analysis.Test
             const string cfgDfaId = "cfg-dfa";
             analyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer(cfgDfaId, CfgBuildingMode.Standard));
             return analyzerProvider;
-        }
-
-        /// <summary>
-        /// Method for parsing a document.
-        /// </summary>
-        /// <param name="fullPath"></param>
-        /// <param name="options"></param>
-        /// <param name="format"></param>
-        /// <param name="copiesFolder"></param>
-        /// <returns></returns>
-        protected override TypeCobol.Parser ParseDocument(string fullPath, TypeCobolOptions options, TypeCobol.Compiler.DocumentFormat format, string[] copiesFolder)
-        {
-            var document = new TypeCobol.Parser();
-            document.Init(fullPath, options, format, copiesFolder, MyAnalyzerProvider());
-            document.Parse(fullPath);
-            return document;
-        }
-
-        protected override CompilationProject CreateCompilationProject(string projectName, string rootDirectory, string[] fileExtensions, DocumentFormat documentFormat, TypeCobolOptions compilationOptions, IAnalyzerProvider analyzerProvider)
-        {
-            return base.CreateCompilationProject(projectName, rootDirectory, fileExtensions, documentFormat, compilationOptions, MyAnalyzerProvider());
         }
 
         [TestMethod]
--- a/TypeCobol.Analysis.Test/PerformanceWithCfg.cs
+++ b/TypeCobol.Analysis.Test/PerformanceWithCfg.cs
@@ -1,0 +1,212 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TypeCobol.Compiler;
+using TypeCobol.Compiler.Diagnostics;
+using TypeCobol.Compiler.Directives;
+using TypeCobol.Compiler.Text;
+using TypeCobol.Test.Compiler.Parser;
+using TypeCobol.Test.Utils;
+
+namespace TypeCobol.Analysis.Test
+{
+    [TestClass]
+    [Ignore]
+    public class PerformanceWithCfg : TypeCobol.Test.Parser.Performance.Performance
+    {
+        /// <summary>
+        /// Method for parsinga document.
+        /// </summary>
+        /// <param name="fullPath"></param>
+        /// <param name="options"></param>
+        /// <param name="format"></param>
+        /// <param name="copiesFolder"></param>
+        /// <returns></returns>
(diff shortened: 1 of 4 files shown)


TypeCobol.Analysis.Test/PerformanceWithCfg.cs
```diff
@@ -0,0 +1,212 @@
...
+            document.Init(fullPath, options, format, copiesFolder, analyzerProvider);
+            document.Parse(fullPath);
+            return document;
+        }
+
+        [TestMethod]
+        [TestCategory("Performance")]
+        [TestProperty("Time", "fast")]
+        //[Ignore]
+        public new void AntlrPerformanceProfiler()
```
smedilol, line 45: You can remove AntlrPerformanceProfiler test, because CodeElement/Antlr parsing is not affected by CFG.

TypeCobol.Analysis.Test/PerformanceWithCfgDfa.cs
```diff
@@ -0,0 +1,212 @@
...
+            document.Init(fullPath, options, format, copiesFolder, analyzerProvider);
+            document.Parse(fullPath);
+            return document;
+        }
+
+        [TestMethod]
+        [TestCategory("Performance")]
+        [TestProperty("Time", "fast")]
+        //[Ignore]
+        public new void AntlrPerformanceProfiler()
```
smedilol, line 45: Same than CFG, you can remove AntlrPerformanceProfiler test.

smedilol (changes requested):

TypeCobol.Analysis.Test/PerformanceWithCfgDfa.cs
mayanje: done in af9e8cb

TypeCobol.Analysis.Test/PerformanceWithCfg.cs
mayanje: done in af9e8cb

TypeCobol.Test/Parser/Performance/Performance.cs
```diff
@@ -368,8 +368,21 @@ public void Part3_FullParsing_TC_DeepTypes()
...
-
+        /// <summary>
+        /// Method for parsinga document.
+        /// </summary>
+        /// <param name="fullPath"></param>
+        /// <param name="options"></param>
+        /// <param name="format"></param>
+        /// <param name="copiesFolder"></param>
+        /// <returns></returns>
+        protected virtual TypeCobol.Parser parseDocument(string fullPath, TypeCobolOptions options, TypeCobol.Compiler.DocumentFormat format, string[] copiesFolder)
```
fm-117, line 379: `ParseDocument` to match PascalCase convention.

About the pattern used, you can improve it and avoid these methods marked as `new` and calling base code. You need 4 classes :
- Performance : abstract base class, contains all the tests and an abstract ParseDocument method to initialize the parser
- 3 concrete classes, 1 per mode
  - NoCfg
  - Cfg
  - Dfa

Each concrete class would implement the ParseDocument method to match its own needs and you can override the Antlr test in Cfg and Dfa to make it empty.

fm-117 (changes requested):

TypeCobol.Test/Parser/Performance/Performance.cs
fm-117: Capture
As you can see, `TestMethod1`  is defined once but translates into multiple tests 👍

mayanje: Sorry for PascalCase convention it's the inverse in Java, I will do the change. Also I'll try your pattern because it's true if it work like that it's fine.

mayanje: At first time I did what to suggest, the Performance base class was not abstract, it didn't work, so maybe with the abstract keyword it will work

mayanje: This method seems not working if tests classes are in different assemblies.

fm-117 (dismissed):
When we'll upgrade the test framework, we'll be able to remove the test duplication, but as for now we have to keep the duplicated `new` methods.

TypeCobol.Test/Parser/Performance/Performance.cs
```diff
@@ -368,8 +368,21 @@ public void Part3_FullParsing_TC_DeepTypes()
...
-
+        /// <summary>
+        /// Method for parsinga document.
+        /// </summary>
+        /// <param name="fullPath"></param>
+        /// <param name="options"></param>
+        /// <param name="format"></param>
+        /// <param name="copiesFolder"></param>
+        /// <returns></returns>
+        protected virtual TypeCobol.Parser ParseDocument(string fullPath, TypeCobolOptions options, TypeCobol.Compiler.DocumentFormat format, string[] copiesFolder)
```
smedilol, line 379: This method seems only for FullParsing. Does it also include Incremental ?

rooksdo (dismissed):

TypeCobol.Test/Parser/Performance/Performance.cs
mayanje: You are right, done in 3b1d62e

smedilol (approved):

TypeCobol.Analysis.Test/PerformanceWithCfg.cs
```diff
@@ -0,0 +1,215 @@
...
+            var document = new TypeCobol.Parser();
+            document.Init(fullPath, options, format, copiesFolder, MyAnalyzerProvider());
+            document.Parse(fullPath);
+            return document;
+        }
+
+        protected override CompilationProject CreateCompilationProject(string projectName, string rootDirectory, string[] fileExtensions, DocumentFormat documentFormat, TypeCobolOptions compilationOptions, IAnalyzerProvider analyzerProvider)
+        {
+            return base.CreateCompilationProject(projectName, rootDirectory, fileExtensions, documentFormat, compilationOptions, MyAnalyzerProvider());
+        }
```
fm-117, line 52: Ok so it looks like `ParseDocument` and `CreateCompilationProject` are respectively used for FullParsing and Incremental. You can improve factorization by removing these 2 methods (as they only differ at the type of analyzer used in derived classes) and replace them with something like :
```csharp
protected abstract void AddAnalyzers(CompositeAnalyzerProvider analyzerProvider);
```
in the base class `Performance`. No need to change anything from original code except to create a `CompositeAnalyzerProvider`, call `AddAnalyzer` and pass the instance either on the `Parser.Init` or into the `CompilationProject`.

fm-117 (changes requested):

rooksdo (approved):

fm-117 (approved):

TypeCobol.Analysis.Test/PerformanceWithCfg.cs
mayanje: Done in 09d4586

smedilol (approved):

rooksdo (approved):

PR #1871, opened 2021-01-26
WI #1869 Remove IASTAnalyzer, use QualityAnalyzer concept instead

### LINKED ISSUE #1869: Add mechanism for external analyzers to check compiler-directing statements
### Description

Here are the following compiler-directing statements that we would like to check by an external analyzer :

-     ENTER, READY or RESET TRACE, and SERVICE RELOAD
-     Extended source library control (BASIS, DELETE, and INSERT statements)
-     Source text manipulation (COPY and REPLACE statements)
-     Exception handling and label handling (USE statement)
-     Controlling compiler listings (*CONTROL, *CBL, EJECT, TITLE, SKIP1, SKIP2, and SKIP3 statements)
-     Specifying compiler options (CBL and PROCESS statements)
-     Specifying COBOL exception handling procedures (USE statements)

@Todo cobol samples

fm-117:
Fixes #1869.


TypeCobol/Compiler/Parser/ProgramClassDocument.cs
```diff
@@ -34,7 +34,7 @@ public ProgramClassDocument(TemporarySemanticDocument previousStepSnapshot, int
         /// <summary>
         /// Snapshot of the code elements document which was used to compute the current step
         /// </summary>
-        public IDocumentSnapshot<ICodeElementsLine> PreviousStepSnapshot { get; private set; }
+        public TemporarySemanticDocument PreviousStepSnapshot { get; private set; }
```
smedilol, line 37: I think other `PreviousStepSnapshot` properties can in TemporarySemanticDocument , CodeElementsDocument, ... can use the most derived type to avoid cast in CompilationUnit.

smedilol (changes requested):

TypeCobol/Compiler/Parser/ProgramClassDocument.cs
fm-117: With current code no, because this property is defined by `INextStepDocumentSnapshot<T>` as an `IDocumentSnapshot<T>`. C# does not allow (yet) override of inherited method return types. As for `ProgramClassDocument`, it does not implement `INextStepDocumentSnapshot` so I was ablet to change property type without impact.

smedilol (approved):

TypeCobol/Compiler/Parser/ProgramClassDocument.cs
smedilol: Ok.

rooksdo (approved):

mayanje (approved):

delevoye (approved):

Issue #1853, opened 2021-01-12
Program.FileConnectors and FileDescriptions are never set


smedilol:
For this sample:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.  Pgm.
       ENVIRONMENT DIVISION.
       INPUT-OUTPUT SECTION.
       FILE-CONTROL.
      *    SEQUENTIAL with FILE STATUS
           SELECT CUSTOMER-F01 ASSIGN TO UT-CUSTOMER-F01
           FILE STATUS IS FS-CUSTOMER.
		   
      *    SEQUENTIAL without FILE STATUS
           SELECT CUSTOMER-F02 ASSIGN TO UT-CUSTOMER-F02.
		   
      *    INDEXED with FILE STATUS
           SELECT STUDENT-F01  ASSIGN TO UT-STUDENT-F01
           ORGANIZATION IS INDEXED
           ACCESS IS RANDOM
           RECORD KEY IS STUDENT-F01-ID
           FILE STATUS IS FS-STUDENT.
		   
      *    INDEXED without FILE STATUS
           SELECT STUDENT-F02  ASSIGN TO UT-STUDENT-F02
           ORGANIZATION IS INDEXED
           ACCESS IS RANDOM
           RECORD KEY IS STUDENT-F02-ID.
		   
       DATA DIVISION.
       FILE SECTION.
       FD  CUSTOMER-F01.
       01  CUSTOMER-F01-FILE.
         05 CUSTOMER-F01-ID  PIC X(8).
         05 FILLER           PIC X(100).
       FD  CUSTOMER-F02.
       01  CUSTOMER-F02-FILE.
         05 CUSTOMER-F02-ID  PIC X(8).
         05 FILLER           PIC X(100).
       FD STUDENT-F01.
       01 STUDENT-F01-FILE.
         05 STUDENT-F01-ID   PIC X(5).
         05 FILLER           PIC X(25).
       FD STUDENT-F02.
       01 STUDENT-F02-FILE.
         05 STUDENT-F02-ID   PIC X(5).
         05 FILLER           PIC X(25).
       WORKING-STORAGE SECTION.
       01 WS-CUSTOMER-F01-FILE         PIC X(128).
       01 WS-CUSTOMER-F02-FILE         PIC X(128).
       01 FS-CUSTOMER                  PIC X(2).
       01 WS-STUDENT-F01-FILE          PIC X(30).
[...]

PR #1872, opened 2021-01-27
WI #1853 Set FileConnectors in Program

delevoye:
About adding additional properties mentionned in issue #1853:

**IOControlEnties** could not be set because they were not found in ProgramClassBuilder.

**SourceComputerProperties, ObjectComputerProperties, SpecialNamesDefinitions, RepositoryOfClassNames** were not set because they seemed to reveal the following problem:
They must be declared in a `CONFIGURATION SECTION` which is only valid in root programs. The IBM compiler forbids to have this section in nested programs, whereas this parser still allows it.

@rooksdo Do you think an issue should be created?

Issue #1874 will fix this.


TypeCobol.Test/Misc/TestFileConnectors.cs
```diff
@@ -0,0 +1,54 @@
...
+            // STUDENT-02
+            var student2 = GetFileControlEntry("STUDENT-F02");
+            Assert.IsNotNull(student2);
+            Assert.AreEqual(student2.FileName?.Name, "STUDENT-F02");
+            Assert.AreEqual(student2.ExternalDataSet?.Name, "UT-STUDENT-F02");
+
+            FileControlEntry GetFileControlEntry(string keyName)
+            {
+                return fileConnectors.FirstOrDefault(pair => pair.Key.Name == keyName).Value;
+            }
```
fm-117, line 51: Why pick entries by name ? I think it is better to iterate over the file connectors dictionary and check pairs in order because:
- code would be simpler and faster
- order of creation is constrained by syntax so it is correct to check the order in dictionary

fm-117 (changes requested):

TypeCobol.Test/Misc/TestFileConnectors.cs
```diff
@@ -0,0 +1,47 @@
...
+                {"STUDENT-F01", "UT-STUDENT-F01"},
+                {"STUDENT-F02", "UT-STUDENT-F02"}
+            };
+
+            var folder = Path.Combine("Parser", "Programs", "Cobol85");
+            var compilationUnit = ParserUtils.ParseCobolFile("FileControl2", DocumentFormat.RDZReferenceFormat, folder);
+            var fileConnectors = compilationUnit.TemporaryProgramClassDocumentSnapshot.Root.MainProgram.FileConnectors;
+
+            Assert.IsNotNull(fileConnectors);
+            Assert.AreEqual(fileConnectors.Count, expectedResults.Count);
```
fm-117, line 33: expected is first argument and actual is second, so you should reverse order for the three `Assert.AreEqual` calls.

fm-117 (changes requested):

fm-117 (approved):

smedilol (approved):

mayanje (approved):

rooksdo (approved):

Issue #1881, opened 2021-02-09
NullReferenceException on incomplete INSPECT statement


smedilol:
**Describe the bug**
**CHILD OF #1835**

Fix NullReferenceException on incomplete statements:
- inspect

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZZBCO0.
       procedure division.
           inspect
           GOBACK.
      
       END PROGRAM DVZZBCO0.
```

**Expected behavior**
No Exception.
The CodeElement must be created even if its properties are not set.

Technical hint:
See method `CreateInspectStatement` in `CodeStatementBuilder`.
For `inspect`, in `CodeStatementBuilder`, check that `context.alphanumericStorageArea()` is not null before using it.
If it's null, then property `statement.InspectedItem` will not be set.

Issue #1835, opened 2020-12-23
Test grammar on stand alone Cobol words


rooksdo:
### Description
Our AntLR grammar seems to fail on some Cobol words (e.g. WHEN).

We need to check all Cobol words as CodeElements to evaluate how to resolve this issue.

rooksdo:
Here the AntLR grammar parsing result file :

[CodeElementsResult.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/5751048/CodeElementsResult.xlsx)

This file contains : 
- the CodeElement type
- the diagnostic emitted
- the exception fired
- a flag that tells if an extra CodeElement was produced [No consumed token]

smedilol:
I suggest to first fix NullReferenceException on these CodeElement:
- EXEC
- EXECUTE
- INSPECT
- RERUN
- SELECT

For END- statements, it seems CodeElement are created. The error comes from node parsing phase.

Then we are left with:
- ACCEPT
- AT
- DECLARE  -> unlikely to ask completion here
- DIVIDE
- GO
- ID  -> unlikely to ask completion here
- IDENTIFICATION -> unlikely to ask completion here
- MULTIPLY
- SEARCH
- SUBTRACT
- USE -> `USE DEBUGGING` works. This is the only statement that starts with `USE`.

What do we do for what is left?
- Create temporary dummy statements (quick but not elegant)
- Rewrite the grammar with CUPS. This will certainly fix the problem (longer but clean solution)
- Try to rewrite Antlr grammar to avoid the problem (no guarantee to fix the problem)

PR #1888, opened 2021-02-18
WI #1881 Correct NullReferenceException on incomplete INSPECT

delevoye:
WI #1881


smedilol (dismissed):

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolExpressionsBuilder.cs
```diff
@@ -1536,15 +1536,20 @@ internal ReceivingStorageArea CreateNumericStorageArea(CodeElementsParser.Numeri
...
-		{
-			var storageArea = new ReceivingStorageArea(StorageDataType.Alphanumeric,
-				CreateIdentifier(context?.identifier()));
+        {
+            var identifier = CreateIdentifier(context?.identifier());
+            if (identifier == null)
+            {
+                return null;
+            }
+            var storageArea = new ReceivingStorageArea(StorageDataType.Alphanumeric, identifier);
```
fm-117, line 1545: Ok this prevent the null identifier to be passed to the `ReceivingStorageArea` constructor but there are other usages where null could be passed. Should we change all of them in this PR ?

smedilol: I suggest to only fix this NRE in this PR, so we can advance on the related issue #1835.
We can use another issue to fix all potentiel NRE.

fm-117: @delevoye Could you please check usages of `CreateAlphanumericStorageArea` ? Now that it may return null, we must ensure that we have not created other potential null ref exceptions...

fm-117 (changes requested):

fm-117 (approved):

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolExpressionsBuilder.cs
delevoye: The other case of potential NRE has been deleted as it was a wrong information in most cases.

smedilol (approved):

rooksdo (approved):

mayanje (approved):

Issue #1876, opened 2021-02-01
New CLI option to specify COPY parsing


rooksdo:
### Description
This issue is complementary to #1870.

We need to have an option to specify that files given to CLI are COPYs. 

The available options in CLI should allow parsing a set of programs and a set of copybooks. 

Possible ways to do this are  :
1) Use the existing CLI option to detect file extension
2) Use 2 options to separate programs and copy files given to CLI, such as : `CLI -pgm pgmfile1 pgmfile2 ... -cpy cpyfile1 cpyfile2 ...`

fm-117:
Actually this is not required for our enhanced Cobol Editor since CLI won't be used. LanguageServer however requires the nature of document, see #1892.

rooksdo:
Closing this issue regarding @fm-117 comment.

PR #1890, opened 2021-02-19
1870 direct copy parsing

### LINKED ISSUE #1870: Allow direct parsing of copy
### Description

Enable the parser to parse directly a copy.

How to do it :
 - When parsing a copy, generate a mock skeleton program in which the content of the target copy is injected in `WORKING-STORAGE SECTION`.
 - Any diagnostic would be reported on the copy line using an offset (due to the use of the skeleton program)

fm-117:
Fixes #1870.

TODO:
- ~~finish integration with `FileCompiler` to let client classes use direct copy parsing~~
- ~~add new unit tests for copy parsing~~

Content of this PR:
- Initial scan state is now defined only by FileCompiler
- Implementation of direct copy parsing based on 3 features
  - wrapping copy source code into a fake program
  - auto replace PartialCobolWords
  - check and remove empty root level number
- Integration up to FileCompiler
  - integration with high-level APIs will have to be done later (see #1892 for LSP server and #1876 for CLI)


smedilol (changes requested):

TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -149,15 +151,14 @@ public void OnStopRefreshParsingStep()
         /// This method does not scan the inserted text lines to produce tokens.
         /// You must explicitely call UpdateTokensLines() to start an initial scan of the document.
         /// </summary>
-        public CompilationDocument(TextSourceInfo textSourceInfo, IEnumerable<ITextLine> initialTextLines,
-            TypeCobolOptions compilerOptions, IProcessedTokensDocumentProvider processedTokensDocumentProvider, List<RemarksDirective.TextNameVariation> copyTextNameVariations) :
-            this(textSourceInfo, initialTextLines, compilerOptions, processedTokensDocumentProvider, null, copyTextNameVariations)
+        public CompilationDocument(bool isCopy, bool isImported, TextSourceInfo textSourceInfo, IEnumerable<ITextLine> initialTextLines, TypeCobolOptions compilerOptions, IProcessedTokensDocumentProvider processedTokensDocumentProvider,
```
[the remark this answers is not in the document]
fm-117: Not exactly, `isImported` means that the source is included in another document through a `COPY` or `EXEC SQL INCLUDE` instruction. It cannot be used with `isCopy` set to `false` so in a sense it also means "use direct parsing" but from the caller point of view, I thought it would be clearer to have the two parameters.

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    public class CopyWrappedAsProgramTokensDocument : ProcessedTokensDocument
+    {
+        private static IEnumerable<Token> BeginProgramSkeleton(ITokensLine targetLine, string copyName)
+        {
+            if (targetLine == null) yield break;
+
+            //IDENTIFICATION DIVISION.
+            //PROGRAM-ID. <copyName>.
+            yield return new Token(TokenType.IDENTIFICATION, 0, 0, targetLine);
```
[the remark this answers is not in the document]
fm-117: done

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            var columnsLayout = textSourceInfo.ColumnsLayout;
+            var start = columnsLayout == ColumnsLayout.CobolReferenceFormat ? 7 : 0;
+            var indent = new string(' ', start);
+
+            var lines = new ImmutableList<CodeElementsLine>.Builder(ImmutableList<CodeElementsLine>.Empty);
+
+            //Begin fake program skeleton
+            lines.Add(IdentificationDivision());
+            lines.Add(ProgramId());
```
smedilol, line 27: We also need `Decimal point is comma` by default:
```cobol
       ENVIRONMENT DIVISION.                 
       CONFIGURATION SECTION.                
       SOURCE-COMPUTER.  IBM-370.            
       SPECIAL-NAMES. DECIMAL-POINT IS COMMA.
```
This can be an option to insert or not this block of code.

smedilol (changes requested):

TypeCobol/Compiler/Scanner/MultilineScanState.cs
```diff
@@ -122,7 +121,7 @@ public void AddSymbolicCharacter(string tokenText)
         /// <summary>
         /// Initialize scanner state
         /// </summary>
-        public MultilineScanState(bool insideDataDivision, bool insideProcedureDivision, bool insidePseudoText, bool insideSymbolicCharacterDefinitions, 
+        private MultilineScanState(bool insideDataDivision, bool insideProcedureDivision, bool insidePseudoText, bool insideSymbolicCharacterDefinitions,
```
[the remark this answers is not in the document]
fm-117: `WithDebuggingMode` is read-only. By removing the parameter, we lose the option to start parsing in debugging mode. What about putting the encoding as first param and make all 3 booleans optional with a default value of `false` ?

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
fm-117: Ok to add this line in this PR, could we work on the optional/configurable aspect in another issue ?

fm-117: Well in fact I'm not sure we do need them because they won't affect program structure and `decimalPointIsComma` is set to `true` in direct copy parsing mode.

TypeCobol/Compiler/Scanner/MultilineScanState.cs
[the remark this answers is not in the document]
smedilol: Ok for default values

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
smedilol: Ok, in this case, we just need to have an unit test about that. I didn't finish to check unit tests.

smedilol: And the option to make it configurable can be done only if someone requests it.

fm-117: Done (5be38c51f6c5993d4178628570dba936c77faa3b)

TypeCobol/Compiler/Scanner/MultilineScanState.cs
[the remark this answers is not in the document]
fm-117: Done (f98a716ddcae85f9c9be2730162d35c4453d120c)

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -622,17 +604,12 @@ void InitLine()
...
-        /// Scan an isolated token in the given context if not null or in following "default" context otherwise:
-        /// - insideDataDivision = true
-        /// - decimalPointIsComma = false
-        /// - withDebuggingMode = false
-        /// - encodingForAlphanumericLiterals = IBM 1147
-        /// - default compiler options
+        /// Scan an isolated token in the given context.
         /// </summary>
-        public static Token ScanIsolatedToken(string tokenText, out Diagnostic error, MultilineScanState scanContext = null)
+        public static Token ScanIsolatedToken(string tokenText, MultilineScanState scanContext, out Diagnostic error)
```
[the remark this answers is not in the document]
fm-117: Done (d80e8a933149fd4b06c793a2f1aa5ffa6d71029a)

TypeCobol.Test/Parser/Copies/AutomaticTagReplacing.rdz-Nodes.txt
```diff
@@ -0,0 +1,9 @@
+--- Nodes ---
+?
+  AutomaticTagReplacing.rdz.cpy
```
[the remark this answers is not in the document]
fm-117: Done (0db0ad7de4967233306dc4b1bf79af2cb1f6c915)

TypeCobol/Compiler/CompilationDocument.cs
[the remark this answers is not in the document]
fm-117: Done (968a5e72f3f656d2fa21a7b4c388457143350a5a)

smedilol (changes requested):

mayanje:
Why there is no CLI option for this feature?

smedilol (changes requested):

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
```diff
@@ -0,0 +1,143 @@
...
+
+namespace TypeCobol.Compiler.Preprocessor
+{
+    /// <summary>
+    /// Special ProcessedTokensDocument that wraps original lines of a copy into a fake program
+    /// to allow full parsing and analysis on copies.
+    /// </summary>
+    public class CopyWrappedAsProgramTokensDocument : ProcessedTokensDocument
+    {
+        private static ISearchableReadOnlyList<CodeElementsLine> Wrap(ISearchableReadOnlyList<CodeElementsLine> originalLines, TextSourceInfo textSourceInfo)
```
smedilol, line 17: I'm not conviced by this wrapping of CodeElementsLine.
It breaks the logic of the parser.
Lines defined in CompilationDocument are not supposed to be altered, hence the use of an ISearchableReadOnlyList.


Is it not possible to use the mechanism of Iterator (like AutoReplacePartialWordsTokensLinesIterator) to return all tokens generated in this class ?
I mean a new iterator on top of AutoReplacePartialWordsTokensLinesIterator. This new iterator will :
- first returns token before the copy content (id division, program Id, data division, working-storage, root 01).
- returns the content of the copy
- returns END PROGRAM tokens.
I think we've already talked about this solution.


This mimic the behavior of copy inclusion without altering Lines of CompilationDocument.

smedilol (commented):
@mayanje CLI option for this feature is in #1876. It has been postpone as we don't need it for the moment.

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
smedilol: Another problem if lines are modified: On LSP, we use differents kinds of object to retrieve Lines:
- CompilationDocument.CobolTextLines
- ProcessedTokensDocument.Lines
- CodeElementsDocument.Lines

Depending on the case, we will have generated CodeElement or not.
For example `OnHover` will detect `IDENTIFICATION DIVISION` CodeElement if you hover on the first line of the COPY.


After secod thought, you could use `CodeElementTokenizer `to return generated CodeElement instead of generating Token.
This way you don't have to worry about Incremental mode and CodeElement will not be attached to Lines.

TypeCobol.Test/Utils/ParserUtils.cs
```diff
@@ -32,15 +32,14 @@ public static CompilationDocument ScanCobolFile(string relativePath, string text
...
-            compiler.CompileOnce();
+            FileCompiler compiler = new FileCompiler(null, textName, project.SourceFileProvider, project, documentFormat.ColumnsLayout, new TypeCobolOptions(), null, isCopy, project);
+            compiler.CompileOnce(ExecutionStep.Preprocessor, false, false);
 
-            return compiler.CompilationResultsForCopy;
+            return compiler.CompilationResultsForProgram;
         }
 
-        public static CompilationUnit ParseCobolFile(string textName, DocumentFormat documentFormat = null, string folder = null, ExecutionStep execToStep = ExecutionStep.SemanticCheck)
+        public static CompilationUnit ParseCobolFile(string textName, string folder, bool isCopy = false, DocumentFormat documentFormat = null, ExecutionStep execToStep = ExecutionStep.SemanticCheck)
```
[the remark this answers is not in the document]
fm-117: C# does not allow this. "A default parameter value of a reference type other than string can only be initialized with null"

TypeCobol/Compiler/Preprocessor/CopyWrappedAsProgramTokensDocument.cs
fm-117: Done in 967cef7e9ab7ebe1b99961ffab15748017bb9b5a.

smedilol: After a quick test, it works fiine now with LSP.
OnHover now detect the correct CodeElement.

TypeCobol/Compiler/Scanner/ScannerStep.cs
```diff
@@ -16,101 +15,74 @@ public static class ScannerStep
...
                 {
-                    if (groupIsContinuationGroup)
+                    lineGroup.Add(tokensLine);
+
+                    //Peek at next line type to decide whether to scan the current line group or continue to accumulate
+                    var nextTokensLine = documentLinesEnumerator.MoveNext() ? documentLinesEnumerator.Current : null;
+                    switch (nextTokensLine?.Type)
                     {
-                        //Group must be scanned as a whole
-                        if (lastScanState == null)
```
smedilol, line 67: `if (lastScanState == null)` has been removed
`lastScanState` cannot be `null` or is `null` handled by called method ?

fm-117: `lastScanState` is expected to be non-null all along the process. It is initialized with non-null value passed by the caller (and parameter has the `NotNull` attribute) then values are taken from lines so they are guaranteed not-null at this point.

smedilol (dismissed):

TypeCobol.Test/Parser/Copies/WithDiagnostics.rdzMix.txt
```diff
@@ -0,0 +1,14 @@
+Line -1[0,0] <37, Warning, General> - Warning: "END PROGRAM" is missing.
```
rooksdo, line 1: I'm not sure about this diagnostic since we're parsing a copy and there's no 'PROGRAM-ID' either.

rooksdo (changes requested):

TypeCobol.Test/Parser/Copies/WithDiagnostics.rdzMix.txt
fm-117: Internally we are always parsing a program, the diagnostic here is created because the parsing is interrupted and the END PROGRAM is never reached. But I think this will change with #1897. I'm waiting for full validation of this PR before merging with latest devs on develop. I'll keep this conversation open until then.

mayanje (dismissed):

TypeCobol.Test/Parser/Copies/WithDiagnostics.rdzMix.txt
rooksdo: Ok let's see the result then. Ideally, the end-user should not messages relative to the wrapping program skeleton.

TypeCobol.Test/Parser/Preprocessor/PreprocessorUtils.cs
```diff
@@ -10,12 +10,19 @@
...
 {
-    internal class PreprocessorUtils
+    internal static class PreprocessorUtils
     {
         private static readonly DocumentFormat _Format = new DocumentFormat(Encoding.Unicode, EndOfLineDelimiter.CrLfCharacters, 0, ColumnsLayout.CobolReferenceFormat);
+        private static readonly MultilineScanState _ScanState = new MultilineScanState(false, false, false, Encoding.Unicode);
 
         public static readonly string Root = "Parser" + Path.DirectorySeparatorChar +"Preprocessor";
 
+        private static ProcessedTokensDocument GetProcessedTokensDocument(this CompilationProject compilationProject, string testName)
```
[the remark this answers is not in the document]
fm-117: Changed to a dedicated `FileCompiler`.

fm-117:
See rebased version #1933.

Issue #1687, opened 2020-03-06
Exception on trying to add an element with same key


rooksdo:
See #1634

```Cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZS0OSM.
       PROCEDURE DIVISION.
       declare procedure Toto.
       end-declare.
       declare procedure Toto.
       end-declare.
                call 
               GOBACK
               .
       END PROGRAM DVZS0OSM.
```
And ask completion after `call`.

![image](https://user-images.githubusercontent.com/57907033/76094926-b9e0cb80-5fc3-11ea-946b-a9d572d527d0.png)

Maybe it's related to line:
`functionDeclarationSignatureDictionary.Add(signatureInformation, proc);`
Protect for duplicate on this line.

delevoye:
@smedilol @mayanje @rooksdo @fm-117 
In this case, elements with the same key are added to a Dictionary and it causes the exception. The exception is raised because the program above is wrong (duplicate declaration of procedure Toto).
What is the expected solution to avoid exception?
- Add a distinctive element to `SignatureInformation` (e.g. an ID);
- Not adding the element if it is already present and add a Diagnostic;
- Something else?

smedilol:
I suggest the same behavior as Visual Studio:
- Not adding the element if it is already present.

There is already a Diagnostic if the procedure declaration is duplicated.

PR #1891, opened 2021-02-22
WI #1687 Protect duplicate add on completion

delevoye:
WI #1687


smedilol (approved):

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
```diff
@@ -176,8 +176,11 @@ public static IEnumerable<CompletionItem> CreateCompletionItemsForProcedures(IEn
                 ((object[])completionItem.data)[1] = signatureInformation;
 
                 //Store the link between the hash and the procedure. This will help to determine the procedure parameter completion context later. 
-                functionDeclarationSignatureDictionary.Add(signatureInformation, proc);
-                completionItems.Add(completionItem);
+                if (!functionDeclarationSignatureDictionary.ContainsKey(signatureInformation))
+                {
+                    functionDeclarationSignatureDictionary.Add(signatureInformation, proc);
+                    completionItems.Add(completionItem);
+                }
```
fm-117, line 183: This means that we only consider the first declaration, if that is what we expect, then ok.

fm-117 (approved):

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
rooksdo: @delevoye 
Can you add a test to show the result with the present PR? The test should show that no exception is fired but that there's still a diagnostic issued by the parser because of duplicate procedure declaration (use the test in issue #1687). You don't have to code anything more but the error should already be handled by the parser.

rooksdo (changes requested):

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
delevoye: @fm-117 
As @smedilol commented in issue #1687, the behavior of Visual Studio has been mimicked: Not adding the element if it is already present.

delevoye: @rooksdo 
Test has been added. As seen with you, it does not show the duplicate procedure declaration because the diagnostic is not created in this case.

smedilol (approved):

rooksdo (approved):

fm-117 (approved):

mayanje (approved):

PR #1898, opened 2021-02-26
1732 copy cpy cpx cpm usage bis

### LINKED ISSUE #1732: Allow for copy type (CPY or CPX/CPM) to be detected while parsing
### Description
Allow for copy type (CPY or CPX/CPM) to be detected while parsing. Here are the two declarations that should be supported :

```cob
01 MYCOPY.         COPY YMYCOPY.
```
or using a `REPLACING` clause

```cob
01 MYCOPY.         COPY YMYCOPY REPLACING :MYCOPY: BY MYCOPY.
```

**Specific pretreatment for CPY must also be supported :        (does not apply to CPX/CPM/LTNV)**

1. Processing copy suffixing if the length of copy name > 7 characters
```cob
01 MYCOPY.         COPY YMYCOPY1. *> Same behavior as now
01 MYCOPY.         COPY YMYCOPY1S. *> Suffixing by 'S'
01 MYCOPY.         COPY YMYCOPY1L. *> Suffixing by 'L' must also be accepted
```
2. Suppressing level-01 when it is already declared in the copy, as in (YMYCOPY declares a level-01) :
```cob
01 MYCOPY.         COPY YMYCOPY.
01 MYCOPY.         COPY YMYCOPY REPLACING :MYCOPY: BY MYCOPY.
```

Currently the parser guess that a COPY is a CPY when a COPY is declared without replacing clause and the 8th character of the COPY name is not a `L`.
This rule must be replaced by a true detection of CPY based on an external file which contains the list of all CPY name.

### Test 1
YMYCOPY (CPY)
```cob
       01 YMYCOPY.
           05 MYCOPY-DATA           PIC X.
```
**Test 1.a**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because level 01 is systematically erased when included copybook is CPY
       COPY YMYCOPY.
       PROCEDURE DIVISION.
           .
       END PROGRAM Pgm.
```
**Test 1.b**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok
       01 MYCOPY.     COPY YMYCOPY.
      *Ok suffixing YMYCOPY with '1'
      *All string matching 'MYCOPY' will be suffixed by '1' (giving MYCOPY1)
       01 MYCOPY1.    COPY YMYCOPY1.
       PROCEDURE DIVISION.
      *   Ok
           Move 'A'             to MYCOPY-Data
      *   Ok because a version of YMYCOPY is also declared suffixed by '1'
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
**Test 1.c**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok for syntax but the REPLACING clause doesn't work without placeholder
       01 MYCOPY1.    COPY YMYCOPY REPLACING ==MYCOPY== BY ==MYCOPY1==.
       PROCEDURE DIVISION.
      *   Ko MYCOPY1-Data is not defined (attempted suffixing by '1' failed because of placeholder)
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
### Test 2
YMYCOPY (CPY with placeholder :MYCOPY:)
```cob
       01 Y:MYCOPY:.
           05 :MYCOPY:-DATA           PIC X.
```
**Test 2.a**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because level 01 is systematically erased when included copybook is CPY
       COPY YMYCOPY.
       PROCEDURE DIVISION.
           .
       END PROGRAM Pgm.
```
**Test 2.b**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because placeholders are not replaced
       01 MYCOPY. COPY YMYCOPY.
       PROCEDURE DIVISION.
           .
       END PROGRAM Pgm.
```
**Test 2.c**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok
       01 MYCOPY. COPY YMYCOPY REPLACING ==:MYCOPY:== BY ==MYCOPY==.
      *Ok the placeholder can be replaced with an added suffix
       01 MYCOPY1. COPY YMYCOPY REPLACING ==:MYCOPY:== BY ==MYCOPY1==.
       PROCEDURE DIVISION.
      *   Ok
           Move 'A'             to MYCOPY-Data
      *   Ok
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
**Test 2.d**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok for declaration but suffixing YMYCOPY not working with placeholder
       01 MYCOPY1. COPY YMYCOPY1 REPLACING ==:MYCOPY:== BY ==MYCOPY==.
       PROCEDURE DIVISION.
      *   Ko because special CPY suffixing mechanism doesn't apply with placeholder
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
**Test 2.e**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok for declaration but suffixing YMYCOPY not working with placeholder, the REPLACING will suffix MYCOPY1
       01 MYCOPY1. COPY YMYCOPY1 REPLACING ==:MYCOPY:== BY ==MYCOPY1==.
       PROCEDURE DIVISION.
      *   Ok thanks to REPLACING clause
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
**Test 2.f**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok for declaration but suffixing YMYCOPY not working with placeholder 
      *Same as declaring {01 MYCOPY1 COPY YMYCOPY. REPLACING ==:MYCOPY:== BY ==MYCOPY==.}
       01 MYCOPY1. COPY YMYCOPY1 REPLACING ==:MYCOPY:== BY ==MYCOPY==.
       PROCEDURE DIVISION.
      *   Ok because special CPY suffixing mechanism doesn't apply with placeholder
           Move 'A'             to MYCOPY-Data
           .
       END PROGRAM Pgm.
```
### Test 3
YMYCOPY (CPX without placeholder :MYCOPY:)
```cob
       01 YMYCOPY.
           05 MYCOPY-DATA           PIC X.
```
**Test 3.a**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok because level 01 is not deleted for CPX/CPM
       COPY YMYCOPY.
       PROCEDURE DIVISION.
      *   Ok
           Move 'A'             to MYCOPY-Data
           .
       END PROGRAM Pgm.
```
**Test 3.b**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because no automatic suffixing mechanism for CPX and YMYCOPY1 doesn't exist
       COPY YMYCOPY1.
       PROCEDURE DIVISION.
           .
       END PROGRAM Pgm.
```
**Test 3.c**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok for syntax but the REPLACING clause doesn't work without placeholder
       COPY YMYCOPY REPLACING ==MYCOPY== BY ==MYCOPY1==.
       PROCEDURE DIVISION.
      *   Ko because MYCOPY1-Data not declared
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
### Test 4
YMYCOPY (CPX with placeholder :MYCOPY:)
```cob
       01 Y:MYCOPY:.
           05 :MYCOPY:-DATA           PIC X.
```
**Test 4.a**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because placeholders not replaced
       COPY YMYCOPY.
       PROCEDURE DIVISION.
           .
       END PROGRAM Pgm.
```
**Test 4.b**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok
       COPY YMYCOPY REPLACING ==:MYCOPY:== BY ==:MYCOPY:==.
       PROCEDURE DIVISION.
      *   Ok
           Move 'A'             to MYCOPY-Data
           .
       END PROGRAM Pgm.
```
**Test 4.c**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because YMYCOPY1 doesn't exist
       COPY YMYCOPY1 REPLACING ==:MYCOPY:== BY ==:MYCOPY:==.
       PROCEDURE DIVISION.
           .
       END PROGRAM Pgm.
```
**Test 4.d**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ok
       COPY YMYCOPY REPLACING ==:MYCOPY:== BY ==:MYCOPY1:==.
       PROCEDURE DIVISION.
      *   Ok
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
**Test 4.e**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *Ko because YMYCOPY1 doesn't exist
       COPY YMYCOPY1 REPLACING ==:MYCOPY:== BY ==:MYCOPY1:==.
       PROCEDURE DIVISION.
      *   Ko because MYCOPY1-Data si not declared
           Move 'A'             to MYCOPY1-Data
           .
       END PROGRAM Pgm.
```
       

### Expected results
|     Test     |   ZComp  |
| ----------- | ----------- |
| 1.a          | Ko           |
| 1.b          | Ok           |
| 1.c          | Ko           |
| 2.a          | Ko           |
| 2.b          | Ko           |
| 2.c          | Ok           |
| 2.d          | Ko           |
| 2.e          | Ok           |
| 2.f          | Ok           |
| 3.a          | Ok           |
| 3.b          | Ko           |
| 3.c          | Ko           |
| 4.a          | Ko           |
| 4.b          | Ok           |
| 4.c          | Ko           |
| 4.d          | Ok           |
| 4.e          | Ko           |

**Technical tip**
As CPY copys are not allowed to be created anymore, we could extract all the existing CPY copys into a file and inspect it to find the copy name, in which case it would be considered of type CPY, otherwise of type CPX or CPM (no distinction between those two can be made using the parser at the present moment).

See property HasSuffix of TextNameVariation.

Se rule which check if there is no replace in `ReplaceTokensLinesIterator` :
`if (CopyReplacingDirective != null && CopyReplacingDirective.RemoveFirst01Level && CopyReplacingDirective.ReplaceOperations.Count == 0)`

mayanje:
**What have been done**

- Add a new option -ycpl {path} to specify a file that contains CPY copy names.
- in E-I mode taking in accoung suffix for COPY names having length >= 8 for CPY copies only using the cpy copy names file
- Review all E-I tests to use a cpy comy names file1732
- Add -ycpl option in the server also
- Add sxtra tests, + change tests syntax.

Resolves #1732


smedilol (changes requested):

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
```diff
@@ -0,0 +1,73 @@
...
+            _CpyCopyNames = cpyCopyNames;
+            if (_CpyCopyNames != null && bSort)
+                Array.Sort(_CpyCopyNames);
+        }
+
+        /// <summary>
+        /// Get the singleton instance. Create one if it does not exists
+        /// </summary>
+        /// <returns>The singleton instance</returns>
+        public static CopyNameMapFile Singleton { get; set; }
```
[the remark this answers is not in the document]
mayanje: But in this case this is exactly a Singleton there is only one Cpy List file for CLI or for the Server, and it simplify tests creation.

TypeCobol.LanguageServer.Test/LSRTestHelper.cs
```diff
@@ -86,6 +86,9 @@ public static void Test(string testFolderName, LsrTestingOptions lsrTestingOptio
             //Specify log file for TC LSP
             var tcLogFile = Path.Combine(workingDir.FullName, "TCLSPLog.txt");
 
+            //Check Cpy Copy names file
+            FileInfo cpyCopiesFile = new FileInfo(Path.Combine(testWorkingDirectory, "input", "CpyCopies.lst"));
```
fm-117, line 90: No need to create a `FileInfo` instance, `File.Exists` will do the job.

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
```diff
@@ -0,0 +1,73 @@
...
+    public class CopyNameMapFile
+    {
+        private string[] _CpyCopyNames;        
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="filepath">The Copy list File Path to map</param>
+        /// <param name="bSort">true to sort the resulting array of Copy names, false otherwise.</param>
+        public CopyNameMapFile(String filepath, bool bSort = false)
```
fm-117, line 27: Call the other constructor from this one to avoid code duplication.

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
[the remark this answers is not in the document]
fm-117: I agree with @smedilol, this would change the current option model. Besides, the CompilerDirectiveBuilders already have an instance of `TypeCobolOptions`.

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -368,12 +375,47 @@ public static OptionSet GetCommonTypeCobolOptions(TypeCobolConfiguration typeCob
...
+            //CPY Map File
+            try {
+                SetCpyCopiesFile(config.CPYMapFilePath);
+            } catch(Exception e) {
+                //Fail to read the Copy File Name, Log
+                System.IO.File.AppendAllText(config.LogFile != null ? config.LogFile : TypeCobolConfiguration.DefaultLogFileName, e.ToString());
+                if (config.CPYMapFilePath != null)
+                {
+                    errorStack.Add(ReturnCode.CopyNameMapFileError, TypeCobolConfiguration.ErrorMessages[ReturnCode.CopyNameMapFileError]);
+                }
```
fm-117, line 387: `VerifFiles` is already there to check external files.

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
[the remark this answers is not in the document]
smedilol: List of intrinsic, list of dependencies, list of analyzers, ... are not singleton even if you can only set one list per CLI.

TypeCobol.LanguageServer.Test/LSRTestHelper.cs
mayanje: But it's for getting the absolute file path also .

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
mayanje: But here it is not only for existence, but for any kind of exception while accessing or reading  the file.

CLI/test/ressources/haltOnMissingCopy_1_EI/input/CpyCopies.lst
```diff
@@ -0,0 +1,3 @@
+YCONVEA
```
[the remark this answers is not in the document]
mayanje: done in 2167af7

TypeCobol.Test/Parser/EILegacy/EIRemarksRemoveLevel01.rdz.cbl
```diff
@@ -1,11 +1,13 @@
 ﻿000000*Don't except any errors
```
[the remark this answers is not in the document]
mayanje: done in 2167af7

TypeCobol.Test/Utils/Comparator.cs
```diff
@@ -49,11 +49,25 @@ public void Init(string[] extensions = null, bool autoRemarks = false, bool Antl
...
-			try { Compiler.CompileOnce(); }
+			try
+            {
+                //Check for the presence of a CpyCopies.lst file.
+                FileInfo fi = new FileInfo(this.Compiler.CobolFile.FullPath);
+                FileInfo copyFileMap = new FileInfo(Path.Combine(fi.DirectoryName, fi.Directory.Name + "CpyCopies.lst"));
+                if (copyFileMap.Exists) { 
+                    TypeCobol.Compiler.Preprocessor.CopyNameMapFile.Singleton = null;
+                    TypeCobol.Compiler.Preprocessor.CopyNameMapFile.Singleton = new TypeCobol.Compiler.Preprocessor.CopyNameMapFile(copyFileMap.FullName, true);
+                }
```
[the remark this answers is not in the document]
mayanje: done in 2167af7

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            _CpyCopyNames = System.IO.File.ReadAllLines(filepath);
+            if (_CpyCopyNames != null && bSort)
```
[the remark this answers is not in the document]
mayanje: done in 2167af7

CLI/test/ressources/extractUsedCopies_EI/input/CpyCopies.lst
```diff
@@ -0,0 +1,7 @@
+YCONVEA
```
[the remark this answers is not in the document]
mayanje: done in 2167af7

CLI/test/CLITest.cs
```diff
@@ -367,6 +367,17 @@ public void TestCustomAnalyzers()
...
         }
+
+#if EUROINFO_RULES
+        /// <summary>
+        /// Test the usage of option -ycpl
+        /// </summary>
+        [TestMethod]
+        public void TestYCpyList()
+        {
+            CLITestHelper.Test("ycopylist", ReturnCode.Success);
```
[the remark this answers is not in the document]
mayanje: done in 2167af7

TypeCobol.Test/Parser/EILegacy/EITestCpyList2A.rdzPGM.txt
```diff
@@ -0,0 +1,15 @@
+--- Diagnostics ---
+Line 1[8,23] <37, Warning, General> - Warning: "END PROGRAM" is missing.
```
rooksdo, line 2: We should not have this diagnostic but as discussed with @fm-117 it seems it goes beyond the scope of this issue so we'll see about this another time.

rooksdo (changes requested):

TypeCobol.Test/Utils/Comparator.cs
```diff
@@ -49,11 +49,26 @@ public void Init(string[] extensions = null, bool autoRemarks = false, bool Antl
...
 		public void Parse() {
-			try { Compiler.CompileOnce(); }
+#if EUROINFO_RULES
+			try
+            {
+                //Check for the presence of a CpyCopies.lst file.
+                FileInfo fi = new FileInfo(this.Compiler.CobolFile.FullPath);
+                FileInfo copyFileMap = new FileInfo(Path.Combine(fi.DirectoryName, fi.Directory.Name + "CpyCopies.lst"));
+                if (copyFileMap.Exists) { 
+                    Compiler.CompilerOptions.CpyCopyNamesMap = new TypeCobol.Compiler.Preprocessor.CopyNameMapFile(copyFileMap.FullName);
```
smedilol, line 59: Initialization of `Compiler.CompilerOptions.CpyCopyNamesMap` could be moved into `Init` method just above.
And you don't need to set `Compiler.CompilerOptions.CpyCopyNamesMap = null;` as it's not a Singleton anymore.
So the method `Parse` can remains unchanged.

smedilol (changes requested):

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -34,7 +37,21 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
+
+        /// <summary>
+        /// Check if using the current Instance, the Given name corresponds to a CPY copy name. 
+        /// </summary>
+        /// <param name="name">The Copy's name</param>
+        /// <returns>true if the name is CPY Copys name, false otherwise.</returns>
+        public bool HasCpyCopy(string name)
+        {
+            return CpyCopyNamesMap != null ? CpyCopyNamesMap.HasCpyCopy(name) : false;
+        }
```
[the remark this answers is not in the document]
mayanje: This code has changed in commit 2167af7

TypeCobol.Test/Utils/Comparator.cs
mayanje: No because this is test and dirctory dependent. and it applies at compile time.

TypeCobol.Test/Parser/EILegacy/EITestCpyList2D.rdzPGM.txt
```diff
@@ -0,0 +1,13 @@
...
+--- Intrinsic:Namespace:Program:Global:Local
+-- DATA --------
+  MYCOPY1:Alphanumeric
+  MYCOPY1-DATA:Alphanumeric
+--- Intrinsic
+-- TYPES -------
+  BOOL:BOOL
+  DATE:DATE
+  CURRENCY:CURRENCY
+  STRING:STRING
```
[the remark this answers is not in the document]
mayanje: No, this is a mistake, it takes me a while to understand why in RDz test, this test is an error. There is a PDS that I will not give the name here, which contains a copy with the same name YMYCOPY, and this copy in put in test and production level, PDS copy take precedence on the copy of this test, that's why.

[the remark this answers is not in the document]
mayanje: But you are right, I didn't use the renamed copy in this test.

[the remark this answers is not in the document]
mayanje: Ok done in 2a78574

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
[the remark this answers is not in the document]
smedilol: I don't see the change.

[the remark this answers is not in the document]
mayanje: Sorry, it was a confusion. here the only thing thta I can do is: 
```csharp
public bool HasCpyCopy(string name) => CpyCopyNamesMap != null ? CpyCopyNamesMap.HasCpyCopy(name) : false;
```
done in 8af0ae7

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -63,6 +66,12 @@ public TypeCobolServer(IRPCServer rpcServer, Queue<MessageActionWrapper> message
         /// </summary>
         public bool UseEuroInformationLegacyReplacingSyntax { get; set; }
 
+#if EUROINFO_RULES
+        /// <summary>
+        /// The Instance of the Cpy Copy names Map
+        /// </summary>
+        public CopyNameMapFile CpyCopyNamesMap { get; set; }
```
fm-117, line 73: No need to have the real object at this point, should be a `string`.

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -218,9 +227,12 @@ protected override InitializeResult OnInitialize(InitializeParams parameters)
             var rootDirectory = new DirectoryInfo(parameters.rootPath);
             string workspaceName = rootDirectory.Name + "#" + parameters.processId;
 
-            // Initialize the workspace.
+            // Initialize the workspace.            
+#if EUROINFO_RULES
+            this.Workspace = new Workspace(rootDirectory.FullName, workspaceName, _messagesActionsQueue, Logger, CpyCopyNamesMap);
```
fm-117, line 232: Why a new constructor instead of setting a property just like the other options ?

TypeCobol.LanguageServer/TypeCobolServerHost.cs
```diff
@@ -296,6 +309,15 @@ static int Main(string[] args)
...
 
+#if EUROINFO_RULES
+                try
+                { //Read Cpy Copy names file.
+                    typeCobolServer.CpyCopyNamesMap = TypeCobol.Tools.Options_Config.TypeCobolOptionSet.GetCpyCopiesFile(CPYMapFilePath);
+                } catch(Exception e)
+                {
+                    logWriter.WriteLine(e.Message);                    
+                }
+#endif
```
fm-117, line 320: No need to load early, see my other remark on TypeCobolServer

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -56,6 +59,13 @@ public class Workspace
         public Queue<MessageActionWrapper> MessagesActionsQueue { get; private set; }
         private Func<string, Uri, bool> _Logger;
 
+#if EUROINFO_RULES
+        /// <summary>
+        /// The Instance of the Cpy Copy names Map
+        /// </summary>
+        public CopyNameMapFile CpyCopyNamesMap { get; set; }
```
fm-117, line 66: Workspace does not need to keep an instance of the copy name map. I think TypeCobolOptions should hold the value of the path to the file and CompilationProject should hold the instance of the loaded file.
- path is a new option so it works just like all other options
- CompilationProject is responsible for loading the file (either at construction time or lazily at first usage of the HasCpyCopy method)

fm-117 (changes requested):

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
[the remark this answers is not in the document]
fm-117: @smedilol you mean `CpyCopyNamesMap?.HasCpyCopy(name) ?? false;`

smedilol (changes requested):

smedilol (changes requested):

TypeCobol.Test/Parser/EILegacy/EITestCpyList1C.rdz.cbl
```diff
@@ -0,0 +1,12 @@
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. Pgm.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+      *Ok for syntax but the REPLACING clause doesn't work without placeholder
+       01 MYCOPY1.    COPY YMYCOPY. REPLACING ==MYCOPY== BY ==MYCOPY1==.
```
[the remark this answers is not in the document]
rooksdo: No there should not be any dot.

TypeCobol.Test/Parser/EILegacy/EITestCpyList2F.rdz.cbl
```diff
@@ -0,0 +1,13 @@
...
+       PROGRAM-ID. Pgm.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+      *Ok for declaration but suffixing YMECOPY not working with placeholder 
+      *Same as declaring 
+      *{01 MECOPY1. COPY YMECOPY. REPLACING ==:MECOPY:== BY ==MECOPY==.}
+       01 MECOPY1. COPY YMECOPY1 REPLACING ==:MECOPY:== BY ==MECOPY==.
+       PROCEDURE DIVISION.
+      *Ok special CPY suffixing mechanism doesn't apply with placeholder
+           Move 'A'             to MECOPY-Data
```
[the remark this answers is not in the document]
rooksdo: Agreed.

rooksdo (changes requested):

TypeCobol.LanguageServer/TypeCobolServer.cs
mayanje: To me it should be done here if it come from server option, because it can come from server options or from Client configuration changed option.

TypeCobol.LanguageServer/Workspace.cs
mayanje: This is an option,  at option resolution time it should be determinated if the file can be loaded, it shall not be a compilation error.

fm-117:
General remark on this PR: i think we could simplify the handling of the `CopyNameMapFile`.
- path to the file should be handled as all other existing options (meaning no new parameters in constructors, reducing impact on existing methods)
- the only class allowed to hold a reference to the loaded file would be `TypeCobolOptions`
  - actual loading of the file can be done either lazily at first usage or in the setter of the file path property
  - special case for the `DidChangeConfiguration` notification in LanguageServer, if the file path has not been changed pass the instance from the old TypeCobolOptions to the new one to avoid second loading

TypeCobol.Test/Parser/EILegacy/EITestCpyList2F.rdz.cbl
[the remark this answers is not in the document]
mayanje: done in 1ed782a

TypeCobol.Test/Parser/EILegacy/EITestCpyList1C.rdz.cbl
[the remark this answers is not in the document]
mayanje: done in 1ed782a

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
```diff
@@ -0,0 +1,51 @@
...
+using TypeCobol.Compiler.Diagnostics;
+using TypeCobol.Compiler.Directives;
+using TypeCobol.Tools.Options_Config;
+
+namespace TypeCobol.Compiler.Preprocessor
+{
+    /// <summary>
+    /// Class that implements, mapping of CPY copy names.
+    /// The Lookup is performed using  BinarySearch, thus it necessary for the underlying array of names to be sorted.
+    /// ALL COPY NAMES MUST BE IN UPPERCASE.
```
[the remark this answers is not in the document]
mayanje: done in c78db60

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -34,7 +37,18 @@ public bool UseEuroInformationLegacyReplacingSyntax
         }
 
 #if EUROINFO_RULES
+        /// <summary>
+        /// The Instance of the Cpy Copy names Map
+        /// </summary>
+        public CopyNameMapFile CpyCopyNamesMap { get; set; }
```
fm-117, line 43: In my opinion should be a private field, and we would have a public `CpyCopyNamesMapFilePath` string property. Load would be performed in the setter of this `CpyCopyNamesMapFilePath` property.

A `Debug.Assert` in `HasCpyCopy(string name)` method would check that either:
- the path is null and then the file is null
- OR the path is not null and then the file is not null too

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -35,6 +38,10 @@ public class TypeCobolConfiguration : ITypeCobolCheckOptions
         public const string DefaultLogFileName = "TypeCobol.CLI.log";
 
 #if EUROINFO_RULES
+        /// <summary>
+        /// The Instance of the Cpy Copy names file
+        /// </summary>
+        public CopyNameMapFile CpyCopyNamesMap { get; set; }
```
fm-117, line 44: `TypeCobolConfiguration` should deal with path only.

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -368,12 +389,50 @@ public static OptionSet GetCommonTypeCobolOptions(TypeCobolConfiguration typeCob
...
+                    TypeCobolConfiguration.DefaultCopyNameFile);
+                if (!File.Exists(cpyMapFilePath))
+                    cpyMapFilePath = null;
+            }
+            if (cpyMapFilePath != null)
+            {
+                return new CopyNameMapFile(cpyMapFilePath);                    
+            }
+            return null;
+        }
```
fm-117, line 434: Would go in TypeCobolOptions (setter of `CpyCopyNamesMapFilePath` as part of the load process). Also don't forget to check if the path is rooted (`Path.IsPathRooted`) to allow full path and relative path to be used.

fm-117 (changes requested):

rooksdo (dismissed):

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
mayanje: doen in e4234a0

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
mayanje: done in e4234a0

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: done in e4234a0

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -483,6 +502,9 @@ public void DidChangeConfigurationParams(string[] arguments)
             var options = TypeCobolOptionSet.GetCommonTypeCobolOptions(Configuration);
 
             var errors = TypeCobolOptionSet.InitializeCobolOptions(Configuration, arguments, options);
+#if EUROINFO_RULES
+            Configuration.CpyCopyNamesMap = Configuration.CpyCopyNamesMap??CpyCopyNamesMap;
+#endif
```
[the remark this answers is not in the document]
mayanje: done in e4234a0

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -156,7 +175,7 @@ public Workspace(string rootDirectoryFullName, string workspaceName, Queue<Messa
             var defaultDocumentFormat = new DocumentFormat(Encoding.GetEncoding("iso-8859-1"), EndOfLineDelimiter.CrLfCharacters, 80, ColumnsLayout.CobolReferenceFormat);
             this.CompilationProject = new CompilationProject(
                 _workspaceName, _rootDirectoryFullName, Helpers.DEFAULT_EXTENSIONS, defaultDocumentFormat,
-                new TypeCobolOptions(), null); //Initialize a default CompilationProject - has to be recreated after ConfigurationChange Notification
+                new TypeCobolOptions(Configuration), null); //Initialize a default CompilationProject - has to be recreated after ConfigurationChange Notification
```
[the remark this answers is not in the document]
mayanje: So we ceate a default configuration and we don't use it. Here the default configurtion shall be used.

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -141,11 +151,20 @@ public bool IsEmpty
...
+#endif
         {
+#if EUROINFO_RULES
+            CpyCopyNamesMap = cpyCopyNamesMap;
+#endif
             MessagesActionsQueue = messagesActionsQueue;
             Configuration = new TypeCobolConfiguration();
+#if EUROINFO_RULES
+            Configuration.CpyCopyNamesMap = cpyCopyNamesMap;
+#endif
```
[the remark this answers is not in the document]
mayanje: done in e4234a0

TypeCobol.LanguageServer/TypeCobolServerHost.cs
mayanje: done in e4234a0

TypeCobol.LanguageServer/TypeCobolServer.cs
mayanje: done in e4234a0

TypeCobol.LanguageServer/TypeCobolServer.cs
mayanje: done in e4234a0

TypeCobol.LanguageServer.Test/LSRTestHelper.cs
```diff
@@ -102,6 +102,10 @@ public static void Test(string testFolderName, LsrTestingOptions lsrTestingOptio
                 useOutline ? "-ol" : "",
                 useCfg ? "-cfg=AsContent" : "");
 
+            //Check Cpy Copy names file
```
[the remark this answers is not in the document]
mayanje: done

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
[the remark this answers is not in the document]
mayanje: nice, done in 2f26fcb

rooksdo (approved):

CLI/src/CLI.cs
```diff
@@ -129,6 +129,21 @@ private ReturnCode Compile()
...
+            ReturnCode optionRetCode = typeCobolOptions.OptionStatusCode;
+            switch (optionRetCode)
+            {
+                case ReturnCode.Success:
+                    break;
+#if EUROINFO_RULES
+                case ReturnCode.CopyNameMapFileError:
+                    //Exit only if we fail to read a Cpy Name file other than the default one.
+                    if (_configuration.CpyCopyNamesMapFilePath.Equals(TypeCobolConfiguration.DefaultLogFileName))
+                        return optionRetCode;
```
fm-117, line 141: Is it required ? If an error occured while loading the file it will be logged, but still we can let the process continue. This would allow us to delete the `OptionStatusCode` property and associated checking.
Besides, the comment is misleading: here we stop only if the file path is the default one.
And I don't see why we should handle the error on the default path differently than on a user-supplied path.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -59,6 +59,16 @@ protected override void OnDidChangeConfiguration(string[] options)
...
+#if EUROINFO_RULES
+            if (!options.Contains("-ycpl") && this.CpyCopyNamesMapFilePath != null)
+            {
+                string[] newOptions = new string[options.Length + 2];
+                Array.Copy(options, newOptions, options.Length);
+                newOptions[options.Length] = "-ycpl";
+                newOptions[options.Length + 1] = this.CpyCopyNamesMapFilePath;
+                options = newOptions;
+            }
+#endif
```
fm-117, line 71: ?? I'm confused. Here we have received updated options from the user in `options` string array, we should not modify its content based on previously stored values.

TypeCobol.LanguageServer/TypeCobolServerHost.cs
```diff
@@ -295,7 +309,9 @@ static int Main(string[] args)
                 typeCobolServer.UseSyntaxColoring = UseSyntaxColoring;
                 typeCobolServer.UseOutlineRefresh = UseOutlineRefresh;
                 typeCobolServer.UseCfgDfaDataRefresh = UseCfg;
-
+#if EUROINFO_RULES
+                typeCobolServer.CpyCopyNamesMapFilePath = CpyCopyNamesMapFilePath ?? TypeCobolConfiguration.DefaultCopyNameFile;
```
fm-117, line 313: I think the use of the default file should be handled directly in `TypeCobolOptions`. It would be factorized for both CLI and LanguageServer.

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -125,6 +128,30 @@ public bool TimerDisabledOption
...
+                try
+                {
+                    this.CompilationProject.CompilationOptions.CpyCopyNamesMapFilePath = value;
+                }
+                catch (Exception e)
+                {
+                    this._Logger(TypeCobolConfiguration.ErrorMessages[ReturnCode.CopyNameMapFileError] + ":" + e.Message, null);
+                }
+            }
+        }
```
fm-117, line 152: Ok this is to ensure that we have the correct file path at initialization time but this is specific to this particular option. Other options are not considered at initialization time, they are in the compilation project only on first `DidChangeConfiguration` notification. I'm not sure we should make an exception here. See my remark below.

TypeCobol.LanguageServer/Workspace.cs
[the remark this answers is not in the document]
fm-117: But this is still partial, here you still use default extensions, default document format and no analyzers. All these are correctly set when we receive the first `DidChangeConfiguration` notification. I think it is better to keep everything default here (meaning `new TypeCobolOptions()`) until we figure out how this initialization problem should be solved. In its current form the language server only works if we send at least one `DidChangeConfiguration` notification and this is independent from this PR.

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -501,7 +527,12 @@ public void DidChangeConfigurationParams(string[] arguments)
                 Configuration.ExecToStep = ExecutionStep.QualityCheck; //Language Server does not support Cobol Generation for now
 
             var typeCobolOptions = new TypeCobolOptions(Configuration);
-
+#if EUROINFO_RULES
+            ReturnCode optionRetCode = typeCobolOptions.OptionStatusCode;
```
fm-117, line 531: Same remark as in CLI, personally I think the log done in `TypeCobolOptions` is enough and that we should not really worry about these kind of exceptions. Checking the file path is enough to ensure a safe load.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -33,8 +37,66 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
+        {
+            get;
+            private set;
+        }
+
 #if EUROINFO_RULES
+        /// <summary>
+        /// The Instance of the Cpy Copy names Map
+        /// </summary>
+        private CopyNameMapFile CpyCopyNamesMap { get; set; }
```
fm-117, line 53: Can be turned into a field now.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -33,8 +37,66 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
+        {
+            get
+            {
+                return _CpyCopyNamesMapFilePath; 
+            }
+            set
+            {
+                if (value == null)
+                {
+                    CpyCopyNamesMap = null;
```
fm-117, line 70: file path should also be set to null here.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -33,8 +37,66 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
+                }
+                else if (!value.Equals(_CpyCopyNamesMapFilePath))
```
fm-117, line 72: Use case insensitive comparison for paths.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -33,8 +37,66 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
+        /// <exception cref="Exception">Any exception if an error occurs.</exception>
+        /// <returns>the CopyNameMapFile instance if one is available, null otherwise</returns>
+        public static CopyNameMapFile GetCpyCopiesFile(string cpyCopiesFilePath)
+        {
+            if (cpyCopiesFilePath != null)
+            {
+                return new CopyNameMapFile(cpyCopiesFilePath);
+            }
+            return null;
+        }
```
fm-117, line 99: No check on the relative/rooted nature of the path ? Use of root process location ?

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -56,6 +59,14 @@ public class TypeCobolConfiguration : ITypeCobolCheckOptions
         public string RawMaximumDiagnostics;
         public string RawOutputFormat = "0";
 
+#if EUROINFO_RULES
+        public static string DefaultCopyNameFile => Path.Combine(Path.GetDirectoryName(System.Reflection.Assembly.GetEntryAssembly().Location),"COPIES_CPY.txt");
```
fm-117, line 63: Actually I don't think we need a default value for this. This is an EI-specific option so it will be used in a controlled environment and we can ensure that it will be correctly set. For public users, a default value would not give any benefit.
@rooksdo, you wanted a default value, what use case was on your mind ?

fm-117 (changes requested):

CLI/src/CLI.cs
mayanje: So let be sure now that we always have a default file present, otherwise we could never be able to tests or run CLI without that file. We still have to do it internally using our TFS build process.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: It is not in fact the previous value, but the value that was passed when the server was started, it cane be the default file, but at this moment I don't know if we will manage this option in the client.

TypeCobol.LanguageServer/TypeCobolServerHost.cs
mayanje: We cannot decide that TypeCobolOptions should set itself to default.

TypeCobol.LanguageServer/Workspace.cs
mayanje: The copy name file is a very special option, the difficulty here, is that sometime we consider it as a global setting, and sometime we consider it as a specific setting.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
fm-117: Either we allow the client to change the path on-the-fly or we only accept value at startup time. If the path may be changed we'll have to check if it's a new value (to avoid loading multiple times the same file).

TypeCobol.LanguageServer/TypeCobolServerHost.cs
fm-117: Here is how I see the logic:
- if the user don't use the option, use the default path
- if the path (relative or absolute) is invalid, log a message and continue with an empty CPY table
- if the path is correct, load the file

All of this can be implemented in `TypeCobolOptions`. I think the logic should be the same for CLI and LanguageServer.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -33,8 +37,66 @@ public bool UseEuroInformationLegacyReplacingSyntax
...
         }
 
+        /// <summary>
+        /// Status code of these option.
+        /// </summary>
+        public ReturnCode OptionStatusCode
+        {
+            get;
+            private set;
+        }
```
fm-117, line 47: I'm not ok with this new property because it means that every client code instanciating a `TypeCobolOptions` should check the status and adapt. The simplest way to work around this is to consider the constructor always safe meaning here that when we fail to load the CPY file, we should just log and keep going with a null/empty CPY table (see my previous remark about the logic).

fm-117 (changes requested):

TypeCobol/Parser.cs
```diff
@@ -145,9 +148,20 @@ public IEnumerable<CodeElement> CodeElements
...
+        {
+            var parser = new Parser();
+            var typeCobolOption = new TypeCobolOptions() { ExecToStep = ExecutionStep.Generate };
+            typeCobolOption.AutoRemarksEnable = autoRemarks;
+            typeCobolOption.CpyCopyNamesMapFilePath = cpyCopyNamesFile;
+            parser.Init(path, typeCobolOption, format, copies, analyzerProvider);
+            parser.Parse(path);
+            return parser;
+        }
+#endif
```
smedilol, line 162: In order to avoid a method specific to EUROINFO_RULES can you use a more generic signature like this:
```suggestion
        public static Parser Parse(string path, DocumentFormat format, IList<string> copies = null, IAnalyzerProvider analyzerProvider = null, TypeCobolOptions typeCobolOption)
        {
            ...
        }
#endif
```

smedilol (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: So we decided that, -ycpl should not be a dynamic LSP client configuration option but only a command line LSP server option.

TypeCobol.LanguageServer/Workspace.cs
mayanje: I Could not change this now, because this options will not be set using DidChangeConfiguration, but only using sever command line options. So it is really a specific option.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
fm-117: yep still true

TypeCobol/Parser.cs
fm-117: I agree we should rely more on `TypeCobolOptions` class to pass options in different test cases otherwise the signatures will have to change frequently.

fm-117 (changes requested):
Some on my remarks have not been addressed, we should discuss how we can avoid multiple file reads.

TypeCobol.LanguageServer/Workspace.cs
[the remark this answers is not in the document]
fm-117: Changed in favor of a command-line only option.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
fm-117: In the end I kept the property as originally intended, sorry for the misleading remark.

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
fm-117: done

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
fm-117: property removed

TypeCobol/Parser.cs
fm-117: done

CLI/test/CLI.Test.csproj
```diff
@@ -56,7 +56,43 @@
...
+    </None>
+    <None Include="..\..\TypeCobol.Test\Parser\EILegacy\YXXXFAL.cpy">
+      <Link>ressources\ycopylist\input\YXXXFAL.cpy</Link>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+    <None Include="..\..\TypeCobol.Test\Parser\EILegacy\\YXXXFAL.cpy">
+      <Link>ressources\ycopylist\input\YXXXFAL.cpy</Link>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Include="..\..\TypeCobol.Test\Parser\EILegacy\\YXXXFAS.cpy">
```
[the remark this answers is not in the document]
fm-117: fixed

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -500,8 +509,10 @@ public void DidChangeConfigurationParams(string[] arguments)
             if (Configuration.ExecToStep >= ExecutionStep.Generate)
                 Configuration.ExecToStep = ExecutionStep.QualityCheck; //Language Server does not support Cobol Generation for now
 
+#if EUROINFO_RULES
+            Configuration.CpyCopyNamesMapFilePath = CpyCopyNamesMapFilePath ?? Configuration.CpyCopyNamesMapFilePath;
```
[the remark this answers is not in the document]
fm-117: We know test if we already have a loaded file

fm-117 (approved):
Ok for me, but I did not review the tests, only the new code.

rooksdo (approved):

fm-117:
This PR is ok but cannot be merged due to @smedilol's change request. See #1945.

Issue #1920, opened 2021-03-17
Lsp server : Not allow outline option to be changed by configuration change.


mayanje:
In v1.5.1 the outilne option can be changed thru configuration changed notification.
This behavior is not compatible with TypeCobol plugin v1.4, because now the outline view is no longer displayed.
So we should disable this possibility.

We should disable this option until TypeCol Plugin v1.5.
Because we need to have this option disabable, for LSR tests options without outline.
With CobolEditEI, this option is enabled only for the Team.

So the solution is to introduce a -team option, so that tis option will be taken in acount only if the -team option is also specified.

smedilol:
The option `-team` is not correct. It's not because currently our IDE restrict an option for our team that it will remains like this forever.
And a custom behavior of our private IDE should not influence how the server works.
The correct solution is to enable the Outline by default (like v1.4) and add a new option to disable the outline if needed.

This way, if another team is allowed to disable the outline, the server will not be impacted.

PR #1921, opened 2021-03-17
WI #1920 Add a -team option for LSP client

mayanje:
**What have been done.**

a new -team option is handled, so that -ol option from LSP client configuration changed notification is taken in account only if -team option is also specified.


smedilol (changes requested):
The option `-team` is not correct. It's not because currently our IDE restrict an option for our team that it will remains like this forever.
And a custom behavior of our private IDE should not influence how the server works.
The correct solution is to enable the Outline by default (like v1.4) and add a new option to disable the outline if needed.

This way, if another team is allowed to disable the outline, the server will not be impacted.

mayanje:
But the -team is a general option for any team allowed to access outline option, allowed team will be managed by the client.
-ol is already a server command line option, not need to add a new option for that. This option is for LSR tester that need to disable or enable -ol option without restarting the server, by by only changing LSP configuration.

smedilol:
As I've said, there is no reason to restrict Outline deactivation to `teams` **on the server side**. This is a client behavior.

mayanje:
Right now this is a client option for LSR purpose test only. Otherwise outline option is always enabled using server comand line, this is the client default option in v1.4 plugin. This option setting is hard coded in the client code, in v1.4 and CobolEditEI user cannot change it. Here we need to allow LSR test writer to disable this option, so we decide to introduce team specific option.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -52,7 +52,8 @@ protected override InitializeResult OnInitialize(InitializeParams parameters)
         /// <param name="options">Client's Options</param>
         protected override void OnDidChangeConfiguration(string[] options)
         {
-            this.UseOutlineRefresh = options.Contains("-ol");
+            if (options.Contains("-team"))
+                this.UseOutlineRefresh = options.Contains("-ol");
```
smedilol, line 56: As our client doesn't pass option `-ol`, just invert the logic. It would make more sense to rename the param to "Deactivate Outline Refresh" : `-dol`.
```suggestion
                this.UseOutlineRefresh = !options.Contains("-dol");
```

mayanje: Sorry doing like that makes most LSR test fail, this is not the moment to review the majority of LSR tests, the -team option is here the alternative option. 
Because doing like that will by default allow outilne on all LSR tests

smedilol: Just change the logic in LSRTestHelper.cs:
```C#
           testOptions += useOutline ? ",\"-ol\"" : "";
```
into
```C#
           testOptions += useOutline ? "":  ",\"-dol\"";
```
Then every tests will pass.

Or you can imagine a more complete option:
without parameter `-ol` : outline is activated by default
`-ol=true` : activate outline
`-ol=false` : activate outline

Anyway goals are:
- Don't introduce a specific option for Euro-Information if we don't need to
- The rule about our team is temporary and is not a security rule:
   -  We just want to hide complex options for the user. 
   - We might decide to open this option for every users in the future.
- Our current TypeCobol client allows users to de/activate Outline and the `-team` will break this behavior.
- The rule is already enforced on the client side, no need to duplicate it on the server side (if it was a security or business rule that would be different)
- Fall back to the behavior of 1.4.x : where Outline is activated by default

mayanje: Finally the combination of -dol an -ol options is the good choice for non-regression with v1.4.
done in 3bb41af.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -52,8 +52,11 @@ protected override InitializeResult OnInitialize(InitializeParams parameters)
         /// <param name="options">Client's Options</param>
         protected override void OnDidChangeConfiguration(string[] options)
         {
-            this.UseOutlineRefresh = options.Contains("-ol");
-            if(options.Contains("-cfg=AsFile"))
+            if (options.Contains("-dol"))
+                this.UseOutlineRefresh = false;
+            else if (options.Contains("-ol"))
+                this.UseOutlineRefresh = true;
```
rooksdo, line 58: @mayanje 
So what's the default value for 'this.UseOutlineRefresh' ? in case neither '-dol' nor '-ol' is passed. I prefer @smedilol's suggestion to have : 

- without parameter `-ol` : outline is activated by default
- `-ol=true` : activate outline
- `-ol=false` : activate outline

rooksdo (changes requested):

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: The default value is always -ol and in passed when the server is started thru the command line.

fm-117: Ok from what I understand, our plugin does not send the `-ol` option in config change notification so without this code, the outline will be disabled at first config change notiifcation.
We'll have to remove the `-dol` option after all our clients have migrated to the latest version (which will send correctly `-ol` option in notification).

fm-117 (dismissed):

TypeCobol.LanguageServer.Test/LSRTestHelper.cs
```diff
@@ -56,7 +56,7 @@ public static void Test(string testFolderName, LsrTestingOptions lsrTestingOptio
                 new DirectoryInfo(testWorkingDirectory + Path.DirectorySeparatorChar + "input" +
                                   Path.DirectorySeparatorChar + copyFolder).FullName.Replace(@"\", @"\\"));
             String testOptions = "";
-            testOptions += useOutline ? ",\"-ol\"" : "";
+            testOptions += useOutline ? "" : ",\"-dol\"";
```
smedilol, line 59: Line 102 in this file:
```C#
useOutline ? "-ol" : "",
```
must also be changed.

smedilol (changes requested):

TypeCobol.LanguageServer.Test/LSRTestHelper.cs
mayanje: The Idea here is that -dol is only for client option and LSR test (to keep compatibility with v1.4), command line option still to be -ol. In the future when 1.5 plugin will be available the -dol option will be removed. CobolEditEI client can use the -dol option for LSP configuration change. The server argument use -ol or not.
No no need to add -dol options as command line option.

fm-117: The "-dol" option is only used to prevent unwanted outline deactivation on a `DidChangeConfiguration` notification:
- with plugin v1.4 : "-ol" sent at startup, nothing on config change so outline is used initially and without "-dol" in config change we keep using it
- with plugin v1.5 (temporarily for compat) : "-ol" sent at startup, "-ol" or "-dol" sent in notification, so outline is used initially and kept or deactivated depending on which option is sent
- with plugin v1.5 (after migration) : "-ol" sent at startup, "-ol" sent or not in notification depending of client choice

The test code should not be modified, we can keep original code here.

```csharp
testOptions += useOutline ? ",\"-ol\"" : "";
```

fm-117 (changes requested):

mayanje:
For this temporary fix, handling -dol only in LSP DidConfigurationChange notification is enough, because even in CobolEditEI plugin only LSR should use it as team options, otherwise -o is always set. When plugin v1.5 will be ready we will have only change again these two lines and remove -dol option.

fm-117 (approved):

smedilol (approved):

rooksdo (approved):

Issue #1918, opened 2021-03-16
Subscripting error when using ::


rooksdo:
### Description

**Parser version : v1.5.1**

Subscripting seems to fail when using :: qualifyer, as in : 
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 StringT.
         02 DataT      Pic X(10000).
         02 LengthT    Pic 99999.
       PROCEDURE DIVISION.
      *Ok
             Move Low-Values       To StringT (1 : LengthT)
      *Ko syntax error
             Move Low-Values       To StringT (1 : StringT::LengthT)
             GOBACK
             .
       END Program Pgm.
```

rooksdo:
Closed by #1927

PR #1927, opened 2021-03-25
WI #1918 We must treat :: as a pure TypeCobol feature

mayanje:
**What have been done**

It is a sad and tricky issue, because it happends now that we must consider :: as a pure typecobol keyword when looking for
pattern ':' (cobol word chars)+ ':'
in a situation like the one of this issue for TypeCobol code: 

```cobol
1 :  String::LengthT
```
the sequene ":  String:" is captured by the pattern but it should not as :: is a TypeCobol Qualified access.
Thus I have no choice than to introduce a new option name -cob to instruct the scanner if the scanning is not marked as pure cobol file to consider token :: as the QualifierToken rather than the end of a pattern ':' (cobol word chars)+ ':'


TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -55,6 +60,7 @@ public TypeCobolOptions(TypeCobolConfiguration config)
             AutoRemarksEnable = config.AutoRemarks;
 #endif
             CheckEndAlignment = config.CheckEndAlignment;
+            this.IsCobolLanguage = config.IsCobolFile;
```
smedilol, line 63: I'd rather use the same name in TypeCobolOptions and TypeCobolConfiguration
```suggestion
            this.IsCobolLanguage = config.IsCobolLanguage ;
```

smedilol (changes requested):
Option `IsCobolLanguage` makes sense.

I'd be a good idea to launch test `CheckParserCobol85` twice: with and without `IsCobolLanguage` to true as our parser must works for both Cobol and TypeCobol.

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -55,6 +55,7 @@ public class TypeCobolConfiguration : ITypeCobolCheckOptions
         public string RawExecToStep = "6";
         public string RawMaximumDiagnostics;
         public string RawOutputFormat = "0";
+        public Boolean IsCobolFile;
```
fm-117, line 58: Maybe use `bool` instead of `Boolean` for a more C#-oriented style 😄.
Also this piece of information may be attached to `TextSourceInfo` instead of option. I think `IsCobolLanguage` or `IsCobolFile` is meaningful for the input file not really at option level. We could see it from a different angle like `EnableTypeCobol` which sounds more like an option.

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
```diff
@@ -243,7 +244,8 @@ public static OptionSet GetCommonTypeCobolOptions(TypeCobolConfiguration typeCob
                 { "diag.cea|diagnostic.checkEndAlignment=", "Indicate level of check end aligment: warning, error, info, ignore.", v => typeCobolConfig.CheckEndAlignment = TypeCobolCheckOption.Parse(v) },
                 { "log|logfilepath=", "{PATH} to TypeCobol.CLI.log log file", v => typeCobolConfig.LogFile = Path.Combine(v, TypeCobolConfiguration.DefaultLogFileName)},
                 { "cfg|cfgbuild", "Standard CFG build.", v => typeCobolConfig.CfgBuildingMode = CfgBuildingMode.Standard},
-                { "ca|customanalyzer=", "{PATH} to a custom DLL file containing code analyzers. This option can be specified more than once.", v => typeCobolConfig.CustomAnalyzerFiles.Add(v) }
+                { "ca|customanalyzer=", "{PATH} to a custom DLL file containing code analyzers. This option can be specified more than once.", v => typeCobolConfig.CustomAnalyzerFiles.Add(v) },
+                { "cob|-cobol", "Indicate that its a pure Cobol85 input file.", v => typeCobolConfig.IsCobolFile = true }
```
fm-117, line 248: Remove '-' to stay in line with other option descriptions.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: done in cf3b16c

TypeCobol/Tools/Options-Config/TypeCobolConfiguration.cs
mayanje: Sur sometime I write boolean (I think java) and completion change it , done in 8b26513

TypeCobol.Test/Parser/Preprocessor/TestReplaceDirective.cs
```diff
@@ -9,10 +9,12 @@ public static void CheckReplaceSingle()
...
         }
 
-        public static void CheckReplacePartial()
+        public static void CheckReplacePartial(bool cobol)
         {
-            string testName = "PgmReplacePartial";
+            string testName = cobol ? "PgmReplacePartial" : "PgmReplacePartialTC";
+            PreprocessorUtils.ReplaceProject.CompilationOptions.IsCobolLanguage = cobol;
             string result = PreprocessorUtils.ProcessReplaceDirectives(testName);
+            PreprocessorUtils.ReplaceProject.CompilationOptions.IsCobolLanguage = false;
```
fm-117, line 17: Can you add a param in `ProcessReplaceDirectives` instead of manipulating the options from outside ?

TypeCobol.Test/Parser/Scanner/TestTokenTypes.cs
```diff
@@ -246,7 +250,9 @@ public static void CheckPartialCobolWords()
                 "if (W-CCOMDE-UN-CHOIX(CCOMDI-:ZONE:) = 'D' or   ",
                 "replace ==:ZONE:== by ==SUPX== ==:SSPRO:== by ==CCTZ023B==."
             };
+            ScannerUtils.CompilerOptions.IsCobolLanguage = cobol;
             string result = ScannerUtils.ScanLines(testLines);
+            ScannerUtils.CompilerOptions.IsCobolLanguage = false;
```
fm-117, line 255: Same remark as in Preprocessor test.

TypeCobol.Test/TestCollection.cs
```diff
@@ -60,7 +60,8 @@ public void CheckScanner()
             TestTokenTypes.CheckPseudoText2();
             TestTokenTypes.CheckNumericLiterals();
             TestTokenTypes.CheckKeywordsAndUserDefinedWords();
-            TestTokenTypes.CheckPartialCobolWords();
+            TestTokenTypes.CheckPartialCobolWords(true);//For Pur Cobol
```
fm-117, line 63: typo: Pure

TypeCobol.Test/TestCollection.cs
```diff
@@ -144,21 +146,24 @@ public void CheckParserRobustness()
             TestTokenSource.Check_CobolTokenSource_WithStartToken();
         }
 
-        [TestMethod]
-        [TestCategory("Parsing")]
-        [TestProperty("Time", "fast")]
-        public void CheckParserCobol85()
+        /// <summary>
+        /// Check Parsre on Cobol85 source code
```
fm-117, line 150: typo: Parse

TypeCobol.Test/TestCollection.cs
```diff
@@ -144,21 +146,24 @@ public void CheckParserRobustness()
...
+        /// </summary>
+        /// <param name="cobol">true if it is rellay for pure Cobol85 language, rather than TypeCobol, false otherwise</param>
```
fm-117, line 152: typo: really

TypeCobol.Test/Utils/Comparator.cs
```diff
@@ -179,6 +182,7 @@ public int GetTestCount()
 				foreach (var comparator in comparators) {
                     Console.WriteLine(comparator.paths.Result + " checked with " + comparator.GetType().Name);
 					var unit = new TestUnit(comparator, debug);
+                    unit.IsCobolLanguage = IsCobolLanguage;
```
fm-117, line 185: strange indent

TypeCobol.Test/Utils/Comparator.cs
```diff
@@ -211,7 +215,8 @@ public int GetTestCount()
 				    }
 				}
 			}
-			if (errors.Length > 0) throw new Exception(errors.ToString());
+			if (errors.Length > 0)
+                throw new Exception(errors.ToString());
```
fm-117, line 219: why the indent change ?

fm-117 (changes requested):

TypeCobol.Test/Parser/Preprocessor/TestReplaceDirective.cs
mayanje: In fact PreprocessorUtils.ReplaceProject.CompilationOptions is a static shared resource, it is the resposability of the caller to do that.

TypeCobol.Test/Parser/Scanner/TestTokenTypes.cs
mayanje: same answer

TypeCobol.Test/Utils/Comparator.cs
mayanje: I needed to set a breakpoint here.

TypeCobol.Test/Utils/Comparator.cs
mayanje: I don't see it

TypeCobol.Test/TestCollection.cs
mayanje: done in 2ef0478

TypeCobol.Test/TestCollection.cs
mayanje: done in 2ef0478

TypeCobol.Test/TestCollection.cs
mayanje: done in 2ef0478

TypeCobol.Test/Parser/Preprocessor/TestReplaceDirective.cs
fm-117: Yes I remember this is some sort of hack, I had to deal with it in #1890. Ok.

TypeCobol.Test/Utils/Comparator.cs
fm-117: The file has a mix of tabs and spaces so indent is inconsistent. But it's ok you can leave it as it is.

TypeCobol.Test/Utils/Comparator.cs
fm-117: VS allows to set breakpoint on parts of line, anyway it does not matter.

TypeCobol.Test/Parser/Preprocessor/ReplaceTestFiles/PgmReplacePartialTC.cbl
```diff
@@ -0,0 +1,25 @@
...
+000004   02 TO-:PREFIX:-TO.
+000001 REPLACE ==:tag:== BY ==var==.
+000002 01 :tag::tag: PIC X.
+000003 01 S:tag::tag: PIC X.
+000004 01 :tag:M:tag: PIC X.
+000005 01 :tag::tag:E PIC X.
+000006 01 S:tag:M:tag: PIC X.
+000007 01 :tag:M:tag:E PIC X.
+000008 01 S:tag::tag:E PIC X.
+000009 01 S:tag:M:tag:E PIC X.
```
fm-117, line 25: Content is duplicated from PgmReplacePartial.cbl, is it possible to keep a single source file and make a distinction only for result file ?

mayanje: It Seems that no :-)

TypeCobol.Test/Parser/Programs/TypeCobol/QualifiedVarSubsriptAccess.rdz.tcbl
```diff
@@ -0,0 +1,15 @@
...
+       PROGRAM-ID. Pgm.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 StringT.
+         02 DataT      Pic X(10000).
+         02 LengthT    Pic 99999.
+       PROCEDURE DIVISION.
+      *Ok
+             Move Low-Values       To StringT (1 : LengthT)
+      *Ko syntax error
```
smedilol, line 11: Comment is wrong, this syntax is OK in TC.
Can you do the same test in Cobol to check that there is an error ?

smedilol (changes requested):

TypeCobol.Test/Parser/Programs/TypeCobol/QualifiedVarSubsriptAccess.rdz.tcbl
mayanje: The test using -cob option is a CLI test called failcoboloptionparse
Change comment in 3174661

TypeCobol.Test/Parser/Preprocessor/TestReplaceDirective.cs
fm-117: About the previous remark, you could use the same `testName` for the `ProcessReplaceDirectives` call thus using a single input source file, and alter the name for the result comparison depending on the `cobol` parameter. Would it work ?

smedilol (dismissed):

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -2408,13 +2408,13 @@ private bool CheckForPartialCobolWordPattern(int startIndex, out int patternEndI
...
             if (line.Length > index && line[index] == ':')
             {
-                patternEndIndex = index;
-                return true;
-            }
-            else
-            {
-                return false;
+                if (compilerOptions.IsCobolLanguage || !(line.Length > index + 1 && line[index + 1] == ':'))
+                {//For Cobol or if ':' is not followed par ':'. Because in TypeCobol we consider :: to be the qualifier symbol.
```
fm-117, line 2412: typo 'not followed by'

fm-117 (changes requested):

TypeCobol/Compiler/Scanner/Scanner.cs
mayanje: @fm-117  It is exactly what I tried to do, I think we should investigate afterwards , what it does not work, it seems that both test will use the same result file, the first result file will win. It is strange for the moment.

mayanje: Type done in 44acf61

TypeCobol.Test/Parser/Preprocessor/ReplaceTestFiles/PgmReplacePartialTC.cbl
fm-117: Ok I've found what's going on it's because of the cache of imported documents. I've tried this and it works :
- delete PgmReplacePartialTC.cbl
- alter PgmReplacePartialTC.txt first line with "Document path PgmReplacePartial"
- use this code in `CheckReplacePartial`

```csharp
            string testName = "PgmReplacePartial";
            PreprocessorUtils.ReplaceProject.CompilationOptions.IsCobolLanguage = cobol;
            PreprocessorUtils.ReplaceProject.ClearImportedCompilationDocumentsCache();
            string result = PreprocessorUtils.ProcessReplaceDirectives(testName);
            PreprocessorUtils.ReplaceProject.CompilationOptions.IsCobolLanguage = false;
            PreprocessorUtils.CheckWithReplaceResultFile(result, testName + (cobol ? string.Empty : "TC"));
```

TypeCobol/Compiler/Scanner/Scanner.cs
mayanje: Ok clearing cache works fine, done in 0386ae2

rooksdo (approved):

fm-117 (dismissed):

fm-117 (approved):

Issue #1887, opened 2021-02-18
SEARCH ALL use table key and index


rooksdo:
When using `SEARCH ALL` instruction, the `WHEN` clause must use the first index declared for the table when subscripting and also use one of the declared keys (ascending or descending).

**Test**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID.  Pgm.
       ENVIRONMENT DIVISION.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01.
        02 DSRCE-TAB  OCCURS 83
                      ASCENDING KEY IS  DSRCE-KEY-A
                      DESCENDING KEY IS DSRCE-KEY-D
                      INDEXED BY Idx, Idx2.
            05 DSRCE-KEY-A           Pic 99.
            05 DSRCE-KEY-D           Pic 99.
            05 DSRCE-SRCE-VAL        Pic S9(3).
      
       01 WS-KEY                     Pic 99.
       01 WS-KEY2                    Pic 99.
       01 WS-SRCE-VAL                Pic S9(3).
       01 I                          Pic 99.
      
       PROCEDURE DIVISION.
      
           SEARCH ALL DSRCE-TAB
              AT END
                 DISPLAY 'VALUE NOT FOUND'
      *  Ok
              WHEN DSRCE-KEY-A (Idx) = WS-KEY
                 DISPLAY 'VALUE FOUND'
           END-SEARCH
      
           SEARCH ALL DSRCE-TAB
              AT END
                 DISPLAY 'VALUE NOT FOUND'
      *  Ko DSRCE-SRCE-VAL is not a key of the table searched
              WHEN DSRCE-SRCE-VAL (Idx) = WS-KEY
                 DISPLAY 'VALUE FOUND'
           END-SEARCH
      
           SEARCH ALL DSRCE-TAB
              AT END
                 DISPLAY 'VALUE NOT FOUND'
      *  Ok
              WHEN DSRCE-KEY-A (Idx) = WS-KEY
               AND DSRCE-KEY-D (Idx) = WS-KEY2
                 DISPLAY 'VALUE FOUND'
           END-SEARCH
      
           SEARCH ALL DSRCE-TAB
              AT END
                 DISPLAY 'VALUE NOT FOUND'
              WHEN DSRCE-KEY-A (Idx) = WS-KEY
      *  Ko DSRCE-SRCE-VAL is not a key of the table searched
[...]

PR #1931, opened 2021-03-30
#1887 search all check when use first index and one key

delevoye:
This is a remake of #1929 because it had been closed by deleting its original target branch.


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -239,6 +239,83 @@ public override bool Visit(Move move)
...
+                    if (dataDefinitionKey?.Parent.CodeElement is DataDescriptionEntry tableDescriptionEntry)
+                    {
+                        if (tableDescriptionEntry.TableSortingKeys != null)
+                        {
+                            var tableSortingKeyNames = tableDescriptionEntry.TableSortingKeys
+                                .Where(key => key.SortDirection?.Value != SortDirection.None)
+                                .Select(key => key.SortKey.Name);
+                            var isTableSortingKey = tableSortingKeyNames.Contains(dataDefinitionKey.Name);
+                            if (!isTableSortingKey)
+                            {
```
fm-117, line 292: Getting the table and its defined keys could be done only once in the visitor constructor ? By examining the search ?

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -239,6 +239,83 @@ public override bool Visit(Move move)
...
+                // TODO : Handle the case of sub-tables
+                if (variableStorageArea?.Subscripts.Count == 1)
+                {
+                    ////////// WHEN must use one of the declared keys (ascending or descending) //////////
+                    var dataDefinitionKey = _whenSearch.GetDataDefinitionFromStorageAreaDictionary(variableStorageArea);
+                    if (dataDefinitionKey?.Parent.CodeElement is DataDescriptionEntry tableDescriptionEntry)
+                    {
+                        if (tableDescriptionEntry.TableSortingKeys != null)
+                        {
+                            var tableSortingKeyNames = tableDescriptionEntry.TableSortingKeys
```
mayanje, line 287: I think this can be simplified as follow:
```csharp
var existsName = tableDescriptionEntry.TableSortingKeys.SingleOrDefault(key => key.SortDirection?.Value != SortDirection.None && key.SortKey.Name.Equals(dataDefinitionKey.Name));
   if (existsName == null)
   {
   // error
   IsInError = true;
   }

```

rooksdo (dismissed):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: It seems to be complicated to examinate the Search and then checking the rule for the WhenSearch as it implies to store elements from the Seach (and handle the case of imbrication...).
This is because we are not in an Analyzer and the children need to be visited for the StoraAreas to be filled.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: Your suggestion has been used with the `Any()` method from Linq instead of `SingleOrDefault()` as the result is not used.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -239,6 +239,81 @@ public override bool Visit(Move move)
...
+                    if (dataDefinitionKey?.Parent.CodeElement is DataDescriptionEntry tableDescriptionEntryKey)
+                    {
+                        if (tableDescriptionEntryKey.TableSortingKeys != null)
+                        {
+                            var existsName = tableDescriptionEntryKey.TableSortingKeys
+                                .Any(key => key.SortDirection?.Value != SortDirection.None && key.SortKey.Name.Equals(dataDefinitionKey.Name, StringComparison.OrdinalIgnoreCase));
+                            if (!existsName)
+                            {
+                                // error
+                                IsInError = true;
```
fm-117, line 292: `return false` ? to stop visit ?

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -239,6 +239,81 @@ public override bool Visit(Move move)
...
+                    // Ensure the subscript is the 1st declared
+                    var isFirstIndexDeclaredInTable = false;
+                    if (dataDefinitionFirstSubscript?.Parent.CodeElement is DataDescriptionEntry tableDescriptionEntrySubscript)
+                    {
+                        isFirstIndexDeclaredInTable = tableDescriptionEntrySubscript.Indexes[0].Name.Equals(dataDefinitionFirstSubscript.Name, StringComparison.OrdinalIgnoreCase);
+                    }
+                    if (!isFirstIndexDeclaredInTable)
+                    {
+                        // error
+                        IsInError = true;
```
fm-117, line 310: same remark/question

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: As the error is found, there is in fact no need to continue visiting.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -239,6 +239,83 @@ public override bool Visit(Move move)
             return true;
         }
 
+        public override bool Visit(WhenSearch whenSearch)
+        {
+            var search = whenSearch.Parent;
+            System.Diagnostics.Debug.Assert(search is Search);
+            if (search.CodeElement is SearchBinaryStatement)
```
fm-117, line 246: Test on `CodeElement.Type` is more efficient. Sorry for the late comment, should have picked up in my previous review.

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: In that case we specifically want `SearchBinaryStatement` (SEARCH ALL) and not `SearchSerialStatement` (SEARCH VARYING).
The distinction is not available in `CodeElementType`.

fm-117: You're right, the info is stored in `StatementType` property of `StatementElement` but it would still require a type-check-cast to access it so no need to change that.

fm-117 (approved):

rooksdo (approved):

Issue #1919, opened 2021-03-16
Procedure signature overloading fails with LENGTH OF


rooksdo:
### Description

**Parser version : v1.5.1**

Procedure signature overloading fails when using `LENGTH OF` key word.

**Pgm1**

```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm1.
       DATA DIVISION.
       LOCAL-STORAGE SECTION.
       01 txt   pic X(100).
       01 l     pic S9(5) comp-5.
       01 RC   Type ReturnCode.
       PROCEDURE DIVISION.
      *Ok
           Call Pgm2::Foo Input txt
                                l
                         Output RC.
      *Ok
           Call Pgm2::Foo Input txt
                     by content l
                         Output RC.

      *Ko Syntax error : No suitable function signature found for 'Pgm2.Foo'
           Call Pgm2::Foo Input txt
                      length of l
                         Output RC.

      *Ko Syntax error : No suitable function signature found for 'Pgm2.Foo'
           Call Pgm2::Foo Input txt
           by content length of txt
                         Output RC.
      
       END PROGRAM Pgm1.
```
**Pgm2**
```cob
       IDENTIFICATION DIVISION .
       PROGRAM-ID. Pgm2.
       PROCEDURE DIVISION.
      
       DECLARE PROCEDURE Foo PUBLIC
                   INPUT Dat     PIC X(100)
                         Len     pic S9(5) comp-5
                  OUTPUT RC      TYPE ReturnCode
               .
       END-DECLARE.
      
       DECLARE PROCEDURE Foo PUBLIC
                   INPUT Dat     PIC X(100)
                         Len     pic S9(5) comp-5
                         Pos     pic S9(5) comp-5
                  OUTPUT RC      TYPE ReturnCode
               .
       END-DECLARE.
       END PROGRAM Pgm2.
```

Issue #1909, opened 2021-03-08
Exception on empty SELECT


smedilol:
**CHILD OF #1835**

**Describe the bug**
An empty `select` instruction cause a NullReferenceException.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       ENVIRONMENT DIVISION.
       INPUT-OUTPUT SECTION.
       FILE-CONTROL.
       SELECT
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 varX                  PIC XX.
       PROCEDURE DIVISION.
           goback
           .
       END PROGRAM Pgm.
```

**Expected behavior**
No exception.

Issue #1903, opened 2021-03-03
Bug on REPLACE (single to multiple tokens)


fm-117:
**Describe the bug**
Parser fails in some case of REPLACE operations.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
           REPLACE
             ==:StrtPgm:== By
             ==
               Set MyVar1 To True
             ==
           .
       WORKING-STORAGE SECTION.
       01 item PIC X.
          88 MyVar1 VALUE 'A'.
       PROCEDURE DIVISION.
             :StrtPgm:
             .
       END Program Pgm.
```

This results in `Line 14[14,22] <27, Error, Syntax> - Syntax error : no viable alternative at input 'Set ... MyVar1'`.
However, this modified sample works fine (use of `:Strt Pgm:` instead of `:StrtPgm:`).

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
           REPLACE
             ==:Strt Pgm:== By
             ==
               Set MyVar1 To True
             ==
           .
       WORKING-STORAGE SECTION.
       01 item PIC X.
          88 MyVar1 VALUE 'A'.
       PROCEDURE DIVISION.
             :Strt Pgm:
             .
       END Program Pgm.
```

**Expected behavior**
Both samples should compile without any error.

**Technical hints**
The failing REPLACE operation is a `SingleToMultipleTokensReplaceOperation` whereas the working one is a `MultipleTokensReplaceOperation`. Check how the processing of those two differs.

Issue #1900, opened 2021-03-02
Exception on empty exec


smedilol:
**CHILD OF #1835**

**Describe the bug**

**To Reproduce**
Cobol code that cause the bug : (if any)
1. First type this source code:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       PROCEDURE DIVISION.
      
           
      
           goback
           .
       END PROGRAM DVZF0OSM.
```
2. Ensure that there is no error (add a space in a blank line and save document if there is any error)
3. Type `exec` on 2 lines above `goback` 
You should have:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       PROCEDURE DIVISION.
      
           exec
      
           goback
           .
       END PROGRAM DVZF0OSM.
```

4. Then an exception appears:
Runtime error: System.NullReferenceException: Object reference not set to an instance of an object.
   at TypeCobol.Compiler.CodeElements.EnumeratedValue..ctor(Token t, Type enumType)
   at TypeCobol.Compiler.Parser.CobolWordsBuilder.CreateExternalName(ExternalName3Context context, SymbolType symbolType, Type enumType)
   at TypeCobol.Compiler.Parser.CobolWordsBuilder.CreateExecTranslatorName(ExecTranslatorNameContext context)
   at TypeCobol.Compiler.Parser.CobolStatementsBuilder.CreateExecStatement(ExecStatementContext context)
   at TypeCobol.Compiler.Parser.CodeElementBuilder.EnterExecStatement(ExecStatementContext context)
   at TypeCobol.Compiler.Parser.Generated.CodeElementsParser.ExecStatementContext.EnterRule(IParseTreeListener listener)
   at Antlr4.Runtime.Tree.ParseTreeWalker.EnterRule(IParseTreeListener listener, IRuleNode r)
   at Antlr4.Runtime.Tree.ParseTreeWalker.Walk(IParseTreeListener listener, IParseTree t)
   at TypeCobol.Compiler.Parser.CodeElementsParserStep.ParseProcessedTokensLinesChanges(TextSourceInfo textSourceInfo, ISearchableReadOnlyList`1 documentLines, IList`1 processedTokensLinesChanges, PrepareDocumentLineForUpdate prepareDocumentLineForUpdate, TypeCobolOptions compilerOptions, PerfStatsForParserInvocation perfStatsForParserInvocation)

[...]

PR #1936, opened 2021-04-13
v1.5.2

rooksdo:
### Changelog

**Improvements**
#1919 Fix Procedure overloading with LENGTH OF clause
#1844 Detect END PROGRAM not matching program name
#1899 Accept missing space after dot
#1895 Handle COBOL keywords in SPECIAL-NAMES CLASS definition 

**Direct copy parsing**
#1870 Implement direct copy parsing mode
#1892 Integrate direct copy parsing with LSP server
#1918 Introduce -cob option to discard '::' as TypeCobol qualifier

**Compiler Rule**
#1887 Check SEARCH ALL statements use first index and one key

**NRE fixes**
#1900 Avoid NullReferenceException on empty EXEC statement
#1909 Avoid NullReferenceException on incomplete SELECT statement
#1805 Fix NRE when SET xxx UP TO xxx is used instead of UP BY 

**Technical**
#1896 #1903 Fix exceptions and scanning errors on some REPLACE operations
#1920 Compatibility option for outline refresh with LSP clients


fm-117 (approved):
Could not find better for description of #1918

mayanje (approved):

PR #1938, opened 2021-04-20
1935 cfgdfa for analyzers

fm-117:
What this PR change:
- CFG builders are now identified by their `CfgBuildingMode`
- CLI and LanguageServer now support dual CFG building in certain circumstances
  - for CLI, if the user specifies both `-cfg` option and `-zcr` option and selected mode for `-cfg` is not `WithDfa`
  - for LanguageServer, we have two options named `-cfg`
    - the first is command-line only and defines if the user wants to receive CFG notifications (and how, `AsFile` or `AsContent`)
    - the second is the same as CLI and is used to define mode of CFG building
    - if the selected mode is not `Standard`, a second CFG builder with standard mode will be created to allow proper notifications


delevoye (dismissed):

TypeCobol.Analysis.Test/TypeCobol.Analysis.Test.csproj
```diff
@@ -20,6 +20,9 @@
     <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <Private>False</Private>
     </Reference>
+    <Reference Include="Antlr4.Runtime, Version=4.6.0.0, Culture=neutral, PublicKeyToken=09abb75b9ed49849, processorArchitecture=MSIL">
+      <HintPath>..\packages\Antlr4.Runtime.4.6.6\lib\net45\Antlr4.Runtime.dll</HintPath>
```
delevoye, line 24: Is there a problem with the versions? One is 4.6.0.0, the other is 4.6.6.
And there is `net45` in the path, while is it written `targetFramework="net47"` in the file packages.config.

delevoye (changes requested):

TypeCobol.Analysis.Test/TypeCobol.Analysis.Test.csproj
fm-117: No it's fine, actually i've copied the reference from `TypeCobol.Analysis` project file.
- the assembly version is `4.6.0.0` but the NuGet package version is `4.6.6`
- we are targeting .NET Framework 4.7 but Antlr provides assemblies up to `net45` so we are using the most up-to-date available DLL (which is compatible with our own target).

delevoye (approved):

rooksdo (approved):

mayanje (approved):

Issue #1937, opened 2021-04-20
LSP: -ca option for the server


mayanje:
**Description**
The -ca is CLI commande line option, not a LSP server command line option. It should also be a LSP server command line option.

PR #1944, opened 2021-04-21
WI #1937 Add -ca option to LSP server

mayanje:


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -55,6 +55,12 @@ public class Workspace
         public EventHandler<string> WarningTrigger { get; set; }
         public Queue<MessageActionWrapper> MessagesActionsQueue { get; private set; }
         private Func<string, Uri, bool> _Logger;
+        public List<string> CustomAnalyzerFiles { get; set; }
```
fm-117, line 58: The property is not used, you can remove it.

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -250,6 +256,25 @@ public bool TryGetOpenedDocumentContext(Uri fileUri, out DocumentContext openedD
...
+            if (customAnalyzerFiles != null)
+            {
+                try
+                {
+                    _customAnalyzerProviders = customAnalyzerFiles.Select(f => AnalyzerProviderLoader.LoadProvider(f));
+                }
+                catch(Exception e)
+                {
+                    _Logger(e.Message, null);
+                }
```
fm-117, line 274: If we want to log exceptions happening during load,w e have to actually perform the load. LINQ Select is deferred so nothing happens here but on the first iteration over `_customAnalyzerProviders`

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -55,6 +55,12 @@ public class Workspace
         public EventHandler<string> WarningTrigger { get; set; }
         public Queue<MessageActionWrapper> MessagesActionsQueue { get; private set; }
         private Func<string, Uri, bool> _Logger;
+        public List<string> CustomAnalyzerFiles { get; set; }
+        /// <summary>
+        /// Custom Analyzer Providers Loaded
+        /// </summary>
+        private IEnumerable<IAnalyzerProvider> _customAnalyzerProviders;
```
fm-117, line 62: We should store an array of `IAnalyzerProvider`, meaning an already enumerated collection, see my remark below.

fm-117 (changes requested):
So you decided to keep the option as a common option ? I thought we would remove it from `GetCommonTypeCobolOptions` and then duplicate it, one specifically for CLI, the other for LS.

rooksdo (dismissed):

TypeCobol.LanguageServer/Workspace.cs
mayanje: done in 78ea44d

TypeCobol.LanguageServer/Workspace.cs
mayanje: done in 78ea44d

TypeCobol.LanguageServer/Workspace.cs
mayanje: done in 78ea44d

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -250,6 +255,29 @@ public bool TryGetOpenedDocumentContext(Uri fileUri, out DocumentContext openedD
             }
         }
 
+        /// <summary>
+        /// Load Custom Analyzers
+        /// </summary>
+        /// <param name="customAnalyzerFiles"></param>
+        internal void LoadCustomAnalyzers(List<string> customAnalyzerFiles)
+        {
+            if (customAnalyzerFiles != null)
```
fm-117, line 264: You can replace this non-nullity test with an assert as we always instantiate the list in TypeCobolServerHost

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -505,8 +533,13 @@ public void DidChangeConfigurationParams(string[] arguments)
 
             //Configure CFG/DFA analyzer + external analyzers if any
             var compositeAnalyzerProvider = new CompositeAnalyzerProvider();
-            compositeAnalyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer(TypeCobolLanguageServer.lspcfgId, Configuration.CfgBuildingMode));
-            compositeAnalyzerProvider.AddCustomProviders(Configuration.CustomAnalyzerFiles);
+            compositeAnalyzerProvider.AddActivator((o, t) => CfgDfaAnalyzerFactory.CreateCfgAnalyzer(TypeCobolLanguageServer.lspcfgId, Configuration.CfgBuildingMode));            
+            if (this._customAnalyzerProviders != null)
```
fm-117, line 537: Same here, the array of analyzers is non-null, so an Assert is better.

fm-117 (approved):
See my remarks, optional changes if you want.

TypeCobol.LanguageServer/Workspace.cs
mayanje: don in 5ec844a

TypeCobol.LanguageServer/Workspace.cs
mayanje: don in 5ec844a

rooksdo (dismissed):

fm-117 (approved):

rooksdo (approved):

PR #1947, opened 2021-04-22
1939 improve expression trees

### LINKED ISSUE #1939: Expression trees improvements
Class `TypeCobol.Compiler.CodeElements.Expression` has very few methods/properties and is not easy to use from outside of the parser.

The goal of this issue is to improve the expression tree API with basic data/operations.

Suggested improvements :
- add a `NodeData` property (type `object` ?) on `Expression` to summarize the nature of the expression node (i.e. for a operation or relation, this would be the operator, name of the variable for a leaf,...)
- add a `GetChildren` method to get left and right children on any expression
- for relational operators, a "shortcut" is currently used to automatically change 'not greater than' into 'LessThanOrEqualTo'. Although it is semantically correct, we lose the original source code written by the developper and the 'not' token is not recoverable from the expression. Introduce a new `RelationalOperator` class instead of enum to capture all the original information.

delevoye:
Fixes #1939


TypeCobol/Compiler/CodeElements/CobolLanguageLevelVisitor.cs
```diff
@@ -207,6 +207,7 @@ bool IsStopVisitingChildren
         bool Visit([NotNull] ConditionalExpression conditionalExpression);
         bool Visit([NotNull] SignCondition signCondition);
         bool Visit([NotNull] RelationCondition relationCondition);
+        bool Visit(RelationalOperator conditionNameConditionOrSwitchStatusCondition);
```
delevoye, line 210: Nom de variable à changer

fm-117: Add `NotNull` attribute also.

TypeCobol/Compiler/CodeElements/CobolLanguageLevelVisitor.cs
```diff
@@ -829,6 +830,11 @@ public virtual bool Visit(JsonParseStatement jsonParseStatement)
             return true;
         }
 
+        public bool Visit(RelationalOperator conditionNameConditionOrSwitchStatusCondition)
```
fm-117, line 833: The method must be `virtual`.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -14,6 +17,19 @@ public abstract class Expression : IVisitable
 
 		public ExpressionNodeType NodeType { get; private set; }
 
+        /// <summary>
+        /// Returns the characteristic element of a node
+        /// For a operation or relation, this would be the operator
+        /// For a leaf, this would be the name of the variable, ...
+        /// </summary>
+        public abstract object NodeData();
```
fm-117, line 25: Use a property instead of a method, we should assume that retrieving the `NodeData` is not a costly operation.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -121,6 +147,16 @@ public class ClassCondition : ConditionalExpression
 
 		public SyntaxProperty<bool> InvertResult { get; private set; }
 
+        public override object NodeData()
+        {
+            return new object[] { DataItem, CharacterClassNameReference, DataItemContentType, InvertResult };
```
fm-117, line 152: I'm not sure about that, using arrays makes it even harder for the caller to use the property. Maybe we should switch to a `string` property ? We need to discuss this with @mayanje.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +257,90 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
-		NotEqualTo
-	}
+    public class RelationalOperator : IVisitable
+    {
+        public static RelationalOperator EqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.EqualTo, null));
+        public static RelationalOperator GreaterThan = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.GreaterThan, null));
+        public static RelationalOperator GreaterThanOrEqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.GreaterThanOrEqualTo, null));
+        public static RelationalOperator LessThan = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.LessThan, null));
+        public static RelationalOperator LessThanOrEqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.LessThanOrEqualTo, null));
+        public static RelationalOperator NotEqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.NotEqualTo, null));
```
fm-117, line 267: We should not create instances that are not tethered to a real `Token` coming from source.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +257,90 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
 
-	/// <summary>
+        public Token NotToken { get; private set; }
+        public SyntaxProperty<RelationalOperatorSymbol> RelationalOperatorSyntaxProperty { get; private set; }
```
fm-117, line 270: naming: "Operator" should be enough no ?

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +257,90 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
+            RelationalOperatorSyntaxProperty = relationalOperatorSymbol;
+        }
+
+        public RelationalOperator(SyntaxProperty<RelationalOperatorSymbol> relationalOperatorSymbol, Token notToken) : 
+            this(relationalOperatorSymbol)
+        {
+            NotToken = notToken;
+        }
+
+        public bool SemanticEqualTo(RelationalOperator other)
```
fm-117, line 283: naming: "SemanticEquals"

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +257,90 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
+    /// Relational operators and their meanings
+    /// </summary>
+    public enum RelationalOperatorSymbol
+    {
+        EqualTo,
+        GreaterThan,
+        GreaterThanOrEqualTo,
+        LessThan,
+        LessThanOrEqualTo,
+        NotEqualTo
```
fm-117, line 316: `NotEqualTo` should be removed now.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +257,90 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
+        public static RelationalOperator GreaterThanOrEqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.GreaterThanOrEqualTo, null));
+        public static RelationalOperator LessThan = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.LessThan, null));
+        public static RelationalOperator LessThanOrEqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.LessThanOrEqualTo, null));
+        public static RelationalOperator NotEqualTo = new RelationalOperator(new SyntaxProperty<RelationalOperatorSymbol>(RelationalOperatorSymbol.NotEqualTo, null));
 
-	/// <summary>
+        public Token NotToken { get; private set; }
+        public SyntaxProperty<RelationalOperatorSymbol> RelationalOperatorSyntaxProperty { get; private set; }
+
+        public RelationalOperator(SyntaxProperty<RelationalOperatorSymbol> relationalOperatorSymbol)
```
fm-117, line 272: Why this constructor is needed ?

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolExpressionsBuilder.cs
```diff
@@ -815,81 +816,87 @@ private ConditionOperand CreateConditionOperand(CodeElementsParser.ConditionOper
...
-			if(context.strictRelation() != null)
-			{
-				bool invertStrictRelation = false;
-				if (context.NOT() != null)
-				{
-					invertStrictRelation = true;
-				}
+		internal RelationalOperator CreateRelationalOperator(CodeElementsParser.RelationalOperatorContext context)
+        {
+            var notToken = ParseTreeUtils.GetFirstToken(context.NOT());
```
fm-117, line 821: No check on `context.NOT() != null` ?

delevoye: The check is done in `GetFirstToken()`, but it can be done before to avoid errors if the method is modified.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
delevoye: It can be removed, it was here to avoid having to specify the token for NotToken.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
delevoye: Indeed, it should be named Operator here.
The problem was for the parameter in constructor, as `operator` is a C# keyword.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
delevoye: I agree, and found a way to avoid that:
Creating a SemanticEquals() between RelationalOperator and RelationalOperatorSymbol.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
delevoye: This is required to find the opposite of a RelationalOperatorSymbol, and for the SemanticEquals().

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
delevoye: I agree that arrays are not a good solution as it requires to know the NodeType to be able to read it.
It could be easier to use as a string.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
mayanje: So NodeData is not good name, here it's a CodeElement, this seems to be a Label (étiquette in french)

delevoye: This is neither a Node nor a CodeElement, the class Expression inherits from IVisitable only. The name `Label` stays relevant nonetheless.

The true purpose of this method is to return an object that allows to say if the Labels of 2 Expressions are equal. Do you think the property Label has a value by itself or would a method `HasSameLabel(Expression other)` be better?

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
mayanje: I think we could simply call it Data or Items or DataItems not NodeData, and assume that it is always an array of object which is returned

mayanje (changes requested):

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
mayanje: I still thiking that HasSameLabel is better because it's for testing Expression Kinds and for a lead named expression value.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -14,6 +17,19 @@ public abstract class Expression : IVisitable
...
+        /// For a operation or relation, this would be the operator
+        /// For a leaf, this would be the name of the variable, ...
+        /// </summary>
+        public abstract object NodeData();
+
+        /// <summary>
+        /// Returns the children of an expression
+        /// </summary>
+        /// <returns>The children of the expression, ordered as left then right, each can be null if there is no child</returns>
+        public abstract (Expression, Expression) GetChildren();
```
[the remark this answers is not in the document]
fm-117: Yes I agree, that's more expressive

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
fm-117: Ok so after discussing this, we keep this `NotEqualTo` with a comment explaining that it is not tied to a syntax in source but use it for a new `SemanticOperator` property which returns the "normalized" operator integrating the NOT token if any.

TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
```diff
@@ -358,15 +358,15 @@ class StartStatementChecker
...
+                if (!statement.RelationalOperator.SemanticEquals(RelationalOperatorSymbol.EqualTo) &&
+                    !statement.RelationalOperator.SemanticEquals(RelationalOperatorSymbol.GreaterThan) &&
+                    !statement.RelationalOperator.SemanticEquals(RelationalOperatorSymbol.GreaterThanOrEqualTo))
+                    DiagnosticUtils.AddError(statement, "START: Illegal operator " + statement.RelationalOperator,
                         context.relationalOperator());
         }
     }
 
-    class AlterStatementChecker
+    class AlterStatementChecker
```
fm-117, line 369: redundant added whitespace

fm-117 (changes requested):

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
fm-117: `Label` is interesting, it would be typed as `string` and be easier to use from client perspective. I'm still looking for a name that conveys the idea that this "label" is linked to the nature of the expression...

fm-117: What about "Discriminator" or "DiscriminatorLabel" ? The word "discriminator" is used in Hibernate persistence framework for a string property that allows to distinguish between types among a class hierarchy.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +243,105 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
+        public SyntaxProperty<RelationalOperatorSymbol> Operator { get; }
+        
+        public RelationalOperator(SyntaxProperty<RelationalOperatorSymbol> @operator, Token notToken)
+        {
+            Operator = @operator;
+            NotToken = notToken;
+        }
 
-	/// <summary>
+        public bool SemanticEquals(RelationalOperator other)
```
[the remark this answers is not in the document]
delevoye: Replaced by `public RelationalOperatorSymbol SemanticOperator`.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -201,20 +243,81 @@ public override bool AcceptASTVisitor(IASTVisitor astVisitor)
...
+            {
+                if (NotToken == null)
+                {
+                    return Operator.Value;
+                }
+                switch (Operator.Value)
+                {
+                    case RelationalOperatorSymbol.EqualTo:
+                        return RelationalOperatorSymbol.NotEqualTo;
+                    case RelationalOperatorSymbol.NotEqualTo:
```
fm-117, line 276: `Operator.Value` should never be equal to `NotEqualTo` so this label should be removed.

fm-117 (changes requested):

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
delevoye: Ok, to clarify, a comment has been added to the `default` case.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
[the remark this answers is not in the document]
delevoye: The method has been renamed to `GetOperands`.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -14,6 +15,24 @@ public abstract class Expression : IVisitable
...
+        /// Returns true if the the characteristic element of the nodes are equivalent
+        /// For a operation or relation, the element would be the operator
+        /// For a leaf, it would be the name of the variable, ...
+        /// </summary>
+        public bool EquivalentOperator(Expression other)
+        {
+            return NodeType == other.NodeType && IsEquivalentOperator(other);
+        }
+
+        protected abstract bool IsEquivalentOperator(Expression other);
```
fm-117, line 28: Why the suffix 'Operator' ? Not every expression is made of an operator. If you're looking for a verb or verbal phrase maybe `IsEquivalent` for the public method and `CheckEquivalent` for the protected method ?

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -121,13 +151,27 @@ public class ClassCondition : ConditionalExpression
 
 		public SyntaxProperty<bool> InvertResult { get; private set; }
 
+        protected override bool IsEquivalentOperator(Expression other)
+        {
+            var otherClassCondition = (ClassCondition) other;
+            return DataItem.SymbolReference?.Name == otherClassCondition.DataItem.SymbolReference?.Name
+                   && CharacterClassNameReference.Name == otherClassCondition.CharacterClassNameReference.Name
+                   && DataItemContentType.Value == otherClassCondition.DataItemContentType.Value
+                   && InvertResult.Value == otherClassCondition.InvertResult.Value;
```
fm-117, line 160: Making the remark here but concerns every `IsEquivalentOperator` implementations: some null checks seems to be missing. For example why add a null check on `InvertResult` in `ToString` and omit it here ?

fm-117 (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/Expressions.rdzPGM.txt
```diff
@@ -0,0 +1,23 @@
...
+  COLORS:Alphanumeric
+  BLUE:Level88
+  GREEN:Level88
+  TURQUOISE:Level88
+--- Intrinsic
+-- TYPES -------
+  BOOL:BOOL
+  DATE:DATE
+  CURRENCY:CURRENCY
+  STRING:STRING
```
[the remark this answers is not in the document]
delevoye: No result format worked for the expressions, so a test in Misc has been created for this purpose.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -14,6 +16,28 @@ public abstract class Expression : IVisitable
...
 		public ExpressionNodeType NodeType { get; private set; }
 
+        /// <summary>
+        /// Returns true if the the characteristic element of the nodes are equivalent
+        /// For a operation or relation, the element would be the operator
+        /// For a leaf, it would be the name of the variable, ...
+        /// </summary>
+        public bool IsEquivalent(Expression other)
+        {
+            return NodeType == other?.NodeType && CheckEquivalence(other);
```
fm-117, line 26: Actually it's clearer with `if (other == null) return false;`

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -398,6 +571,19 @@ public class NumericVariableOperand: ArithmeticExpression {
 		public IntegerVariable IntegerVariable { get; private set; }
 		public NumericVariable NumericVariable { get; private set; }
 
+        protected override bool CheckEquivalence(Expression other)
+        {
+            Debug.Assert(other is NumericVariableOperand);
+            var otherNumericVariableOperand = (NumericVariableOperand) other;
+            return IntegerVariable?.ToString() == otherNumericVariableOperand.IntegerVariable?.ToString()
```
fm-117, line 578: case-insensitive comparison

fm-117 (changes requested):

TypeCobol.Test/TypeCobol.Test.csproj
```diff
@@ -44,6 +44,9 @@
   </Choose>
   <ItemGroup>
     <Compile Include="**\*.cs" />
+    <None Include="Misc\**\*" Exclude="Misc\**\*.cs" >
```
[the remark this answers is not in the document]
delevoye: I tried to use `<Content>` but, oddly, the file `Expressions-expected.txt` is not being included.

TypeCobol.Test/Misc/TestExpressions.cs
```diff
@@ -0,0 +1,221 @@
...
+            AssertEquivalence(0, 1, firstSign);
+            AssertNotEquivalent(1, 2, firstSign);
+            AssertEquivalence(2, 3, firstSign);
+            AssertNotEquivalent(3, 4, firstSign);
+            AssertNotEquivalent(4, 5, firstSign);
+
+            // Class
+            var firstClass = expressionList.FindIndex(e => e.NodeType == ExpressionNodeType.ClassCondition);
+            AssertEquivalence(0, 1, firstClass);
+            AssertNotEquivalent(0, 2, firstClass);
```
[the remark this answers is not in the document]
delevoye: The goal was to compare `IF Var-txt IS NUMERIC` and `IF Var-txt IS NOT NUMERIC`, so the order of 0 and 1 have been exchanged to keep a consistency in tests.

TypeCobol/Compiler/CodeElements/Expressions/Expression.cs
```diff
@@ -66,6 +91,18 @@ public class LogicalOperation : ConditionalExpression
...
 		public ConditionalExpression RightOperand  { get; private set; }
 
+        protected override bool CheckEquivalence(Expression other)
+        {
+            Debug.Assert(other is LogicalOperation);
+            var otherLogicalOperation = (LogicalOperation)other;
+            return Operator.Value == otherLogicalOperation.Operator.Value;
+        }
+
+        public override (Expression, Expression) GetOperands()
```
mayanje, line 101: So it seems that this method GetOperands is not used now, I am thinking if it can be helpfull?

delevoye: It is planned to be used in external analyzers as it greatly simplifies the code.

mayanje (approved):

rooksdo (approved):

Issue #1946, opened 2021-04-22
Forbid use of EXHIBIT keyword


rooksdo:
### Description
OS/VS COBOL keyword `EXHIBIT` is not supported by IBM compiler. An error should be detected for this instruction specifically in EI environment.

Sample
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. foo.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 WS-FOO PIC X(42).
       01 WS-BAR PIC X(42).
       PROCEDURE DIVISION.
      * Non-Compliant
           EXHIBIT NAMED WS-FOO WS-BAR.
       END PROGRAM foo.
```

**Expected result**
"E-I compilation doesn't allow use of EXHIBIT instruction."

fm-117:
Current diagnostic for this sample is: `Line 9[20,24] <27, Error, Syntax> - Syntax error : no viable alternative at input 'EXHIBIT ... NAMED'`

fm-117:
Closed with PR #1959, a new customized messsage is used for unsupported keywords.

PR #1959, opened 2021-05-11
WI #1946 Improve error message for unsupported EXHIBIT statement

fm-117:


delevoye (dismissed):

delevoye (approved):

mayanje (approved):
Ok, let's stay simple

TypeCobol.Test/Parser/Programs/Cobol85/UnsupportedKeywords.rdzMix.txt
```diff
@@ -0,0 +1,16 @@
...
+       WORKING-STORAGE SECTION.
+       01 WS-FOO PIC X.
+       01 WS-BAR PIC X.
+      *EXHIBIT, NAMED and CHANGED are not reserved keywords
+       01 EXHIBIT PIC X.
+       01 CHANGED PIC X.
+       01 NAMED PIC X.
+       PROCEDURE DIVISION.
+      *EXHIBIT statement is not supported
+Line 13[20,24] <27, Error, Syntax> - Syntax error : 'EXHIBIT' is not supported
```
rooksdo, line 13: Because this is a E-I specific feature, I would prefer to use the following message :
"E-I compilation doesn't allow use of EXHIBIT instruction."

rooksdo: Is this checked only for E-I environment?

rooksdo (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/UnsupportedKeywords.rdzMix.txt
fm-117: No the error will be displayed regardless of the compilation configuration (E-I or regular). It is not a check because the keyword is not known, it is an ANTLR parsing error. In this PR, we adjust the message so as to explain to the end user that the keyword/statement is not supported.
I don't think this should be considered as E-I specific because `EXHIBIT` is not part of IBM cobol specs and never has been supported in our parser. To reject it in E-I environment only, we would have to support it in the first place.

rooksdo: OK it makes sense.

rooksdo (approved):

PR #1963, opened 2021-05-19
WI #1961 Add IsCopy property into TextSourceInfo object

### LINKED ISSUE #1961: Analyzers should know if the current source being scanned is a copy or a program
When implementing an external analyzer, it is required to know the nature of the current source file.
The goal of this issue is to adapt the parser so as to let external anlyzer know whether the file is a copy or a program.

To transfer the nature of the file, a new boolean property `IsCopy` will be included in `TextSourceInfo` and an instance of `TextSourceInfo` will be passed to external analyzers at creation time (methods `CreateSyntaxDrivenAnalyzers` and `CreateQualityAnalyzers` from `IAnalyzerProvider`).

fm-117:
Fixes #1961.

No need to change `IAnalyzerProvider.CreateQualityAnalyzers` signature because a `TextSourceInfo` instance is available in each document passed in `Inspect` methods.

### DIFF
--- a/TypeCobol/Compiler/FileCompiler.cs
+++ b/TypeCobol/Compiler/FileCompiler.cs
@@ -1,8 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Threading;
 using JetBrains.Annotations;
 using TypeCobol.Compiler.CodeModel;
 using TypeCobol.Compiler.Directives;
@@ -10,7 +8,6 @@ using TypeCobol.Compiler.File;
 using TypeCobol.Compiler.Preprocessor;
 using TypeCobol.Compiler.Scanner;
 using TypeCobol.Compiler.Text;
-using TypeCobol.Tools;
 
 namespace TypeCobol.Compiler
 {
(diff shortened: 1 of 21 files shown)


TypeCobol/Compiler/FileCompiler.cs
```diff
@@ -154,41 +158,34 @@ public class FileCompiler
...
             {
-                TextDocument = new ReadOnlyTextDocument(sourceFile?.Name, sourceFile?.Encoding, columnsLayout, sourceFile?.ReadChars());
-            }
-            // 2.b Load it in an existing text document in memory
-            else if (sourceFile != null)
-            {
-                TextDocument = textDocument;
-                textDocument.LoadChars(sourceFile.ReadChars());
+                Debug.Assert(sourceFile != null);
+                Debug.Assert(fileInfo != null);
```
delevoye, line 162: `fileInfo` can be given with the value `null` to the function.
It would mean `sourceFile` will be null too.

delevoye (changes requested):

TypeCobol/Compiler/FileCompiler.cs
fm-117: `fileInfo` can be `null` but in that case that would mean we have a non-null `textDocument`. Both `fileInfo` and `textDocument` describe the input to be parsed and one of them should be not null but not both. I will add a `Debug.Assert(textDocument == null)` in the first `if` to emphasize this. As for `sourceFile`, the assertion here is merely to avoid ReSharper warning because if we have a non-null `fileInfo` then it means we performed a search in `SourceFileProvider` and we found something otherwise we would have been in exception.

fm-117: See 48f2dc71c62515b29de1ede469594cd9623a410c.

delevoye: OK, this is easier to read this way.

delevoye (approved):

mayanje (approved):

rooksdo (approved):

Issue #1975, opened 2021-06-04
LSP Add to missingCopies params CPY Copies


mayanje:
The LSP Client need to know if a missing copy is a CPY copy or not, so we need to add a new property in MissingCopiesParam class as a list of missing CPY Copy names.

fm-117:
Done with #1979.

PR #1979, opened 2021-06-07
WI #1975 Add cpyCopies field

mayanje:
**What have been done**
A new field cpyCopies has been added in the class MissingCopiesParams, this field contains all mising CPY Copies.


TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/Missing Copies Notification/MissingCopiesParams.cs
```diff
@@ -15,5 +15,9 @@ class MissingCopiesParams
         /// List of missing copies
         /// </summary>
         public List<string> Copies { get; set; }
+        /// <summary>
+        /// List of missing CPY copies
+        /// </summary>
+        public List<string> cpyCopies { get; set; }
```
fm-117, line 21: `CpyCopies` would be more consistent with naming style.

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -171,7 +171,13 @@ protected virtual void OnDidReceiveExtractUseCopies(ExtractUseCopiesParams param
                 {
                     var missingCopiesParam = new MissingCopiesParams();
                     missingCopiesParam.textDocument = parameter.textDocument;
+#if EUROINFO_RULES
+                    missingCopiesParam.Copies = copiesName.Where(s => !base.Workspace.CompilationProject.CompilationOptions.HasCpyCopy(s)).ToList();
+                    missingCopiesParam.cpyCopies = copiesName.Where(s => base.Workspace.CompilationProject.CompilationOptions.HasCpyCopy(s)).ToList();
+#else                    
                     missingCopiesParam.Copies = copiesName;
+                    missingCopiesParam.cpyCopies = new List<string>();
+#endif
```
fm-117, line 180: Can you factorize this into a single method to avoid duplication with `TypeCobolServer` ?
Some other remarks:
- you can use LINQ `ToLookup` method to split copies names into 2 groups in a single call, instead of iterating the list twice
- no need to use the `base` keyword as the `Workspace` property is not redefined

fm-117 (changes requested):

fm-117:
Also we could use this PR to remove the client to server notification as it is not used, but it requires updates in clients.
See:
- `Workspace.UpdateMissingCopies`
- `TypeCobolCustomLanguageServer.OnDidReceiveMissingCopies`
- `TypeCobolCustomLanguageServer.CallReceiveMissingCopies`

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
mayanje: done in 1cf225e

fm-117 (approved):

rooksdo (approved):

fm-117 (approved):

Issue #1985, opened 2021-06-18
Add Options -cpyr to report all Used COPY names


mayanje:
**Description**
It is usefull to know all copy names used by a program along with their associated suffixed copy names

mayanje:
Closed by PR #1986

PR #1986, opened 2021-06-18
WI #1985 Add -cpyr option

mayanje:
**What have been none:**

The -cpyr option has been added in EI mode
All used copy names along with their suffixed names are collected an reported in the specified file in a csv format:

comy_name;sufixed_name_1; suffixed_naame_2; ....; suffixed_name_n 

The TypeCobolOption class is used to stored a dictionary of set of names.


TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -99,5 +108,54 @@ public TypeCobolOptions Clone()
...
+                            {
+                                sw.Write(';');
+                                sw.Write(suffixed);
+                            }
+                            sw.WriteLine();
+                        }
+                    }
+                }
+            }
+        }
```
fm-117, line 158: This method should be moved into CLI as it is more focused on file manipulation. Also it would avoid the duplicate nullity check on the file path.

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
```diff
@@ -475,6 +475,14 @@ internal static IList<DocumentChange<IProcessedTokensLine>> ProcessTokensLinesCh
 
                 TUVienna.CS_CUP.Runtime.Symbol ppSymbol = directivesParser.parse();
 
+#if EUROINFO_RULES
+                if (document.CompilerOptions.ReportUsedCopyNamesPath != null && 
+                    document.CompilerOptions.ExecToStep <= ExecutionStep.Preprocessor) 
+                {
+                    continue;
+                }
+#endif
```
fm-117, line 484: I guess this is for performance reasons but this is ambiguous. When the option is set, we will end up with an incomplete PreprocessorStep and results in `CompilationDocument` will be partial.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -13,6 +13,14 @@ public class TypeCobolOptions : IBMCompilerOptions, ITypeCobolCheckOptions
...
 #if EUROINFO_RULES
         public bool AutoRemarksEnable { get; set; }
+        /// <summary>
+        /// Path to generrate collected used copy names
+        /// </summary>
+        public string ReportUsedCopyNamesPath { get; private set; }
+        /// <summary>
+        /// Collected CopyNames if -cpyr report is selected.
+        /// </summary>
+        public System.Collections.Generic.Dictionary<string, System.Collections.Generic.HashSet<string>> CollectedCopyNames;
```
fm-117, line 23: I don' think this should be held in the options class. Look at `CLI` class, I think we already have everything we need, we already collect used copies, the format is different here because we also want to list the variations for each.

Currently we store used copies in a `HashSet<string>` in `CLI`, so let's switch to a `Dictionary<string, HashSet<string>>` like you suggested and depending on the option `-exc` or `-cpyr` we change the format of the output file. The method `WriteCopiesFile` would have to be adapted.

CLI/src/CLI.cs
```diff
@@ -524,6 +528,11 @@ private ReturnCode AddErrorsAndComputeReturnCode()
             CheckExternalDiagnostics(_dependenciesDiagnostics);
             CheckExternalDiagnostics(_intrinsicsDiagnostics);
 
+#if EUROINFO_RULES
+            if (this._configuration.ReportUsedCopyNamesPath != null && _configuration.ExecToStep <= ExecutionStep.Preprocessor)
+                return returnCode;
+#endif
```
fm-117, line 534: I don't see why we need to do that.

fm-117 (changes requested):

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
mayanje: Yes it's reaaly for performance reasons specific to this option.

CLI/src/CLI.cs
mayanje: To avoid return ReturnCode.MissingCopy

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: Yes it's true but the purpose here is to be able to collect information out of CLI class, using only TypeCobol.dll

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: Yes it's true but the purpose here is to be able to collect information out of CLI class, using only TypeCobol.dll

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
fm-117: So you're ok with the fact that the resulting `CompilationDocument` instance will end up incomplete when using this option ?

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
fm-117: Same remark as above, we should be able to rely only on `CopyTextNamesVariations` without having to create a new data storage. Besides `TypeCobolOptions` should not be concerned with data collection.

CLI/src/CLI.cs
fm-117: Ok so this logic should also apply to `-exc` option.

What about something like:

```csharp
if (_missingCopies.Count > 0 && _configuration.ExtractedCopiesFilePath != null && _configuration.ReportUsedCopyNamesPath != null)
{
    returnCode = ReturnCode.MissingCopy;
}
```

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
mayanje: Yes, but another condition is that ExecToStep <=  ExecutionStep.Preprocessor

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: Ok I understand

mayanje: done in 
016fda9

CLI/src/CLI.cs
mayanje: the thing is that -cpyr options is only an EI option, and in this situation the MissingCopy is not considered as an error, this options is intended to be called even if the collected copy names are missing, there is also another condition _configuration.ExecToStep <= ExecutionStep.Preprocessor

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
fm-117: Could you add a comment to explain the optimization ?

TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -778,5 +785,55 @@ public virtual IList<Diagnostic> AllDiagnostics()
...
+                string copyName = copy.TextName.ToUpper();
+                if (!CollectedCopyNames.ContainsKey(copyName))
+                {
+                    CollectedCopyNames[copyName] = new System.Collections.Generic.HashSet<string>();
+                }
+                if (copy.Suffix != null)
+                {
+                    System.Collections.Generic.HashSet<string> suffixedNames = CollectedCopyNames[copyName];
+                    suffixedNames.Add(copyName + copy.Suffix.ToUpper());
+                }
```
fm-117, line 806: Simplify with `TryGetValue`

fm-117 (changes requested):

fm-117 (approved):

mayanje:
Ok, refactoring made is correct.

fm-117 (approved):

PR #1994, opened 2021-06-29
WI #1926 Correct duplicate diagnostics on CALL using special register by reference

### LINKED ISSUE #1926: Missing diagnostic on CALLs using special register
**Describe the bug**
Our parser currently allows use of `LENGTH OF` in CALLs without specifying passing style `BY CONTENT`. This is wrong in Cobol and should be signaled with a diagnostic (error level).

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm1.
       DATA DIVISION.
       LOCAL-STORAGE SECTION.
       01 txt   pic X(100).
       01 l     pic S9(5) comp-5.
       01 RC    pic 99.
       PROCEDURE DIVISION.
      *KO cannot use LENGTH OF by reference
           Call "SOMEPGM" USING txt
                   by reference length of txt
                                RC.
      
      *KO cannot use LENGTH OF by reference
           Call "SOMEPGM" USING txt
                                length of txt
                                RC.
      
      *Ok
           Call "SOMEPGM" USING txt
                   by content   length of txt
                   by reference RC.
      
       END PROGRAM Pgm1.
```

**Expected behavior**
Diagnostics on wrong lines with meaningful error message.

**To discuss**
- expected behavior for TypeCobol procedure calls
  - same as Cobol ?
  - allow the developer to omit the `BY CONTENT` but auto-generates it ?
- What about `ADDRESS OF` ?

delevoye:
Fix #1926
Diagnostic on LENGTH OF by reference was already there, but another problem appeared:
Parameters were checked multiple times, generating duplicated diagnostics.


TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
```diff
@@ -201,53 +201,52 @@ public static void OnCodeElement(CallStatement statement, CodeElementsParser.Cal
                 if (context.programNameOrProgramEntryOrProcedurePointerOrFunctionPointerVariable() == null)
                     DiagnosticUtils.AddError(statement, "Empty CALL is not authorized", context.Start);
 
-                foreach (var call in context.callUsingParameters()) CheckCallUsings(statement, call);
+                // Check each parameter of the CALL
+                foreach (var inputParameter in statement.InputParameters) CheckCallUsings(statement, inputParameter);
```
fm-117, line 205: Ok the foreach should go into the method itself as it is called `CheckCallUsings`.

TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
```diff
@@ -201,53 +201,52 @@ public static void OnCodeElement(CallStatement statement, CodeElementsParser.Cal
...
+        private static void CheckCallUsings(CallStatement statement, CallSiteParameter input)
         {
-            foreach (var input in statement.InputParameters)
-            {
-                // TODO#249 these checks should be done during semantic phase, after symbol type resolution
-                // TODO#249 if input is a file name AND input.SendingMode.Value == SendingMode.ByContent OR ByValue
-                //	DiagnosticUtils.AddError(statement, "CALL .. USING: <filename> only allowed in BY REFERENCE phrase", context);
-                bool isFunctionCallResult = input.StorageAreaOrValue != null &&
-                                            input.StorageAreaOrValue.StorageArea is FunctionCallResult;
+            var errorPosition = input.StorageAreaOrValue?.MainSymbolReference?.NameLiteral.Token;
```
fm-117, line 214: This means that the error will always be positionned on the argument name, what about ommited parameters ? What about messages that concern the whole arrgument including its sharing mode ?
The current build fails, maybe it is because you changed the error positionning. I think we need to find a better strategy for this part.

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
delevoye: See https://github.com/TypeCobolTeam/TypeCobol/pull/1994/commits/6eaf2db4f3c76fb1dc27fecd27c5fb750c99ea19 .

TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
delevoye: For the position of the error, if you give it a context, it will take the first token of this context (so it positions the error on the whole argument).
I think that on ommited parameters, the `AddError` method using a context had a `null` returned for first token, as it has now from `errorPosition`.
Looking at the cases in the function, it seems the new position is only clearer every time. As a developper, I would preferer having the error on the parameter rather than on the "BY".

fm-117: To achieve positioning on a whole arg like `by reference length of txt`, we would have to use the constructor of `Position` class, specifying manually start and end so if @rooksdo is ok with positioning only on arg name, it's good for me too.
I'm not sure for ommitted though because, in previous code we would use the first token of the context so that should give the `OMMITTED` keyword and not `null`, could you please check that ?

delevoye: Previously, the diagnostic from `BY VALUE OMITTED` was placed on the `BY`, even on the case where OMITTED was the second parameter by value (`BY VALUE toto OMITTED`).
Now the error is placed on the token `OMITTED`. (See Call.ko.txt)

TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
```diff
@@ -201,52 +201,58 @@ public static void OnCodeElement(CallStatement statement, CodeElementsParser.Cal
...
                             DiagnosticUtils.AddError(statement,
-                                "CALL .. USING: Illegal <literal> in BY REFERENCE phrase", context);
+                                "CALL .. USING: Illegal <literal> in BY REFERENCE phrase", errorPosition);
                     }
 
                     //BY VALUE
-                    if (input.IsOmitted && input.SharingMode.Value == ParameterSharingMode.ByValue)
-                        DiagnosticUtils.AddError(statement, "CALL .. USING: Illegal OMITTED in BY VALUE phrase", context);
+                    if (inputParameter.IsOmitted && inputParameter.SharingMode.Value == ParameterSharingMode.ByValue)
+                        DiagnosticUtils.AddError(statement, "CALL .. USING: Illegal OMITTED in BY VALUE phrase", inputParameter.Omitted?.Token);
```
fm-117, line 255: Omitted is not null

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CodeElementCheckers.cs
delevoye: Nullity test has been removed and a comment has been added to emphasize the difference of the position for the diagnostic.

fm-117 (approved):

mayanje (approved):

TypeCobol.Test/Parser/CodeElements/CALL.ko.txt
```diff
@@ -7,8 +7,10 @@ Line 7[8,8] <27, Error, Syntax> - Syntax error : extraneous input 'g' expecting
...
 Line 11[17,21] <27, Error, Syntax> - Syntax error : missing {CONTENT, REFERENCE, VALUE} at 'MALUS' RuleStack=codeElement>callStatement>cobolCallStatement>callUsingParameters,  OffendingSymbol=[17,21:MALUS]<UserDefinedWord>
-Line 13[14,21] <27, Error, Syntax> - Syntax error : CALL .. USING: Illegal function identifier RuleStack=codeElement>callStatement>cobolCallStatement>callUsingParameters,  OffendingSymbol=[14,21:FUNCTION]<FUNCTION>
-Line 14[14,19] <27, Error, Syntax> - Syntax error : CALL .. USING: Illegal LENGTH OF in BY REFERENCE phrase RuleStack=codeElement>callStatement>cobolCallStatement>callUsingParameters,  OffendingSymbol=[14,19:LENGTH]<LENGTH>
+Line 13[23,28] <27, Error, Syntax> - Syntax error : CALL .. USING: Illegal function identifier OffendingSymbol=[23,28:RANDOM]<IntrinsicFunctionName>
+Line 14[14,19] <27, Error, Syntax> - Syntax error : CALL .. USING: Illegal LENGTH OF in BY REFERENCE phrase OffendingSymbol=[14,19:LENGTH]<LENGTH>
```
rooksdo, line 11: Why is mentionned `in BY REFERENCE` for instruction on line 14?

delevoye: It is because the default way of passing data is BY REFERENCE, and when specified, it can be BY REFERENCE, BY VALUE, or BY CONTENT.

rooksdo (approved):

PR #1998, opened 2021-07-02
WI #1997 Always create code element for ACCEPT statement

### LINKED ISSUE #1997: Allow empty ACCEPT statement in grammar
Child of #1835.

Modify the ANTLR grammar to allow `ACCEPT` as a stand alone keyword. The code element need to be created and then checked manually.

fm-117:
Fixes #1997.


TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -3843,13 +3843,13 @@ sentenceEnd:
...
 
-acceptDataTransfer:
-	ACCEPT alphanumericStorageArea (FROM mnemonicForEnvironmentNameReferenceOrEnvironmentName)?;
+fromEnvironment:
+	FROM mnemonicForEnvironmentNameReferenceOrEnvironmentName;
 
-acceptSystemDateTime:
-	ACCEPT alphanumericStorageArea FROM ((DATE yyyyMmDd?) | (DAY yyyyDdd?) | DAY_OF_WEEK | TIME);
+fromSystemDateTime:
+	FROM ((DATE yyyyMmDd?) | (DAY yyyyDdd?) | DAY_OF_WEEK | TIME);
```
delevoye, line 3852: Is it possible to have the message coming from ANTLR?
As you said: maybe by keeping the initial grammar but changing the builder only.

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolStatementsBuilder.cs
```diff
@@ -52,20 +52,19 @@ public CobolStatementsBuilder(CobolWordsBuilder cobolWordsBuilder, CobolExpressi
 		 // ACCEPT STATEMENT //
 		//////////////////////
 
-		internal AcceptFromInputDeviceStatement CreateAcceptDataTransferStatement(CodeElementsParser.AcceptDataTransferContext context) {
+		internal AcceptFromInputDeviceStatement CreateAcceptDataTransferStatement(ReceivingStorageArea receivingStorageArea, CodeElementsParser.FromEnvironmentContext context) {
```
delevoye, line 55: Is the name of the function still senseful?

delevoye (changes requested):

TypeCobol/AntlrGrammar/CobolCodeElements.g4
fm-117: Using the original grammar with a modified builder does not seem to work. We end up with unwanted calls to `EnterAcceptStatement`, themselves resulting into `AcceptStatement` code elements with no consumed tokens. Although I don't have the explanation for those extra calls, it seems that the modified grammar is more reliable, producing the correct number of code elements.

I've changed it using your suggestion on the syntax for the optional fragement.

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolStatementsBuilder.cs
fm-117: As it still produces the instance of the code element, i's say yes 😄.

TypeCobol/AntlrGrammar/CobolCodeElements.g4
delevoye: If the last solution works, it should be fine as it stays understandable.

delevoye (approved):

mayanje (approved):

rooksdo (approved):

PR #2006, opened 2021-07-16
WI #2005 Convert ClassCondition.DataItem to ConditionalOperand

### LINKED ISSUE #2005: Convert DataItem to an Expression in ClassCondition
The field `DataItem` of `ClassCondition` is currently a `StorageArea` but it would mean that it should be compared in `CheckEquivalence()`, which is not the spirit of this method as it compares the operator of the expression.
A solution is to have `DataItem` as a `ConditionalOperand`, which is used to hold operands of `ConditionalExpression` (the parent class of `ClassCondition`).

As a result, we can have Exp0 "equivalent" to Exp1. But still have a mean to differentiate them
```cobol
* Exp0
Var-txt2 IS NUMERIC
* Exp1
Var-txt IS NUMERIC
```

delevoye:
Fix #2005


TypeCobol/Compiler/Parser/CodeElementBuilder/CobolExpressionsBuilder.cs
```diff
@@ -703,15 +705,16 @@ internal ConditionalExpression CreateClassCondition(CodeElementsParser.ClassCond
...
+					new ConditionOperand(new Variable(CreateIdentifier(context.identifier()))),
 					dataItemContentType,
 					invertResult);
 			}
 
             // Collect storage area read/writes at the code element level
             if (classCondition.DataItem != null)
             {
-                this.storageAreaReads.Add(classCondition.DataItem);
+                Debug.Assert(classCondition.DataItem.Variable?.StorageArea != null); // This is how we created it above
```
fm-117, line 716: It's not a new `Debug.Assert` but instead the condition must be updated. In fact you should factorize the storage area creation outside of the main `if`. From there you'll have a variable to test in order to update the dictionary. Easier.

fm-117 (changes requested):

mayanje (approved):

fm-117 (approved):

Issue #2009, opened 2021-07-20
Various unexpected exceptions in RDz Plugin LSR mode


mayanje:
**Purpose**

The purpose of this issue is to track any unexpected exception (Like NullReferenceException) that can be raised while working in LSR mode with the RDz plugin.

![image](https://user-images.githubusercontent.com/19149269/126347462-f9546ca1-61e3-4d41-b44b-f90e2fefa306.png)

fm-117:
On line https://github.com/TypeCobolTeam/TypeCobol/blob/1a4d2a4a8334cece75e995795b43160e7ddfeeaf/TypeCobol.LanguageServer/Workspace.cs#L738
the only item that could be `null` is `DiagnosticsEvent` event handler so it's easy to fix, turn the call into `DiagnosticsEvent?.Invoke(...)` just like in line 743.

mayanje:
Fixed in  #2010

PR #2010, opened 2021-07-21
WI #2009 Bad assign DiagnosticsDetected elsewhere

mayanje:
**What have been done**

The DiagnosticsEvent was assigned on document opened and document closed. It should not be the case.


fm-117 (changes requested):
I think we should also unsubscribe from `LoadingIssueEvent`, `ExceptionTriggered` and `WarningTrigger` in the `OnShutdown` (so as it mirrors the `OnInitialize` subscriptions).
Also can you turn these into real `event` fields instead of delegates ? This would clarify the intended usage.
- `DiagnosticsEvent`
- `DocumentModifiedEvent`
- `ExceptionTriggered`
- `LoadingIssueEvent`
- `MissingCopiesEvent`
- `WarningTrigger`

fm-117 (approved):

delevoye (approved):

Issue #2014, opened 2021-07-23
CFG Implement Dominator


mayanje:
**Description**
The purpose of this issue is to add dominator concept to a CFG graph:
https://en.wikipedia.org/wiki/Dominator_(graph_theory)
Dominator algorithms will be used for implementing, quality check analysis.

PR #2015, opened 2021-07-23
WI #2014 Add basic dominator algorithms

mayanje:
**What has been done**
Basic dominator algorithms have been added.

**Possible Improvement**
Later use faster dominator algorithm, but much more complicate to develop and to understand.


TypeCobol.Analysis/Graph/ControlFlowGraph.Dominator.cs
```diff
@@ -0,0 +1,127 @@
...
+            BitSet workSet = new BitSet(dominators.Length); //workingSet
+            do
+            {
+                change = false;
+                foreach (int n in blocks)
+                {
+                    if (n == this.RootBlock.Index)
+                        continue;
+                    BasicBlock<N, D> b = this.AllBlocks[n];
+                    if (b.PredecessorEdges == null)
```
fm-117, line 47: Isn't it a `Debug.Assert` instead ? We ensure that predecessor edges are set with previous call to `SetupPredecessorEdgesFromRoot`.

TypeCobol.Analysis/Graph/ControlFlowGraph.Dominator.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="dominators">The set of Dominators</param>
+        /// <param name="writer"></param>
+        public void DumpDominators(BitSet[] dominators, System.IO.TextWriter writer)
+        {
+            for (int i = 0; i < dominators.Length; i++)
+            {
+                BitSet doms = dominators[i];
+                if (doms != null)
+                {
+                    BasicBlock<N, D> b = this.AllBlocks[i];
```
fm-117, line 121: Is `b.Index` not always equal to `i` ? Intuitively I thought that this method should be static because it does not depend on the current CFG (except if we include other info from the block).
Also why the double braces ? Is it a common notation ?

fm-117 (changes requested):

TypeCobol.Analysis/Graph/ControlFlowGraph.Dominator.cs
mayanje: Yes done in 
d9ea371

TypeCobol.Analysis/Graph/ControlFlowGraph.Dominator.cs
mayanje: done in 
ff21062

TypeCobol.Analysis/Graph/ControlFlowGraph.Dominator.cs
mayanje: and 
9dcd9e1 double bracket removed

fm-117 (approved):

PR #2017, opened 2021-07-29
WI #2012 Add diagnostic on PIC X to Numeric with usage

### LINKED ISSUE #2012: Add a Diagnostic when affecting PIC X/Group to numeric with USAGE
When moving a PIC X/Group to an numeric with an USAGE (COMP/BINARY...), the result is unpredictable and might lead to an ABEND at runtime when using the numeric variable.
To prevent such a situation, a warning should be issued.

**Example:**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Pgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 alpha        pic X(03).
       01 alpha2       pic X(03).

       01 num1         pic 9(03).
       01 num2         comp-1.
       01 num3         comp-2.
       01 num4         pic 9(03) comp-3.
       01 num5         pic 9(03) comp-4.
       01 num6         pic 9(03) comp-5.
       01 num7         pic 9(03) packed-decimal.
       01 num8         pic 9(03) binary.
       01 num9         pic 9(03) comp.

       PROCEDURE DIVISION.

      *   Ok
           move alpha          to alpha2

      *   Ok because num1 is an extended numeric
           move alpha          to num1
      *   Ko because it is a comp variable
           move alpha          to num2
      *   Ko because it is a comp variable
           move alpha          to num3
      *   Ko because it is a comp variable
           move alpha          to num4
      *   Ko because it is a comp variable
           move alpha          to num5
      *   Ko because it is a comp variable
           move alpha          to num6
      *   Ko because it is a comp variable
           move alpha          to num7
      *   Ko because it is a comp variable
           move alpha          to num8
      *   Ko because it is a comp variable
           move alpha          to num9

      *   Ok because it is a function
           move FUNCTION WHEN-COMPILED          to num4
      *   Ok because it is a function
           move FUNCTION CURRENT-DATE           to num4
      *   Ok because it is a function
           move LENGTH                   alpha  to num4
      *   Ok because it is a function
           move LENGTH OF                alpha  to num4

           GOBACK
           .

       END PROGRAM Pgm.
```

delevoye:
Fix #2012


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -214,15 +214,37 @@ public override bool Visit(Move move)
...
+                    if (senderIsAlphanumeric
+                        && receiver is DataOrConditionStorageArea dataOrConditionStorageArea
+                        && move.StorageAreaWritesDataDefinition != null
+                        && move.StorageAreaWritesDataDefinition.TryGetValue(dataOrConditionStorageArea, out var receiverDataDefinition))
+                    {
+                        if (receiverDataDefinition.DataType == DataType.Numeric || receiverDataDefinition.DataType == DataType.NumericEdited)
+                        {
+                            if (receiverDataDefinition.Usage != null && receiverDataDefinition.Usage != DataUsage.None)
+                            {
+                                DiagnosticUtils.AddError(move, $"Moving alphanumeric '{senderDataDefinition.Name}' to numeric '{receiverDataDefinition.Name}' declared with an USAGE is dangerous as using the numeric may lead to unexpected results.", code: MessageCode.Warning);
```
fm-117, line 241: Message could be shortened: `$"Moving alphanumeric '{senderDataDefinition.Name}' to numeric '{receiverDataDefinition.Name}' declared with an USAGE may lead to unexpected results."`

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -214,15 +214,37 @@ public override bool Visit(Move move)
                 {
                     return true;
                 }
+
+                var senderIsAlphanumeric = false;
+                DataDefinition senderDataDefinition = null;
+                if (moveSimple.SendingVariable?.StorageArea?.Kind == StorageAreaKind.DataOrCondition
```
fm-117, line 220: Is testing of storage area kind required ? I mean if we can find it in the dictionary we can go on ? See also equivalent remark for line 233.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -214,15 +214,37 @@ public override bool Visit(Move move)
...
+                }
+
                 foreach (var area in moveSimple.StorageAreaWrites)
                 {
                     var receiver = area.StorageArea;
                     if (receiver is FunctionCallResult)
                         DiagnosticUtils.AddError(move, "MOVE: illegal <function call> after TO");
+
+                    if (senderIsAlphanumeric
+                        && receiver is DataOrConditionStorageArea dataOrConditionStorageArea
```
fm-117, line 233: Why the cast ? If testing is required, you can check the kind just like the sender.

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: The test on kind avoids having `FunctionCallResults`, which we do not want to check for now.
You are right when saying it is similar with the cast line 233.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: It was done like this above, but the cast does not seem to be required in both cases, changing it to test on kind.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
delevoye: That's indeed shorter while keeping the clarity.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -226,19 +226,21 @@ public override bool Visit(Move move)
                 foreach (var area in moveSimple.StorageAreaWrites)
                 {
                     var receiver = area.StorageArea;
-                    if (receiver is FunctionCallResult)
+                    if (receiver == null) continue;
+
+                    if (receiver.Kind == StorageAreaKind.FunctionCallResult)
                         DiagnosticUtils.AddError(move, "MOVE: illegal <function call> after TO");
 
                     if (senderIsAlphanumeric
```
fm-117, line 234: This is now an `else if` as the kinds are mutually exclusive.

fm-117 (dismissed):

fm-117 (approved):

mayanje (approved):

Issue #2027, opened 2021-08-13
Data retention in LS semantic updater timer


fm-117:
See _Workspace.cs_, line 335:

https://github.com/TypeCobolTeam/TypeCobol/blob/83122902e5a5a58f85d83d716523263683e3b6dd/TypeCobol.LanguageServer/Workspace.cs#L335

The lambda expression used for the `Elapsed` event creates a closure that will capture one instance of `FileCompiler`. According to memory profiler, this instance is never collected even after document has been closed.

The goal of this issue is to rewrite the semantic updater mechanism so as to ensure that the `FileCompiler` instance is properly collected.

Issue #2026, opened 2021-08-13
Remove static Token EndOfFile


fm-117:
In _Token.cs_ at line 417:

https://github.com/TypeCobolTeam/TypeCobol/blob/83122902e5a5a58f85d83d716523263683e3b6dd/TypeCobol/Compiler/Scanner/Token.cs#L417

A `Token` instance should not be shared amongst several documents, each source file must get its own end-of-file token.
Besides, depending on how the token is used, it may retain large amount of data through its `TokenSource` and this could cause memory leaks.
- Track every usage of the shared token and update accordingly
  - compare to `TokenType.EndOfFile` instead of comparing to the instance
  - create a new instance when needed

PR #2030, opened 2021-08-16
WI #2026 #2027 Avoid FileCompiler leaks in LanguageServer

fm-117:
- do not capture `FileCompiler` instance but `Uri` instead
- ensure `FileCompiler` is removed from all collections when document is closed
- ensure correct instance is released when clsoing and opening the same document

- remove static token `END_OF_FILE`


delevoye (approved):

mayanje (approved):

fm-117:
@delevoye Please squash merge this PR after build is ok, thanks.

Issue #1955, opened 2021-04-30
Remove ISyntaxDrivenAnalyzer concept


fm-117:
Child issue of #1783, already described in #1935 but copied here.

**Remove the notion of `ISyntaxDrivenAnalyzer` and integrate CFG/DFA building as part of the `SemanticCheck` step**

The concept of `ISyntaxDrivenAnalyzer` has been introduced to support CFG/DFA builders but it does not really match with the concept of analyzer as it is used now. CFG builders are currently tied to the Node phase (`ProduceTemporarySemanticDocument`) but as an external analyzer implementor viewpoint, basing an analyzer on the node buidling process ha no real benefit. It is not incremental and it allows to access less information than in the final QualityCheck step.

The idea here is to focus on Quality analyzers only and consider the CFG/DFA as a normal byproduct of the Node phase. The `TemporarySemanticDocument` would expose a new property/method giving access to the graphs produced during the step (depending on the CFG building mode selected).

Quality analyzers would still be considered as independent from each other.
The `IAnalyzer` interface would be merged with `IQualityAnalyzer` and the `Identifier` property and `GetResult` method would be deleted as they do not really make sense for Quality analyzers.

This solution has great impact on the architecture:
- the dependency from `TypeCobol.Analysis` to `TypeCobol` is reverted and consequently the project `TypeCobol.Analysis` would be merged into `TypeCobol`
- code analysis would be performed only on fully built syntax tree and not during build. This is not a limitation as all the information available during built is still present after cross-check.

PR #2034, opened 2021-08-20
WI #1847 secure external analyzers

### LINKED ISSUE #1847: Improve parser security towards external analyzers
Following #1825.

The goal of this issue is to add defense mechanism (i.e. `try-catch` blocks) against exceptions thrown by external code analyzers.
While some steps are safe, not every external call is secured, check exceptions:
- at dynamic load
- when creating SyntaxDrivenAnalyzers / ASTAnalyzers
- during code analysis itself
  - Node phase
  - QualityCheck step

delevoye:
Closes #1847 

This PR does not handle the case of exceptions occuring in the SyntaxDrivenAnalyzers.
Removing this analyzer is possible and it will correct this part, see issue #1955.


TypeCobol/Analysis/AnalyzerProviderLoader.cs
```diff
@@ -17,19 +19,28 @@ public static class AnalyzerProviderLoader
...
-                var provider = LoadProvider(assemblyFilePath);
-                compositeAnalyzerProvider.AddProvider(provider);
+                try
+                {
+                    var provider = LoadProvider(assemblyFilePath);
+                    compositeAnalyzerProvider.AddProvider(provider);
+                }
+                catch (Exception exception)
+                {
+                    File.AppendAllText(logFilePath, $"Failed to load analyzer provider on path {assemblyFilePath}\n{exception.Message}\n{exception}");
```
fm-117, line 36: Use `Environment.NewLine` instead of `\n` or use `File.AppendAllLines` to let the framework determine the newline separator. As for the error message, I would rather have

"Failed to load analyzer provider from path [path]."
"[exception.GetType().FullName] has been thrown."
"[exception.Message]"

Note: using `exception` in interpolated string will call its `ToString` method which usually return the exception type name but it may be redefined.

TypeCobol/Compiler/CompilationUnit.cs
```diff
@@ -286,7 +286,17 @@ public void RefreshCodeAnalysisDocumentSnapshot()
...
                     Dictionary<string, object> results = new Dictionary<string, object>();
-                    var analyzers = _analyzerProvider?.CreateQualityAnalyzers(CompilerOptions);
+                    IQualityAnalyzer[] analyzers = null;
+                    try
+                    {
+                        analyzers = _analyzerProvider?.CreateQualityAnalyzers(CompilerOptions);
+                    }
+                    catch (Exception exception)
+                    {
+                        var diagnostic = new Diagnostic(MessageCode.AnalyzerFailure, Diagnostic.Position.Default, $"Error while creating quality analyzers with provider : {_analyzerProvider}", exception.Message, exception);
```
fm-117, line 296: `_analyzerProvider` may be null here. I think you should create a local function `TryCreateQualityAnalyzers` and start with a code like `if (_analyzerProvider != null && TryCreateQualityAnalyzers(out var analyzers)` and then continue with the analysis itself.

TypeCobol/Analysis/AnalyzerProviderLoader.cs
```diff
@@ -17,19 +19,28 @@ public static class AnalyzerProviderLoader
...
+                }
             }
         }
 
         /// <summary>
         /// Look for a suitable type into the assembly given and instantiate a new IAnalyzerProvider from it.
+        /// <remarks>Caller must catch potential exceptions.</remarks>
```
fm-117, line 43: Absolutely ! We can even emphasize this by renaming the method to `UnsafeLoadProvider` ?

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -70,14 +70,22 @@ private TAnalyzer[] Concat<TAnalyzer>(TAnalyzer[] fromBase, Func<IAnalyzerProvid
...
+                        var analyzers = selector(analyzerProvider);
+                        result.AddRange(analyzers);
+                    }
+                    catch (Exception e)
+                    {
+                        Console.WriteLine(e);
+                        throw;
+                    }
                 }
+
```
fm-117, line 88: In fact, we don't have to change the `CompositeAnalyzerProvider`, we must secure the code at 3 locations:
- Load
- Create
- Call to analyzer itself

For load operations, it's ok with what you've done: the unsafe load is already try-catch protected and the `AddCustomProviders` is now safe.
For call to analyzer we are missing the protection on the call to `GetResult` for syntax-driven analyzers
For the create part, you secured the quality analyzer case in CompilationUnit.cs but we need to do the same for syntax-driven. So I would like you to reset `CompositeAnalyzerProvider` to its original state and add 2 try-catch blocks in `ProduceTemporarySemanticDocument` to secure the creation and result collection phases.

fm-117 (changes requested):

fm-117 (changes requested):

TypeCobol/Analysis/AnalyzerProviderLoader.cs
delevoye: Message has been updated as suggested.

TypeCobol/Analysis/AnalyzerProviderLoader.cs
delevoye: That's a good idea !

TypeCobol/Compiler/Parser/ProgramClassParserStep.cs
```diff
@@ -129,7 +129,15 @@ public static void PrepareCupParser()
...
-                    diagnostics.AddRange(customAnalyzer.Diagnostics);
+                    try
+                    {
+                        diagnostics.AddRange(customAnalyzer.Diagnostics);
+                    }
+                    catch (Exception exception)
+                    {
+                        var diagnostic = new Diagnostic(MessageCode.AnalyzerFailure, Diagnostic.Position.Default, customAnalyzers.GetType().FullName, exception.Message, exception);
+                        diagnostics.Add(diagnostic);
+                    }
```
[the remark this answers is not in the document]
delevoye: See d5b452bb6b22a64724bdcdfe1d92251aeb60c83a.

TypeCobol.Analysis.Test/CfgTestUtils.cs
```diff
@@ -49,7 +49,7 @@ static CfgTestUtils()
 
             void AddAnalyzerProvider(CfgBuildingMode mode)
             {
-                var analyzerProvider = new AnalyzerProvider();
+                var analyzerProvider = new AnalyzerProvider(str => throw new Exception(str));
```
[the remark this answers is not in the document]
delevoye: See d5b452bb6b22a64724bdcdfe1d92251aeb60c83a.

CLI/src/CLI.cs
```diff
@@ -122,11 +122,11 @@ private ReturnCode Compile()
             var rootSymbolTable = LoadIntrinsicsAndDependencies();
 
             //Add analyzers
-            var analyzerProvider = new CompositeAnalyzerProvider();
+            var analyzerProvider = new CompositeAnalyzerProvider(str => File.AppendAllText(_configuration.LogFile ?? TypeCobolConfiguration.DefaultLogFileName, str));
             var reports = RegisterAnalyzers(analyzerProvider);
 
             //Add external analyzers
-            analyzerProvider.AddCustomProviders(_configuration.CustomAnalyzerFiles);
+            analyzerProvider.AddCustomProviders(_configuration.CustomAnalyzerFiles, _configuration.LogFile ?? TypeCobolConfiguration.DefaultLogFileName);
```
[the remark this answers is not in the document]
delevoye: See d5b452bb6b22a64724bdcdfe1d92251aeb60c83a.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -15,21 +16,62 @@ public class AnalyzerProvider : IAnalyzerProvider
...
+                            syntaxDrivenAnalyzers.Add(sda);
+                        }
+                    }
+                    catch (Exception exception)
+                    {
+                        Logger($"Failed to create syntax driven analyzers from activator {sdaActivator.Method}.{Environment.NewLine}{exception.GetType().FullName} has been thrown.{Environment.NewLine}{exception.Message}");
+                    }
+                }
+            }
+            return syntaxDrivenAnalyzers.ToArray();
```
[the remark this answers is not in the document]
delevoye: See f20f266b6c7f7d0ab495426694ff8214697e6a38.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -15,21 +16,62 @@ public class AnalyzerProvider : IAnalyzerProvider
...
+            Logger = logger;
+        }
 
         public virtual ISyntaxDrivenAnalyzer[] CreateSyntaxDrivenAnalyzers(TypeCobolOptions options, TextSourceInfo textSourceInfo)
         {
-            return _sdaActivators?
-                .Select(sdaActivator => sdaActivator(options, textSourceInfo))
-                .Where(sda => sda != null)
-                .ToArray();
+            if (_sdaActivators == null) return new ISyntaxDrivenAnalyzer[]{};
```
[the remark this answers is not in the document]
delevoye: See f20f266b6c7f7d0ab495426694ff8214697e6a38.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -15,21 +16,62 @@ public class AnalyzerProvider : IAnalyzerProvider
...
+            var syntaxDrivenAnalyzers = new List<ISyntaxDrivenAnalyzer>();
+            foreach (var sdaActivator in _sdaActivators)
+            {
+                if (sdaActivator != null)
```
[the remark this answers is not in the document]
delevoye: See f20f266b6c7f7d0ab495426694ff8214697e6a38.

TypeCobol/Compiler/CompilationUnit.cs
```diff
@@ -228,8 +228,18 @@ public void ProduceTemporarySemanticDocument()
...
+                    Diagnostic exceptionDiagnostic = null;
+                    try
+                    {
+                        customAnalyzers = _analyzerProvider?.CreateSyntaxDrivenAnalyzers(CompilerOptions, TextSourceInfo);
+                    }
+                    catch (Exception exception)
+                    {
+                        Debug.Assert(_analyzerProvider != null);
+                        // Create a diagnostic to register it later when possible
+                        exceptionDiagnostic = new Diagnostic(MessageCode.AnalyzerFailure, Diagnostic.Position.Default, _analyzerProvider.GetType().FullName, exception.Message, exception);
```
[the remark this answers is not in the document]
delevoye: See f20f266b6c7f7d0ab495426694ff8214697e6a38.

TypeCobol/Compiler/CompilationUnit.cs
```diff
@@ -321,6 +347,26 @@ public void RefreshCodeAnalysisDocumentSnapshot()
...
+                        {
+                            try
+                            {
+                                createdAnalyzers = _analyzerProvider.CreateQualityAnalyzers(CompilerOptions);
+                                return true; // Analyzers have been created
+                            }
+                            catch (Exception exception)
+                            {
+                                var diagnostic = new Diagnostic(MessageCode.AnalyzerFailure, Diagnostic.Position.Default, $"Error while creating quality analyzers with provider : {_analyzerProvider.GetType().FullName}", exception.Message, exception);
+                                diagnostics.Add(diagnostic);
```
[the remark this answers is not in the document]
delevoye: See f20f266b6c7f7d0ab495426694ff8214697e6a38.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -10,103 +12,177 @@ namespace TypeCobol.Analysis
     /// <summary>
     /// IAnalyzerProvider default implementation. Can be used directly by adding activation delegates
     /// with the appropriate AddActivator method or by subclassing it.
+    /// It can also be used to allow composition of multiple providers into one.
+    /// This ensures that the creation of IAnalyzers won't interrupt the parsing.
```
fm-117, line 16: 'This class also ensures', it's not related to the previous sentence about composition of providers.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -10,103 +12,177 @@ namespace TypeCobol.Analysis
...
     /// </summary>
-    public class AnalyzerProvider : IAnalyzerProvider
+    public class AnalyzerProviderWrapper : IAnalyzerProvider
     {
+        /// <summary>
+        /// Method used to log exceptions.
+        /// <remarks>Should not be used directly, use the accessor Logger instead.</remarks>
+        /// </summary>
+        [NotNull]
```
fm-117, line 24: The `[NotNull]` attribute should be placed on the protected property but not here.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -10,103 +12,177 @@ namespace TypeCobol.Analysis
...
+        private Action<string> _logger = DoNothing;
+
+        /// <summary>
+        /// Method used to log exceptions.
+        /// </summary>
+        protected Action<string> Logger
+        {
+            get => _logger;
+            private set => _logger = value ?? DoNothing;
```
fm-117, line 33: - the setter should be protected to let inheritors have a public parameterless constructor and later set the logger
- could you add parentheses around the assigned value: `set => _logger = (value ?? DoNothing);` it is unclear in the current form.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -10,103 +12,177 @@ namespace TypeCobol.Analysis
...
     /// </summary>
-    public class AnalyzerProvider : IAnalyzerProvider
+    public class AnalyzerProviderWrapper : IAnalyzerProvider
     {
+        /// <summary>
+        /// Method used to log exceptions.
+        /// <remarks>Should not be used directly, use the accessor Logger instead.</remarks>
+        /// </summary>
+        [NotNull]
+        private Action<string> _logger = DoNothing;
```
fm-117, line 25: No need to assign here, it is done in constructor.

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -10,103 +12,177 @@ namespace TypeCobol.Analysis
...
         /// <summary>
-        /// Add an activation delegate to produce a new instance of ISyntaxDrivenAnalyzer.
+        /// Create analyzers of the given type from providers and catch exceptions thrown.
+        /// <remarks>If the creation of an analyzer fails, it will fail the creation of all analyzers for this particular provider but not for the others.</remarks>
         /// </summary>
-        /// <param name="activator">Non-null Func delegate to create a new ISyntaxDrivenAnalyzer.</param>
-        public void AddActivator([NotNull] Func<TypeCobolOptions, TextSourceInfo, ISyntaxDrivenAnalyzer> activator)
+        /// <typeparam name="TAnalyzer">Type of IAnalyzer created.</typeparam>
+        /// <param name="createAnalyzers">Function to create analyzers from a provider.</param>
+        /// <returns>Non-null list of created analyzers. Might be empty and can contain null values.</returns>
```
fm-117, line 70: But the first instruction returns `null`...

TypeCobol/Analysis/AnalyzerProvider.cs
```diff
@@ -10,103 +12,177 @@ namespace TypeCobol.Analysis
...
         }
 
         /// <summary>
         /// Add an analyzer provider to this one.
         /// </summary>
         /// <param name="provider">Instance of IAnalyzerProvider.</param>
-        public void AddProvider(IAnalyzerProvider provider)
+        public void AddProvider([NotNull] IAnalyzerProvider provider)
         {
+            Debug.Assert(provider != null);
```
fm-117, line 174: Either remove this assertion or add it for the 2 other AddActivator methods to improve consistency.

TypeCobol/Compiler/CompilationUnit.cs
```diff
@@ -35,8 +34,16 @@ public class CompilationUnit : CompilationDocument
...
             PerfStatsForCodeQualityCheck = new PerfStatsForCompilationStep(CompilationStep.CodeQualityCheck);
+            if (analyzerProvider != null)
+            {
+                if (!(analyzerProvider is AnalyzerProviderWrapper analyzerProviderWrapper))
+                {
+                    analyzerProviderWrapper = new AnalyzerProviderWrapper();
+                    analyzerProviderWrapper.AddProvider(analyzerProvider);
+                }
+                _analyzerProvider = analyzerProviderWrapper;
+            }
```
fm-117, line 45: It's correct but I think it would be clearer that way:
```csharp
if (analyzerProvider is AnalyzerProviderWrapper analyzerProviderWrapper)
{
    _analyzerProvider = analyzerProviderWrapper;
}
else if (analyzerProvider != null)
{
    _analyzerProvider = new AnalyzerProviderWrapper();
    _analyzerProvider.AddProvider(analyzerProvider);
}
else
{
    _analyzerProvider = null;
}
```

It's more verbose but it makes the possible values for the field more apparent.

In fact I wonder if we should allow `null` for the field. An empty wrapper could be used.

fm-117 (changes requested):

fm-117 (changes requested):
I'll let you change the CompilationUnit part and add comment in the wrapper class if needed, the rest is ok.

TypeCobol/Analysis/AnalyzerProvider.cs
delevoye: Assertion removed as the methods have the `[NonNull]` annotation.

TypeCobol/Compiler/CompilationUnit.cs
delevoye: The initialization is now more verbose.
Concerning allowing or not the value `null`, I think keeping it `null` stays in the spirit of the parser which is to have null instead of empty object.

fm-117 (approved):

PR #2036, opened 2021-09-01
WI #1846 multi line diagnostics

### LINKED ISSUE #1846: Support diagnostics on multiple lines
Following #1845 and to match LSP specifications, our parser should allow definition of diagnostics spanning over multiple lines.
In our current implementation, only line start is considered. When serializing to JSON (for LSP messages) line start is duplicated and send also as line end for each diagnostic.

The goal of this issue is to add a `LineEnd` property in `Diagnostic` class and update all usages accordingly.

delevoye:
Close #1846


CLI/src/Writer.cs
```diff
@@ -141,7 +141,14 @@ public class XMLWriter : AbstractErrorWriter {
...
             writer.WriteElementString("MSGNUMBER",
                 "TC-" + error.Info.Code.ToString().PadLeft(5, '0') + AsIBMSuffix((int) error.Info.Severity));
-            writer.WriteElementString("MSGLINE", error.Line.ToString());
+            if (error.LineStart == error.LineEnd)
+            {
+                writer.WriteElementString("MSGLINE", error.LineStart.ToString());
+            }
+            else
+            {
+                writer.WriteElementString("MSGLINE", $"{error.LineStart}:{error.LineEnd}");
```
delevoye, line 150: Is that valid? How to know if this is correct?

fm-117: I don't know. We have to find how this format is used by our clients.

delevoye: There is [this page](https://www.ibm.com/docs/en/zvm/7.2?topic=areas-msgline) in the IBM's documentation saying that [RSCS](https://www.ibm.com/docs/en/zvm/7.1?topic=features-rscs-networking-zvm) creates chain of MSGLINE for multiple lines.

fm-117: Hum interesting find, but I don't think it's _that_ MSGLINE. The XML file written here is used to transfer diagnostics to our private RDZ editor (instead of using LSP).

delevoye: Apparently [this is the schema](https://www.ibm.com/docs/en/adfz/developer-for-zos/15.0.0?topic=mechanism-error-feedback-xml-schema) we need to follow.

fm-117 (approved):

mayanje (approved):

Issue #2039, opened 2021-09-22
Store cycles during recursive perform detection


fm-117:
The current recursive PERFORM detection stores only the PERFORM node that lead to a cycle.
To improve diagnostics, we need to keep the cycle that has been detected.

**Technical:**
See if the `RecursivityGroupSet` can be replaced with a `Stack<...>` that would keep track of perform chains.

PR #2052, opened 2021-10-21
WI #2039 Compute cyclic call using transitive closure.

mayanje:
**What have been done**

We noticed that all cyclic perform call chains were not detected by the curent implementation: a -> b -> c -> a was not detected.
Thus as we know all direct calls, that is to say we know: a -> b, b -> c  and c -> a
The idea here is to compute the transitive closure of all known direct calls.
So we compute the transitive closure of the direct call relation which is know as a transitive relation: a->b and b->c implies a->c
During the transitive closure computation, we compute the call chains of cycles in parallel.

### DIFF
--- a/TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
+++ b/TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
@@ -82,10 +82,7 @@ namespace TypeCobol.Analysis.Cfg
             else 
             {
                 procedures.Add(procedure);
-                foreach (var sentence in procedure)
-                {
-                    sentences.Add(sentence);
-                }
+                sentences.AddRange(procedure);
             }
             target = new PerformTarget(sentences, procedures);
             _performTargetCache[p] = target;


TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -986,13 +986,196 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
+                if (!reportedProcedures.Contains(item.Value.Item1))
+                {
+                    if (item.Value.Item3.Contains(item.Key))
+                    {
+                        this.Cfg.AddRecursivePerform(item.Value.Item1, item.Value.Item2);
+                    }
+                }
+            }
+
+            void dfs(Procedure a)
```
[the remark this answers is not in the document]
mayanje: Ok this algorithm is not my invention, it is based on J. Eve, R. Kurki-Suonio transitive closure calculation algorithm, you can google on it, there are  many adaptation. I implemented also an adaptation for gathering perform chain path.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -1056,6 +1248,7 @@ void StoreSentences(IEnumerable<Sentence> sentences)
                         if (block is BasicBlockForNodeGroup group0)
                         {
                             isPerform = true; //To avoid a second dynamic cast
+                            AddDirectCallRelation(p, currentProcedure, (PerformProcedure)group0.Instructions.Last.Value, callRelation, callRelationDomain);
 
                             //Is there a recursion in the graph ?
```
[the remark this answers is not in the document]
mayanje: Yes done in d80af72

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -1023,30 +1206,39 @@ private void ResolvePendingPERFORMProcedure(Tuple<PerformProcedure, SectionNode,
...
             //And finally relocate the Graph.
             RelocateBasicBlockForNodeGroupGraph(p, group, clonedBlocksIndexMap);
 
-            void StoreSentences(IEnumerable<Sentence> sentences)
+            void StoreSentences(IEnumerable<Sentence> sentences, List<Procedure> procedures)
             {
+                int n = 0;
                 foreach (var sentence in sentences)
                 {
+                    var currentProcedure = n >= procedures.Count ? procedures[procedures.Count - 1] : procedures[n++];
```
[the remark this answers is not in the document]
mayanje: done in 9bf7eb1

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -986,13 +986,196 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
+            {
+                if (!callRelation.TryGetValue(caller, out var callees))
+                {
+                    callees = new Tuple<PerformProcedure, List<Node>, HashSet<Procedure>>(performCaller, new List<Node>(), new HashSet<Procedure>());
+                    callRelation[caller] = callees;
+                }
+                if (!callees.Item3.Contains(_calleeProcedure))
+                {
+                    callees.Item3.Add(_calleeProcedure);                    
+                }
```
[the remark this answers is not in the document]
mayanje: don in c5f891a

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -986,13 +986,196 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
+                addToRelation(calleeProcedure, callee);
+            }
+            callRelationDomain.Add(caller);
+
+            void addToRelation(Procedure _calleeProcedure, PerformProcedure _callee)
+            {
+                if (!callRelation.TryGetValue(caller, out var callees))
+                {
+                    callees = new Tuple<PerformProcedure, List<Node>, HashSet<Procedure>>(performCaller, new List<Node>(), new HashSet<Procedure>());
+                    callRelation[caller] = callees;
```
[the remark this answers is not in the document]
mayanje: done in c5f891a

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -986,13 +986,196 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
+                    addToRelation(currentProcedure, callee);
+                }
+            }
+            else
+            {
+                addToRelation(calleeProcedure, callee);
+            }
+            callRelationDomain.Add(caller);
+
+            void addToRelation(Procedure _calleeProcedure, PerformProcedure _callee)
```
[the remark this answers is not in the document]
mayanje: do,e in c5f891a

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Paragraph.cs
```diff
@@ -60,7 +60,7 @@ public override IEnumerator<Sentence> GetEnumerator()
 
             public override void AccumulateSentencesThrough(List<Sentence> sentences, Procedure end, out Procedure last)
             {
-                if (_sentences != null)
+                if (_sentences != null && sentences != null)
```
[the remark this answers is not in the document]
mayanje: done in 9bf7eb1

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -986,13 +986,196 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
+        /// Add a Direct Call to the Call Relation. 
+        /// For a PERFORM THRU all intermediate procedure are added to the relation.
+        /// </summary>
+        /// <param name="caller">The caller of the PERFOM procedure</param>
+        /// <param name="callee">The PERFOM procedure called</param>
+        /// <param name="callRelation">CALL Relation</param>
+        /// <param name="callRelationDomain">The domain of the call relation</param>
+        private void AddDirectCallRelation(PerformProcedure performCaller, Procedure caller, PerformProcedure callee, 
+            Dictionary<Procedure, Tuple<PerformProcedure, List<Node>, HashSet<Procedure>>> callRelation,
+            HashSet<Procedure> callRelationDomain)
```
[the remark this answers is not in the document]
mayanje: done in 9bf7eb1

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+        private class PerformTarget
+        {
+            /// <summary>
+            /// All target sentences
+            /// </summary>
+            internal List<Sentence> Sentences { get; private set; }
+            /// <summary>
+            /// All target procedures.
+            /// </summary>
+            internal List<Procedure> Procedures { get; private set; }
```
fm-117, line 22: You can remove private setters, they are redundant

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+                this.Sentences = sentences;
+                this.Procedures = procedures;
+            }
+        }
+
+        /// <summary>
+        /// Cache of PerformTarget instances already calculated for a Perform procedure 
+        /// </summary>
+        private Dictionary<PerformProcedure, PerformTarget> _performTargetCache;
+        private PerformTarget ComputePerformTarget(PerformProcedure p, SectionNode sectionNode)
```
fm-117, line 35: naming, this is more like `GetPerformTarget` since we do not compute the target each time the method is called.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            if (_performTargetCache == null)
+                _performTargetCache = new Dictionary<PerformProcedure, PerformTarget>();
+            if (_performTargetCache.TryGetValue(p, out var target)) {
```
fm-117, line 39: Java indent style

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+using TypeCobol.Compiler.CodeElements;
+using TypeCobol.Compiler.Nodes;
+using SectionNode = TypeCobol.Compiler.Nodes.Section;
+
+namespace TypeCobol.Analysis.Cfg
+{
+    public partial class ControlFlowGraphBuilder<D>
+    {
+        /// <summary>
+        /// PerfomrTarget class to capture target sentences and procedures from a PERFORM instruction.
```
fm-117, line 11: Typo: PerformTarget class

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (_performTargetCache.TryGetValue(p, out var target)) {
+                return target;
+            }
+
+            SymbolReference procedureReference = p.CodeElement.Procedure;
+            SymbolReference throughProcedureReference = p.CodeElement.ThroughProcedure;
+
+            Node procedureNode = ResolveProcedure(p, sectionNode, procedureReference);
+            if (procedureNode == null)
+                return null;
```
fm-117, line 48: This remark goes for all the `return null;` instructions in this method. We should actually store the `null` value inside the dictionary to prevent attempting resolving the target multiple times.
Even better: we should create a special instance `Unresolved` with empty sentences and procedures to use whenever a `PerformProcedure` cannot have its target properly resolved. This way the method would always return a non-null value.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    var currentProcedure = this.CurrentProgramCfgBuilder.AllProcedures[currentProcedureNumber];
+                    currentProcedure.AccumulateSentencesThrough(sentences, throughProcedure, out var lastProcedure);
+                    currentProcedureNumber = lastProcedure.Number + 1;
+                    procedures.Add(currentProcedure);
+                    while (n < sentences.Count)
+                    {
+                        sentences[n++].Procedure = currentProcedure;
+                    }
+                }
```
fm-117, line 80: No need to track the current procedure using the `n` index. The `Procedure` property of each `Sentence` object can (and should) be set at construction time. See my previous remark in this PR to see how to do it.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+            } else
```
fm-117, line 81: Java indent style ?

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Sentence.cs
```diff
@@ -17,6 +17,11 @@ private class Sentence : ProcedureDivisionRegion
             /// </summary>
             public int? FirstBlockIndex { get; }
 
+            /// <summary>
+            /// The associated procedure
+            /// </summary>
+            internal Procedure Procedure { get; set; }
```
fm-117, line 23: Should be read-only, set in constructor

TypeCobol.Analysis/Resource.Designer.cs
```diff
@@ -1,10 +1,10 @@
 ﻿//------------------------------------------------------------------------------
 // <auto-generated>
-//     This code was generated by a tool.
-//     Runtime Version:4.0.30319.42000
+//     Ce code a été généré par un outil.
```
fm-117, line 3: Strange the file is generated in French now ?

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -987,66 +987,177 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
+            }
+            callPerformRelationDomain.Add(caller);
 
-                Procedure throughProcedure = _nodeToProcedure[throughProcedureNode];
-                if (procedure.Number > throughProcedure.Number)
+            void addToRelation(Procedure procedure)
+            {
+                if (!callPerformRelation.TryGetValue(caller, out var callees))
+                {
+                    callees = new Tuple<PerformProcedure, List<Node>, HashSet<Procedure>>(performCaller, new List<Node>(), new HashSet<Procedure>());
```
fm-117, line 1017: `List<Node>` can be turned into `List<PerformProcedure>` if i recall correctly.

fm-117 (changes requested):

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
fm-117: Do we need it ? What about this much simpler algorithm ?

```csharp
        public static Dictionary<N, HashSet<N>> TransitiveClosure<N>(Dictionary<N, HashSet<N>> graph)
        {
            var transitiveClosure = new Dictionary<N, HashSet<N>>();
            var root = graph.FirstOrDefault().Key;
            if (root != null)
            {
                ComputeDirectAndIndirectSuccessors(root);
            }
            return transitiveClosure;

            HashSet<N> ComputeDirectAndIndirectSuccessors(N node)
            {
                if (transitiveClosure.TryGetValue(node, out var result))
                {
                    return result;
                }

                var directSuccessors = graph[node];
                result = new HashSet<N>(directSuccessors);
                transitiveClosure.Add(node, result);

                foreach (var directSuccessor in directSuccessors)
                {
                    foreach (var indirectSuccessor in ComputeDirectAndIndirectSuccessors(directSuccessor))
                    {
                        result.Add(indirectSuccessor);
                    }
                }

                return result;
            }
        }
```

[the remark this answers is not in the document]
mayanje: Ok here is the url of the Eve paper: http://i.stanford.edu/pub/cstr/reports/cs/tr/75/508/CS-TR-75-508.pdf
It seems to me that the algorithm that you suggest is in the best case in O(n^2) and in the worst case in O(n^4) because it can implies 4 foreach loops.
In the paper it is say that computing transitive closure is in the best case be in O(n^2) and in the worst case in O(n^3), and it is said in the paper that the algorithm in the paper usually works in O(n) to O(n^2) where n is the number of vertices.

And I don't see in your algorithm where to compute perform call path chain.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Sentence.cs
mayanje: It is set in PerformTarget

TypeCobol.Analysis/Resource.Designer.cs
mayanje: Maybe it's because I have a french version of VS2019

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: I would like, but it's because of ControlFlowGraph<N, D>.AddRecursivePerform(PerformProcedure perform, List<N> recursiveJump)

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: I would say this an abnormal situation it should never happened, let's take it in account with an assert.

mayanje: In fact calling an inexisting procedure is a semantic error, so no need to assert here, and simply return null here without storing such error value.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: done in 4b969b9

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: done in 4b969b9

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: done in 4b969b9

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: done in 4b969b9

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: done in 4b969b9

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.Sentence.cs
mayanje: right done in 4b969b9

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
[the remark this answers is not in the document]
fm-117: Yes it was an attempt at writing a simpler algorithm to help understanding the transitive closure. Anyway my code does not work properly for cyclic graphs so we can forget about it.

However the main question still remains, do we need to compute the full closure ? Our goal is to gather cyclic chains but not necessarily building the transitive closure of the perform call graph.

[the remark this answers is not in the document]
mayanje: My answer is yes we need to compute the full transitive closure, it is this computation that discovers Strongly connected components (that is to say cyclic chains).

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformCallRelation.cs
```diff
@@ -0,0 +1,202 @@
...
+                // For each procedure in the relation's domain
+                foreach (var a in Domain)
+                { // For each procedure 'a' not yet considered, treat it.
+                    var na = Visited(a);
+                    if (na == 0)
+                        dfs(a);
+                }
+
+                // Depth First Search Traversal.
+                void dfs(Procedure a)
```
smedilol, line 125: This last version of the code without Tuple is much clearer.
Can you use more meaningful name for variables?
Like `procedure` or at least `proc` instead of `a` for the name of the procedure.
Same remarks for variables : `d`, `fa`, `b`, `nb`, ...

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformCallRelation.cs
```diff
@@ -0,0 +1,202 @@
...
+                    S.Push(a);
+                    int d = S.Count;
+                    N[a] = d;
+                    var fa = GetPerformCallChain(a);
+                    if (fa.Procedures != null)
+                    {   //For all procedure b called by a
+                        foreach (var b in fa.Procedures.ToArray())
+                        {
+                            int nb = Visited(b);
+                            if (nb == 0)
```
smedilol, line 136: With my previous remarks on meaningful variable name, you could inline variable:
```C#
                            if (Visited(b)== 0)
```

smedilol (changes requested):

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -1174,24 +1181,27 @@ private void ResolvePendingPERFORMProcedures()
...
                 for (int i = 0; i < clonedPerforms.Count; i++)
                 {
                     //We are using a regular for instead of foreach because new groups may also be created during this second pass.
                     //New groups/performs to resolve are added at tail and all are processed during this second pass
-                    ResolvePendingPERFORMProcedure(clonedPerforms[i], clonedPerforms);
+                    ResolvePendingPERFORMProcedure(clonedPerforms[i], clonedPerforms, performCallRelation);
                 }
 
+                // Report Recursive PERFORM call
+                ReportRecursivePerformCall(performCallRelation);
```
fm-117, line 1203: As we have now a new mechanism to detect recursive PERFORMs, we can remove the `RecursivityGroupSet` property from `BasicBlockForNodeGroup` class. I've checked the code, it is now written only, never read.

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
```diff
@@ -226,7 +226,7 @@ internal void AddWrongOrderPerformThru(PerformProcedure performThru, N procedure
         /// </summary>
         /// <param name="perform">Perform node</param>
         /// <param name="recursiveJump">Recursive jump node</param>
-        internal void AddRecursivePerform(PerformProcedure perform, N recursiveJump)
+        internal void AddRecursivePerform(PerformProcedure perform, List<N> recursiveJump)
```
fm-117, line 229: There is no need to check for the existence of the perform node entry in the dictionary now, we can call `Add` each time. The second parameter should be renamed also (but I can't find a proper name yet, it is not the cycle, it's all the perform reached through the one passed as first arg).

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -634,18 +634,15 @@ public override void Exit(Node node)
...
         /// </summary>
-        /// <param name="sentence">The sentence to link.</param>
+        /// <param name="sentence">The sentence to link.</param>       
         private void LinkBlockSentenceToCurrentSectionParagraph(Sentence sentence)
         {
-            var currentProcedure = (Procedure) this.CurrentProgramCfgBuilder.CurrentParagraph ??
-                                   this.CurrentProgramCfgBuilder.CurrentSection;
-
-            if (currentProcedure != null)
+            if (sentence.Procedure != null)
```
fm-117, line 640: Now that I read it again, `sentence.Procedure` should never be null since we have a default root section.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -986,67 +986,83 @@ private Node ResolveProcedure(Node node, SectionNode sectionNode, SymbolReferenc
...
-                //Store all sentences from the procedure into the group
-                StoreSentences(procedure);
-            }
-
+            StoreSentences(performTarget);
             //And finally relocate the Graph.
             RelocateBasicBlockForNodeGroupGraph(p, group, clonedBlocksIndexMap);
 
-            void StoreSentences(IEnumerable<Sentence> sentences)
+            void StoreSentences(PerformTarget target)
```
fm-117, line 1061: Parameter is not used, remove it please.

fm-117 (changes requested):

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: It is not a good practice to store null values for semantic error reason.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Yep, there is a bug in presence of DECLARATIVE:
https://github.com/TypeCobolTeam/TypeCobol/issues/2081

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformCallRelation.cs
mayanje: Done in 35d0734

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformCallRelation.cs
mayanje: Done in 35d0734

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Done in 35d0734

TypeCobol.Analysis/Graph/ControlFlowGraph.cs
mayanje: Done in 35d0734

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
mayanje: Done in 35d0734

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            Procedure procedure = _nodeToProcedure[procedureNode];
+            List<Sentence> sentences = new List<Sentence>();
+            List<Procedure> procedures = new List<Procedure>();
+            if (throughProcedureReference != null)
+            {
+                Node throughProcedureNode = ResolveProcedure(p, sectionNode, throughProcedureReference);
+                if (throughProcedureNode == null)
+                    return null;
```
fm-117, line 61: The `null` value should be stored here also.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,94 @@
...
+                
+                Procedure throughProcedure = _nodeToProcedure[throughProcedureNode];
+                if (procedure.Number > throughProcedure.Number)
+                {
+                    // the second procedure name is declared before the first one.
+                    this.Cfg.AddWrongOrderPerformThru(p, procedureNode, throughProcedureNode);
+                    return null;
```
fm-117, line 68: And here too

fm-117 (changes requested):

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
mayanje: done in 49d4035

smedilol (changes requested):

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformCallRelation.cs
```diff
@@ -0,0 +1,204 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TypeCobol.Compiler.Nodes;
+
+namespace TypeCobol.Analysis.Cfg
+{
+    public partial class ControlFlowGraphBuilder<D>
```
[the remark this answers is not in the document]
mayanje: I think yes because only ControlFlowGraphBuilder knows how to discover direct perfrom calls. Once the graph is fully calculated direct perform calls context will be difficult to reproduce without retraversing the graph and analyzing each block to look for performs and perform targets..

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.cs
```diff
@@ -1056,16 +1072,7 @@ void StoreSentences(IEnumerable<Sentence> sentences)
...
-                            //Is there a recursion in the graph ?
-                            if (group.RecursivityGroupSet.Get(group0.GroupIndex) && !group0.HasFlag(BasicBlock<Node, D>.Flags.Recursive))
-                            {
-                                //Flag group and store recursive perform
-                                group0.SetFlag(BasicBlock<Node, D>.Flags.Recursive, true);
-                                Node offendingInstruction = group0.Instructions.Last.Value;
-                                System.Diagnostics.Debug.Assert(offendingInstruction != null);
-                                this.Cfg.AddRecursivePerform(p, offendingInstruction);
-                            }
+                            AddDirectCallPerformRelation(p, currentProcedure, (PerformProcedure)group0.Instructions.Last.Value, performCallRelation);
```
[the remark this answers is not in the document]
mayanje: Yes but this is a specialized knowledge whereas BasicBlockForNodeGroup is for a general purpose.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformCallRelation.cs
```diff
@@ -0,0 +1,204 @@
...
+                Domain.Add(caller);
+                Domain.Add(callee);
+            }
+
+            /// <summary>
+            /// Compute the transitive closure of the PERFORM CALL relation.
+            /// We browse the graph represented using the Call Relation, then we found during the visit cycles.
+            /// In the same time we compute the call chain.
+            /// </summary>
+            internal void TransitiveClosure()
```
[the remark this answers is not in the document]
mayanje: Yes this will be an  improvement because IBM compiler reports theses warnings in such situation:

    12  IGYCB7309-W   There may be a loop from the "PERFORM" statement at "PERFORM (line 12.1)" to itself. 
    12  IGYCB7310-W   The "PERFORM" statement at "PERFORM (line 12.1)" cannot reach its exit.

But it seems that qualis does not report such situation also.

But our parser report a CFG error like this:
--- Diagnostics ---
Line 12[13,21] <37, Warning, General> - Warning: Statement '     go to b' located at line 10, column 13 prevents this PERFORM statement to reach its exit.

Ok for such situation I agree, Abstract Interpretation is welcome to be very similar with IBM Compiler.

TypeCobol.Analysis/Cfg/ControlFlowGraphBuilder.PerformTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+                    procedures.Add(currentProcedure);
+                }
+            } 
+            else 
+            {
+                procedures.Add(procedure);
+                foreach (var sentence in procedure)
+                {
+                    sentences.Add(sentence);
+                }
```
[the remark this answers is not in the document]
mayanje: Yes done in 99deb05

TypeCobol.Analysis.Test/BasicCfgInstrs/CyclicPerformRecursive.diag
```diff
@@ -0,0 +1,11 @@
+﻿--- Diagnostics ---
+Line 5[12,20] <37, Warning, General> - Warning: A recursive loop has been encountered while analyzing 'PERFORM a', recursive instruction is '    perform b' at line 9.
```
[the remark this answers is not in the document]
mayanje: In fact this the current way of reporting, but the Perform Call Chain contains all information to do that.

Issue #2054, opened 2021-10-25
Uncaught exception in ValueNeedsCompilationContext


mayanje:
**Description**
In the property AlphanumericValue.IsSymbolicCharacterReference an uncaught exception can be raised InvalidOperationException

![image](https://user-images.githubusercontent.com/19149269/138653223-cd6f3665-66f5-4586-a462-9953656bdd7a.png)

PR #2055, opened 2021-10-25
WI #2054 Uncaught  exception in ValueNeedsCompilationContext

mayanje:
**What has been done**

Just debug assert and return false.


mayanje:
But we can wait for a while, since we don't know the situation,, we just want here to  avoid un uncaught situation so that in LSP mode there is no uncaught exception. That's why the suggestion here is to assert in debug mode.

fm-117:
You did not change anything since my last review.

Please read again my latest comment, in its current form the 'fix' won't work. By asserting and returning `false`, you delay the exception until the next `throw` in `Value` property. The class is designed to handle a defined list of token types, if we decide to add the capabilty for handling a new type it has to be done on the whole class.

And in my opinion, this is not a good strategy because the current design seems correct so this class works in itself. The problem (which can not be reproduced yet) lies before this code.

TypeCobol/Compiler/CodeElements/Expressions/SyntaxValue.cs
```diff
@@ -257,6 +258,7 @@ public virtual bool ValueNeedsCompilationContext
                     case TokenType.IntegerLiteral: // See: GitHub #2054
                     default:
                         System.Diagnostics.Debug.Assert(false, "Unexpected literal value: " + Token.TokenType);
+                        LoggingSystem.LogMessage(LogLevel.Error, "Unexpected literal value: " + Token.TokenType + Environment.NewLine + new System.Diagnostics.StackTrace().ToString());
```
fm-117, line 261: This should work. If you prefer, you can restore the original throw, wrap the single usage of the `ValueNeedsCompilationContext` property in a try-catch block and use `LoggingSystem.LogException`. The stacktrace will be automatically included in the trace.

smedilol (approved):

fm-117 (approved):
This is a temporary modification until we find out why the method fails so the `DumpToken` method would have to be removed when we'll find a real fix.

Issue #2057, opened 2021-10-27
Copy CPY used with EXEC SQL INCLUDE must be treated as standard Cobol copy


smedilol:
**Describe the bug**
Our private mechanism for copy of type "CPY" must be used only when a copy is used with instruction `COPY xxxx`.
Instruction 
```
  EXEC SQL INCLUDE xxxxxx 
       END-EXEC.
```
must treat the copy a standard copy (not CPY).

**Expected behavior**
YMyCpy.cpy
```cobol
        01 MyCpy.
            05 var1 pic X.
```
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
      *No error because there is no level 01
           EXEC SQL INCLUDE YMyCpy
           END-EXEC.
       01 var2 pic X.
       procedure division.
      *No error
           move MyCpy to var2
           goback
           .
       END PROGRAM MyPGM.
```

PR #2060, opened 2021-11-05
WI #2057 Exclude CompilerDirectiveType.EXEC_SQL_INCLUDE

mayanje:


TypeCobol/Compiler/CupPreprocessor/CompilerDirectiveBuilder.cs
```diff
@@ -85,7 +85,7 @@ public virtual void EnterCopyCompilerStatement(TypeCobol.Compiler.Scanner.Token
             copy.TextName = GetName(qualifiedTextName.TextName);
             copy.TextNameSymbol = qualifiedTextName.TextName;
 #if EUROINFO_RULES
-            if (copy.TextName != null)
+            if (copy.TextName != null && copy.Type != CompilerDirectiveType.EXEC_SQL_INCLUDE)
```
fm-117, line 88: I think  this is not the correct place to add this test, it should be with `this.TypeCobolOptions.HasCpyCopy(declaration.TextName)` call (located line 113).

```csharp
if (copy.Type != CompilerDirectiveType.EXEC_SQL_INCLUDE && declaration != null && this.TypeCobolOptions.HasCpyCopy(declaration.TextName))
{
    // Declaration found and copy name starts with Y => apply the legacy REPLACING semantics to the copy directive
    copy.RemoveFirst01Level = true;
	[...]
}
```

mayanje: What I simply understand is that EXEC_SQL_INCLUDE is not concerned with E-I legacy

fm-117 (approved):

smedilol:
After test, it was in fast a problem with our SQLCA copy : #2252.

Issue #2059, opened 2021-10-28
Bug EI legacy - too many level 01 are removed inside CPY


smedilol:
**Describe the bug**
**FOLLOWING OF #441**

For our custom CPY, there is bug, we remove too many level 01 inside COPY.
The rule is still the same than #441:
> If a COPY is loaded with our legacy mechanism, then all its level 01 variable declared at column 8 or 9 must be removed.
> Level 01 declared in column 10 or 11 must be kept.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
        01 KeepMe0.    COPY MyCPY.
       procedure division.
           goback
           .
       END PROGRAM MyPGM.
```

MyCPY.cpy
```cobol
       01  FILLER.
000000 01  RemoveMe1                          PIC X(10000). 
000000 01  RemoveMe2 REDEFINES ADPE99. 
000000      05 KeepMe1 pic X.
000000 01 RemoveMe3.
000000      05 KeepMe2 pic X.
000000  01 RemoveMe4 pic X.
000000   01 KeepMe3 pic X.
000000    01 KeepMe4.
000000      05 KeepMe5 pic X.
000000   01 KeepMe6 pointer.
000000   01 KeepMe7 pointer.
000000   01 KeepMe8 pointer.
```

**Expected behavior**
In the previous example:
- all variables beggining with "RemoveMe" must be removed.
- all variables beggining with "KeepMe" must be kept.

PR #2061, opened 2021-11-05
WI #2059 Use < 10 rather than <= 10

mayanje:


fm-117 (approved):

smedilol (approved):
Ok for me. 
Just a note about the commit message: `WI #2059 Use < 10 rather than <= 10`.
The commit message should explain what and/or why vs. how.
In this context, the commit message could be:
`WI #2059 CPY: Don't delete level 01 in col 10 and 11`

PR #2071, opened 2021-11-10
2070 refresh nodes without missing copy notification

mayanje:
**What have been done**

ProgramClassEvent class has been added a new property Options to disable MissingCopiesEvent and DiagnosticsEvent to be sent if needed, when a ProgramClassEvent event is handled by the FinalCompilationStepCompleted method.

### DIFF
--- a/TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
+++ b/TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
@@ -143,7 +143,7 @@ namespace TypeCobol.LanguageServer.TypeCobolCustomLanguageServerProtocol
             var context = GetDocumentContextFromStringUri(parameter.textDocument.uri, Workspace.SyntaxTreeRefreshLevel.NoRefresh);
             if (context != null && context.FileCompiler != null)
             {
-                this.Workspace.RefreshSyntaxTree(context.FileCompiler, Workspace.SyntaxTreeRefreshLevel.ForceFullRefresh, ProgramClassEvent.Option.None);
+                this.Workspace.RefreshSyntaxTree(context.FileCompiler, Workspace.SyntaxTreeRefreshLevel.ForceFullRefresh);
             }
         }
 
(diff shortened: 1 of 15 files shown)


TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/TypeCobolCustomLanguageServer.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using TypeCobol.Compiler;
```
fm-117, line 4: Not needed, please remove

fm-117 (changes requested):
I'm not ok with the current implementation because the new Option is passed through the parser even though it is not concerned by it. The option value is just sent to the parser, and comes back through the event. This should be handled entirely in LanguageSever.

Also I've understood why @mayanje  had to change something in the server: the `Workspace.UpdateMissingCopies` method has changed since PR #2013 and this can indeed create notification loops. But do we really need to have this auto-refresh in the parser ?
I think the missing copy notification should be one-way, the parser is not really concerned by the fact that the client succeeded or failed to load missing copies especially if the client can request a refresh at any time. What do you think @smedilol ? My idea to solve this issue is to remove the  client-to-server notification and let the client decide when a refresh should be triggered.

smedilol:
There is at least one bug on the IDE side that needs to be fixed:
When the parser/LSP send a notification for missing copy, the IDE will delete the copy if it's present and then download it again.
The IDE is not supposed to delete the copy in this case.

On the case of a "refresh all copy", then it's ok to delete the copy.

If you fix this bug, do we still need this PR?

If so, I agree with fm-117, the new Option should be only on LSP side not on the parser side too.
The parser always fire the event, then LSP decide to send a notification or not.
Or let the client skip refreshNode until it has downloaded all missing copies.

mayanje:
So there is something strange here, why the parser/LSP will send a missing copy if it is present?
But actually the client send back a MissingCopyNotification only when it has finished to load all requested COPY, it is only when the server receives the notification from the client, then the server refreshNode.

mayanje:
As suggested by fm-117 the fix is client side not server side, the client should not send back a notification if none of requested copy is resolved.

PR #2073, opened 2021-11-10
WI #1593 Use whenCondition with no expression

mayanje:
**What have been done**

This PR is the new PR from  #1810 , it implements the idea suggested by smedilol to use a whenCondition rule which can have only a WHEN keyword.
- LSR and Cobol85 tests


TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
```diff
@@ -45,8 +45,7 @@ public static Node GetMatchingNode(FileCompiler fileCompiler, CodeElement codeEl
             {
                 return node;
             }
-
-            return null;
+            return codeElementToNode.FirstOrDefault(t => t.Key.Equals(codeElement)).Value;
```
fm-117, line 48: What is that ?? At this point `codeElementToNode` maybe null and if it is not, it does not contain the entry for `codeElement` key, so the `FirstOrDefault` will return a default `KeyValuePair` leading to a `null` value return.
Please revert.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -1341,6 +1341,22 @@ public virtual void StartWhenConditionClause(List<TypeCobol.Compiler.CodeElement
...
+                var whenGroup = SyntaxTree.CurrentNode.Parent.Children[SyntaxTree.CurrentNode.Parent.ChildrenCount - 2];
+                System.Diagnostics.Debug.Assert(whenGroup.ChildrenCount > 0);
+
+                var whenNode = whenGroup.Children[whenGroup.ChildrenCount - 1];
+                System.Diagnostics.Debug.Assert(whenNode.CodeElement != null);
+                System.Diagnostics.Debug.Assert(whenNode.CodeElement.Type == CodeElementType.WhenCondition ||
+                    whenNode.CodeElement.Type == CodeElementType.WhenSearchCondition);
+
+                //Syntax error.
+                DiagnosticUtils.AddError(whenNode, "Missing statement in \"when\" clause");
```
fm-117, line 1358: Diagnostics are usually created in checkers. Can we move that ?

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
```diff
@@ -86,6 +86,11 @@ public bool UseEuroInformationLegacyReplacingSyntax
         /// </summary>
         public TypeCobolCheckOption CheckRecursivePerforms { get; set; }
 
+        /// <summary>
+        /// Use to set that these options are used by some tests.
+        /// </summary>
+        public bool IsTestMode { get; set; }
```
fm-117, line 92: So in some tests, it appears that the `Debug.Assert` in CrossChecker fails, but we cannot add such property in `TypeCobolOptions`. It has no meaning outside of the project.

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolStatementsBuilder.cs
```diff
@@ -1435,7 +1435,12 @@ private UnstringStatement.Receiving CreateUnstringReceivingFields(CodeElementsPa
 
         internal CodeElement CreateWhenCondition(CodeElementsParser.WhenConditionContext context) {
 			var statement = new WhenCondition();
-			statement.SelectionObjects = BuildObjectArrayFromParserRules(context.comparisonRHSExpression(), ctx => CreateEvaluateSelectionObject(ctx));
+			var rhsExprs = context.comparisonRHSExpression();
+			statement.SelectionObjects = BuildObjectArrayFromParserRules(rhsExprs, ctx => CreateEvaluateSelectionObject(ctx));
+			if (rhsExprs.Length == 0)
+			{
+				DiagnosticUtils.AddError(statement, "Missing condition in \"when\" clause", context);
```
fm-117, line 1442: Maybe move that to a CodeElement checker ?

fm-117 (changes requested):

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
mayanje: I agree codeElementToNode can be null, thus I am going to replaced this line by:
codeElementToNode?.FirstOrDefault(t => t.Key.Equals(codeElement)).Value;
Without that line it does not work.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
mayanje: But here  it's a syntactical error, why move it in crosschecker, I think crosschecker is used for semantic errors ?

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
mayanje: Sorry I don't know why I came to this conclusion, I retested again, you are right it must be reverted.

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: So shall we remove this Debug Assert?

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
fm-117: I've run tests (everything except ThirdParty) in EI_Debug configuration on your branch replacing this line with the original `return null;` and it went fine. Can you point out the case where it won't work ?
The dictionary uses  the default comparer so it relies on `GetHashCode` and `Equals` methods of code elements. If this line does not return `null`, it means that `GetHashCode` and `Equals` are not synced for some CE.

mayanje: Yes it will be reverted

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolStatementsBuilder.cs
mayanje: I think it can stay here, as these are syntactical errors.

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
mayanje: done in 6a515c8

TypeCobol/Compiler/Directives/TypeCobolOptions.cs
mayanje: Assert removed in 6a515c8

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolStatementsBuilder.cs
smedilol: We create all of our Diagnostics in checkers (CodeElementCheckers in this case).
It's true we have 3 diagnostics created in CobolStatementsBuilder but this has been done a long time ago and should be moved.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
smedilol: Same remarks than CodeElementBuilder, all diagnostics are created in a checker.

TypeCobol/Compiler/Diagnostics/NodeCheckers.cs
```diff
@@ -0,0 +1,32 @@
+﻿using TypeCobol.Compiler.CodeElements;
+using TypeCobol.Compiler.Nodes;
+using TypeCobol.Compiler.Parser;
+
+namespace TypeCobol.Compiler.Diagnostics
+{
+    #region NodeCheckers
+
+    class WhenConditionNodeChecker
```
fm-117, line 9: class can be marked `static`

TypeCobol/Compiler/Diagnostics/NodeCheckers.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public static void OnNode(Node node)
```
fm-117, line 11: Argument could be more strictly typed as `When`

fm-117 (changes requested):

smedilol (changes requested):

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -1341,6 +1341,7 @@ public virtual void StartWhenConditionClause(List<TypeCobol.Compiler.CodeElement
 
         public virtual void EndWhenConditionClause()
         {
+            WhenConditionNodeChecker.OnNode(SyntaxTree.CurrentNode);
```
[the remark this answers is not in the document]
mayanje: It seems that no, I tried and I got that other tests no longer pass, but also here is a syntactical error it should be reported during the parsing phase.

TypeCobol/Compiler/Diagnostics/NodeCheckers.cs
mayanje: In fact at parsing time it should be a Then node, but the parser is not aware about that can only assert here

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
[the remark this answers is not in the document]
smedilol: What do you mean by "parsing phase"? ExecutionStep.SemanticCheck?
Our architecture rules state that every check of this kind are done in a checker during CrossCheck step (ExecutionStep.CrossCheck ). Why would we need to make an exception in this case?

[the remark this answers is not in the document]
fm-117: This should go there, right ?

https://github.com/TypeCobolTeam/TypeCobol/blob/4cc51f941a01dbb62eb94ba0f2db5330d22ffc00/TypeCobol/Compiler/Diagnostics/CrossChecker.cs#L471

fm-117 (approved):
I'm ok with this PR. The only remaining thing is the location of the new Node check for `Then`. As discussed, I think it should go into the CrossChecker and the file NodeCheckers.cs can be deleted, the check code could be directly added into the visit methood for `Then` node.

mayanje:
One thing I don't understand is that if the check is performed later in CrossChecker, it does not work and there are many unexpected results, with many tests that no longer pass.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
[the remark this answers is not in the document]
smedilol: Or directly on `Visit(Evaluate)` because we don't need to do this check on every `when`.
We only need to check that the last `when` different than a `when other` and the `when other` contains at least one instruction.
And I just notice that there is no test for all cases.
We could also check `search`/`search all` but it's a little out of the scope of this issue. I can write an issue about that.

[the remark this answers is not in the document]
smedilol: I've created issue #2087 and #2088 for similar problems, but we can only focus on the evaluate right now.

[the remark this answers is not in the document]
mayanje: What I am saying is that when  I move this check in CrossChecker I got unexpected results.

smedilol:
I've updated #1593 with new rules to check `when` and `when other` and code samples.

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolStatementsBuilder.cs
mayanje: Doing this gives unexpected behavior.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -452,9 +452,45 @@ bool CheckOperand(ConditionOperand operand)
...
+                    if (delta > 0 && evaluate.Children[count - (delta - 2)] is WhenOther &&
+                        evaluate.Children[count - (delta - 1)] is Then thenNode &&
+                        thenNode.ChildrenCount == 0 &&
+                        evaluate.Children[count - delta] is WhenGroup whenGroup1)
+                    {
+                        return whenGroup1;
+                    }
+                    return null;
+                }
+            }
```
smedilol, line 493: I suggest this shorter code and with less hard coded constants like `2`, `3` ...
Main part of the code comes from your previous version of `WhenConditionNodeChecker`:
Class WhenConditionNodeChecker can be deleted.
```suggestion
            bool whenOtherSeen = false;
            //Start to loop on children from the end because:
            //- Grammar enforce that there is 0 to 1 "whenOther" and that it's the last element of an "evaluate"
            //- Invalid empty "when" can only be the last "when"
            for (int i=evaluate.ChildrenCount-1;i>=0;i--)
            {
                if(evaluate.Children[i] is WhenOther)
                {
                    whenOtherSeen = true;
                } 
                else if(evaluate.Children[i] is Then then)
                {
                    if (then.ChildrenCount == 0)
                    {
                        // i is the Then node
                        // i-1 is the When group
                        var whenGroup = then.Parent.Children[i-1];
                        System.Diagnostics.Debug.Assert(whenGroup.ChildrenCount > 0);

                        //Get the last When of the WhenGroup. This is the invalid empty When
                        var whenNode = whenGroup.Children[whenGroup.ChildrenCount - 1];
                        System.Diagnostics.Debug.Assert(whenNode.CodeElement?.Type == CodeElementType.WhenCondition);

                        //Syntax error.
                        DiagnosticUtils.AddError(whenNode, "Missing statement in \"when\" clause");
                    }
                    break; //Previous "when" are allowed to have no instructions (#1593)

                }
            } 
            if (!whenOtherSeen)
            {
                DiagnosticUtils.AddError(evaluate, "\"when other\" is missing", MessageCode.Warning);
            }
```

smedilol (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
mayanje: done in ab444e9

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -452,9 +452,37 @@ bool CheckOperand(ConditionOperand operand)
 
         public override bool Visit(Evaluate evaluate)
         {
-            if (evaluate.GetChildren<WhenOther>().Count == 0)
-                DiagnosticUtils.AddError(evaluate,
-                    "\"when other\" is missing", MessageCode.Warning);
+            bool whenOtherSeen = false;
+            //Start to loop on children from the end because:
+            //- Grammar enforce that there is 0 to 1 "whenOther" and that it's the last element of an "evaluate"
```
fm-117, line 457: It also enforces that whenOther is not empty.

fm-117 (approved):

TypeCobol.Test/Parser/Programs/Cobol85/WhenTest.rdzMix.txt
```diff
@@ -0,0 +1,48 @@
...
+      *KO missing statement in "when other" clause
+                when other
+Line 34[12,23] <27, Error, Syntax> - Syntax error : extraneous input 'end-evaluate' expecting {AcceptStatement, AddStatement, AllocateStatement, AlterStatement, CallStatement, CancelStatement, CloseStatement, ComputeStatement, ContinueStatement, DeleteStatement, DisplayStatement, DivideStatement, EntryStatement, EvaluateStatement, ExecStatement, ExitMethodStatement, ExitProgramStatement, ExitStatement, FreeStatement, GobackStatement, GotoStatement, IfStatement, InitializeStatement, InspectStatement, InvokeStatement, JsonGenerateStatement, JsonParseStatement, MergeStatement, MoveStatement, MultiplyStatement, OpenStatement, PerformProcedureStatement, PerformStatement, ReadStatement, ReleaseStatement, ReturnStatement, RewriteStatement, SearchStatement, SetStatement, SortStatement, StartStatement, StopStatement, StringStatement, SubtractStatement, UnstringStatement, WriteStatement, XmlGenerateStatement, XmlParseStatement, ProcedureStyleCall}
+           end-evaluate      
+           move "hi" to var1
+           goback.
+Line 37[8,28] <27, Error, Syntax> - Syntax error : Unexpected orphan "PROGRAM END".
```
smedilol, line 47: Is this error due to the previous empty `when other`?
It's just to be sure that an empty `when` won't produce unexpected error.

TypeCobol.Test/Parser/Programs/Cobol85/WhenTestNoEndEvaluate.rdz.cbl
```diff
@@ -0,0 +1,34 @@
...
+                when other continue
+           search var1
+                when
+                when var2
+                    when var3 display "search"
+                when
+           end-search
+           evaluate true
+                when bool1 display "evaluate"
+      *KO missing statement in "when other" clause
```
smedilol, line 30: Without `end-evaluate`, this `when other` is not empty anymore.
Can you change the comment ?

smedilol (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/WhenTest.rdzMix.txt
mayanje: Yes the end's parent is in this situation is a sentence

fm-117 (approved):

TypeCobol.Test/Parser/Programs/Cobol85/WhenTestNoEndEvaluate.rdz.cbl
mayanje: done in 
75d47f4

smedilol (approved):

Issue #2064, opened 2021-11-08
Replace a partial word with multiples words


smedilol:
**Describe the bug**
A replace clause that match a partial cobol word and replace it with multiple tokens doesn't works.

**To Reproduce**
Replace not working:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       REPLACE ==:T:==                 BY ==05  PIC X(70) VALUE ==
               .
       WORKING-STORAGE SECTION.
       01 W-TAB-ENT.
         :T: 'val'.
       END PROGRAM MyPGM.
```
Error :  `Syntax error : extraneous input '(' expecting {separator, statement starting keyword, keyword}`

**Expected behavior**
No error, the source code after replace is:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
       DATA DIVISION.
       REPLACE ==:T:==                 BY ==05  PIC X(70) VALUE ==
               .
       WORKING-STORAGE SECTION.
       01  W-TAB-ENT.
           05  PIC X(70) VALUE 'val'. 
       END PROGRAM MyPGM.
```

PR #2074, opened 2021-11-15
WI #2064 store all tokens channels in pseudo texts.

mayanje:
**What have been done**

The fact was that Pseudo Text tokens don't include whitespaces, whitespaces are delimiters in rescanning, and also delimiters for capturing PICTURE strings. The previous rescanning mechanism used to add a whitespace between each rescanned token, this was not good for PICTURE String. Thus the fix was to keep original whitespaces in the pseudo text so that original whitespaces are used while rescanning.


fm-117:
For the record: https://github.com/TypeCobolTeam/TypeCobol/pull/1798#discussion_r519975880

TypeCobol/Compiler/Preprocessor/ReplaceTokensLinesIterator.cs
```diff
@@ -517,33 +504,80 @@ internal static Token GenerateReplacementToken(Token originalToken, string repla
...
+                        {
+                            endTokIdx = j;
+                            break;
+                        }
+                        tokens.Add(replacementTokens[j]);
+                    }
+                    bool isGroup = (endTokIdx - startTokIdx) > 1;
+                    T replacedToken = isGroup
+                        ? replacedToken = creator(new TokensGroup(rescannedToken.TokenType, new List<Token>(tokens)))
+                        : replacedToken = creator(tokens[0]);
```
fm-117, line 567: duplicate assignments

TypeCobol/Compiler/Preprocessor/ReplaceTokensLinesIterator.cs
```diff
@@ -517,33 +504,80 @@ internal static Token GenerateReplacementToken(Token originalToken, string repla
...
+                int i = 0;
+                int[] columns = new int[replacementTokens.Length + 1];                
                 StringBuilder sb = new StringBuilder();
-                foreach (var t in replacedTokens)
+                foreach (var t in replacementTokens)
                 {
-                    sb.Append(t.Text);
-                    sb.Append(' ');
+                    columns[i++] = sb.Length;
+                    sb.Append(t.Text);
```
fm-117, line 523: Unwanted whitespace

TypeCobol/Compiler/Preprocessor/ReplaceTokensLinesIterator.cs
```diff
@@ -517,33 +504,80 @@ internal static Token GenerateReplacementToken(Token originalToken, string repla
...
                 string tokenText = sb.ToString();
+                int startTokIdx = 0;
+                int endTokIdx = 0;                
+                List<T> newReplacedTokens = new List<T>(replacementTokens.Length);
                 TokensLine tempTokensLine = TokensLine.CreateVirtualLineForInsertedToken(0, tokenText);
                 tempTokensLine.InitializeScanState(scanState);
                 var tempScanner = new TypeCobol.Compiler.Scanner.Scanner(tokenText, 0, tokenText.Length - 1, tempTokensLine, new TypeCobolOptions(), true);
                 Token rescannedToken = null;
-                int i = 0;
+                List<Token> tokens = new List<Token>((replacementTokens.Length >> 1) + 1);
```
fm-117, line 534: Why setting this capacity ? Why using bit shifting operator ?

fm-117 (changes requested):

TypeCobol/Compiler/Preprocessor/ReplaceTokensLinesIterator.cs
mayanje: I just wanted to optizme space here, tokens is used for gathering Tokens Groups, so in most cases grouped tokens will never exceed (ReplacementToken.Length / 2) + 1; Right shift operator is the same than dividing by 2.

TypeCobol.Test/Parser/Preprocessor/ReplaceResultFiles/PgmReplaceFunction.txt
```diff
@@ -4,24 +4,25 @@
...
-[8,21:call]<CALL>
+[8,21:   call]<CALL>
 [8,21:'CREATE-ERROR']<AlphanumericLiteral>(',Y,Y){CREATE-ERROR}
 [8,21:using]<USING>
 [8,21:CCTERR]<UserDefinedWord>
 [8,21:CCTEXT]<UserDefinedWord>
 [8,21:CCTFAL]<UserDefinedWord>
 [8,21:CCTZON]<UserDefinedWord>
-[8,21:CCTFRE]<UserDefinedWord>
+[8,21:                             ]<UserDefinedWord>
```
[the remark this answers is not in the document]
mayanje: Yep this a bug in how Tokens Group were gathering, fixed in c52be48

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
```diff
@@ -633,7 +633,7 @@ private static IList<DocumentChange<IProcessedTokensLine>> ProcessTokensLinesCha
                         tokensLineWithCopyDirective.ImportedDocuments.Keys.Where(
                             c => c.TextName != null || c.COPYToken.TokenType == TokenType.EXEC).ToArray())
                     {
-                        if (copyDirective.TextName == null)
+                        if (copyDirective.TextName == null || tokensLineWithCopyDirective.ScanStateBeforeCOPYToken == null)
```
fm-117, line 636: Do we have cases where the scan state dictionary is `null` ? The token line with COPY directive should have this dictionary set according to this code:

https://github.com/TypeCobolTeam/TypeCobol/blob/4cc51f941a01dbb62eb94ba0f2db5330d22ffc00/TypeCobol/Compiler/Scanner/TokensLine.cs#L73-L80

So maybe turn it into a `Debug.Assert` ?

TypeCobol/Compiler/Preprocessor/ReplaceTokensLinesIterator.cs
fm-117: I know what the operator does but it is unusual to use it. I think it is clearer for the reader to see the divide by 2 operation explicitly. Or add a comment to explain this optimization.

fm-117 (changes requested):

TypeCobol/Compiler/Preprocessor/ReplaceTokensLinesIterator.cs
mayanje: change in c52be48

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
mayanje: I really don't know but I put this because it happened. Note that the code that you report is in the AddToken() method, thus if we have for any reason a TokenLine instance in which no token has been added ScanStateBeforeCOPYToken can be null, thus it is not an assertion. But I got such situation, for instance when executing some tests I think.

fm-117: I ran all unit tests in EI_Debug with the second part of the condition commented and encountered no error.

mayanje: I got it,maybe I will be able to reproduce it one day.

smedilol (dismissed):

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
fm-117: The condition has already been added for the cup method and it's on develop now, but i checked the tests without it and it works so could you:
- update this branch with latest changes from develop
- remove the condition in cup method and turn it into an assert

TypeCobol/Compiler/Scanner/TokensGroup.cs
```diff
@@ -77,5 +77,30 @@ public override string ToString()
...
+                    return base.SourceText;
+                else 
+                {
+                    StringBuilder sb = new StringBuilder();
+                    foreach (Token token in OriginalTokens)
+                    {
+                        sb.Append(token.SourceText);
+                    }
+                    String t = sb.ToString();
+                    return t;
```
fm-117, line 100: Maybe this should be generalized for all TokensGroup not only PartialCobolWord ? I thinl it is more related to the fact that original tokens span over multiple lines instead of TokenType equals to PartialCobolWord.

fm-117 (changes requested):
Keep also in mind my remark about the condition that we could remove in CupProcessTokensLinesChanges.

TypeCobol/Compiler/Scanner/TokensGroup.cs
mayanje: done in fe1bab6

TypeCobol/Compiler/Preprocessor/PreprocessorStep.cs
fm-117: Ok there is a bug on EXEC SQL INCLUDE, the ScanStateBeforeCOPYToken should be non-null. See #2105.
No need to change the test here in this PR.

fm-117 (approved):

smedilol (approved):

Issue #258, opened 2016-07-11
Replace a partial Cobol word with an empty string produce a NullReferenceException


smedilol:
With this code:
```cobol
001390*REPLACE ==:-MAJ:== BY ==-TRACE==.                                01340000
001400*         /                                                       01350000
001420 REPLACE ==:-MAJ:== BY ==      ==.                                01370000
024820     PERFORM EACGAPEL-C0001:-MAJ:                                 24770000
025090*---------------------------------------------------------------* 25040000
025100 EACGAPEL-C0001-TRACE.                                            25050000
000000     continue
025220     .                                                            25170000
025230                                                                  25180000
025240*---------------------------------------------------------------* 25190000
025250 EACGAPEL-C0001.                                                  25200000
000000     .
```
The line `PERFORM EACGAPEL-C0001:-MAJ:` produce the exception

PR #2080, opened 2021-11-17
WI #2075 Use an empty SpaceSeparator token if empty replace.

mayanje:
**What have been done**

This issue is in relation with #258, then case when the replacement token is null and both part to be replaced and the partial word are the same were missing, because in such sitution the resulting token is an empty token.
So the fix was to create an empty SpaceSeparator token.

A new test is added.


fm-117 (approved):

TypeCobol.Test/Parser/Preprocessor/ReplaceTestFiles/PgmEmptyPartialWordReplace2.cpy
```diff
@@ -0,0 +1,13 @@
+﻿001390*REPLACE ==:-MAJ:== BY ==-TRACE==.                                01340000
```
smedilol, line 1: Can you also create a test with a full program under "TypeCobol.Test/Parser/Programs/Cobol85/" just to be sure this PR integrates well with the rest ?

smedilol (changes requested):

fm-117 (approved):

smedilol (approved):

Issue #2070, opened 2021-11-09
LSP Allow RefreshNodes Without Missing Copy Notification


mayanje:
**Description**
Missing Copy Notification can lead to recursive Missing Copy Notification when a refresh nodes notification is requested by the client, for instance after a Refresh All Copy actions and there still unresolved copy.
So the strategy to avoid such situation is to refresh node without sending again a Missing Copy Notification to the client.

fm-117:
Isn't that a clienr preoccupation ? I don't see anything wrong with sending a MissingCopyNotification at the end of the RefreshNodes process.

mayanje:
The fact here is that the client notifies the server that it has finished to load missing copies, when the server received the notification it reparses opened document, thus in this specific case the server should not send again a missing copy notification, because we can enter in an infinite loop of notifications.

fm-117:
Yes I've understood where the loop comes from, no need to repeat.

In my opinion the server should continue sending a MissingCopyNotification after a node refresh but the client should not send any notification after loading copies. Copy loading is a client-side process only. Depending on whether the client succeeded or not to load all the copies, it could ask (so it is a request not a notification) the server for a node refresh. I personally think that the MissingCopyNotification is one-way only from server to client.

mayanje:
It cannot be a request, a request is a blocking operation until the server responds to the request.
NodeRefreshParams is a notifiation, if so we should add a new options to this notification to not resend MissingCopyNotification

mayanje:
As suggested by fm-117 the fix is client side not server side, the client should not send back a notification if none of requested copy is resolved.

fm-117:
But we still need to remove the code MissingCopyNotification from client to server, so after the client has been updated we'll have to change the server.

PR #2082, opened 2021-11-23
WI #2070 Clear Copy Cache

mayanje:
** What have been done**
It is necessary to clear the COPY cache.

### DIFF
--- a/TypeCobol.LanguageServer/Workspace.cs
+++ b/TypeCobol.LanguageServer/Workspace.cs
@@ -644,6 +644,7 @@ namespace TypeCobol.LanguageServer
         public void RefreshOpenedFiles()
         {
             RefreshCustomSymbols();
+            this.CompilationProject?.ClearImportedCompilationDocumentsCache();
 
             lock (_lockForOpenedDocuments)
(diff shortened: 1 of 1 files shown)


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -644,6 +644,7 @@ public void UpdateMissingCopies(Uri fileUri, List<string> RemainingMissingCopies
         public void RefreshOpenedFiles()
         {
             RefreshCustomSymbols();
+            this.CompilationProject?.ClearImportedCompilationDocumentsCache();
```
fm-117, line 647: `CompilationProject` can't be `null`. It is safe to call the method without check.

fm-117 (changes requested):

fm-117 (approved):

TypeCobol.LanguageServer/Workspace.cs
mayanje: ok done in 0fd8647

smedilol (approved):

Issue #1880, opened 2021-02-08
Optimize duplicate paragraph check (CrossChecker)


smedilol:
The current paragraph detection iterate over all declared paragraphs to find duplicate.
The problem is that this iteration is done each time a paragraph is visited in CrossChecker.

## Quick code
Here's a quick code to gain about 20% performance in incremental mode. 
Tested on my PC. Need to be confirmed on the target system.

**CrossChecker**
```C#
        public static void CheckParagraph(Paragraph paragraph)
        {
            var paragraphs = paragraph.SymbolTable.GetParagraphs(paragraph.Name);

            if(paragraphs != null)
            {
                var filter = paragraphs.Where(p => p.SemanticData.Owner == paragraph.SemanticData.Owner);

                //Paragraphs can't have the same name within the same section
                if (filter.Any())
                {
                    //Get the name of the scope to display in diagnostic message
                    var scope = string.IsNullOrEmpty(paragraph.Parent.Name) ? paragraph.Parent.ID : paragraph.Parent.Name;
                    DiagnosticUtils.AddError(paragraph, $"Paragraph \'{paragraph.Name}\' already declared in {scope}", MessageCode.Warning);
                }
            }

            CheckIsNotEmpty("Paragraph", paragraph);
        }
```

**SymbolTable**
```C#
        /// <summary>
        /// Get all paragraphs in the current scope.
        /// </summary>
        /// <returns>The collection of paragraph names</returns>
        public List<Paragraph> GetParagraphs(string paragraphName)
        {
            if(Paragraphs.TryGetValue(paragraphName, out List<Paragraph> paragraphs)) {
                return paragraphs;
            }
            return null;
        }
```
TODO :
- Don't return the internal list of Paragraphs. Rather return ReadOnlyList of IEnumerable.
- Create a method GetParagraphs(string paragraphName, Symbol owner) to avoid "filter" in CrossChecker ?
     - This will avoid to return a null value

PR #2098, opened 2021-12-16
WI #1880 Use Semantic data to optimize CrossChecker.CheckParagraph

mayanje:
**What have been done**

Semantic data on symbols having Paragraphs are used to optimize lookup, then the paragraph is looked up  directly in the owner paragraph domain.

These are the perfomance gains in milliseconds (dopt = optimized time, dcur = old time using previous implementation) 
mesured with the test: Parser/Programs/Cobol85/ParagraphSection.rdz.cbl 

dopt=1, dcur=91
dopt=0, dcur=1
dopt=0, dcur=2
dopt=0, dcur=0

We can notice that in  the worst case the gain is of 90%.


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1223,14 +1223,35 @@ public static void CheckSection(Section section)
...
+                    break;
+                default:
+                    System.Diagnostics.Debug.Assert(paragraph.SemanticData.Owner.Kind == Symbol.Kinds.Program ||
+                        paragraph.SemanticData.Owner.Kind == Symbol.Kinds.Function ||
+                        paragraph.SemanticData.Owner.Kind == Symbol.Kinds.Section);
+                    break;
+            }
+            // Look for the paragraph in the domain
+            Scopes.Container<ParagraphSymbol>.Entry entry = paragraphs.Lookup(paragraph.Name);
+            if (entry.Count > 1)
```
fm-117, line 1251: Before testing the count, you can assert that the entry is not null as it contains the paragraph itself.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1223,14 +1223,35 @@ public static void CheckSection(Section section)
...
             {
                 //Get the name of the scope to display in diagnostic message
-                var scope = string.IsNullOrEmpty(paragraph.Parent.Name) ? paragraph.Parent.ID : paragraph.Parent.Name ;
+                var scope = string.IsNullOrEmpty(paragraph.Parent.Name) ? paragraph.Parent.ID : paragraph.Parent.Name;
```
fm-117, line 1254: This is the owner's name so by keeping a reference on the owner outside of the switch, you can simplify this line also.

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1223,14 +1223,35 @@ public static void CheckSection(Section section)
...
+                case Symbol.Kinds.Section:
+                    {
+                        SectionSymbol section = (SectionSymbol)paragraph.SemanticData.Owner;
+                        paragraphs = section.Paragraphs;
+                    }
+                    break;
+                default:
+                    System.Diagnostics.Debug.Assert(paragraph.SemanticData.Owner.Kind == Symbol.Kinds.Program ||
+                        paragraph.SemanticData.Owner.Kind == Symbol.Kinds.Function ||
+                        paragraph.SemanticData.Owner.Kind == Symbol.Kinds.Section);
```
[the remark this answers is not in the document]
mayanje: If you like, done in c903c1d

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
mayanje: The fact is that the diagnostic, can report 'procedure-division' as location.

fm-117 (approved):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
fm-117: Well, in fact 'procedure-division' is not a very good name, the program's name would be better.

fm-117 (changes requested):

fm-117 (approved):
Ok to keep the current parent name mechanism for diagnostic. Sorry for the double validation.

smedilol:
What is the test `Parser/Programs/Cobol85/ParagraphSection.rdz.cbl`? How do you run it?
What about our Performance tests ?

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1223,14 +1223,34 @@ public static void CheckSection(Section section)
...
+                        SectionSymbol section = (SectionSymbol)paragraph.SemanticData.Owner;
+                        paragraphs = section.Paragraphs;
+                    }
+                    break;
+                default:
+                    System.Diagnostics.Debug.Fail("A Program, a Function or a Section was expected as paragraph's owner.");
+                    break;
+            }
+            // Look for the paragraph in the domain
+            Scopes.Container<ParagraphSymbol>.Entry entry = paragraphs.Lookup(paragraph.Name);
```
smedilol, line 1248: The only way to resolve a Symbol is the SymbolTable, we've already discuss about that.

smedilol (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1223,14 +1223,34 @@ public static void CheckSection(Section section)
...
+            System.Diagnostics.Debug.Assert(entry != null);
+            if (entry.Count > 1)
```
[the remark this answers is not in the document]
mayanje: done in b1ae8ce

smedilol (approved):

PR #2099, opened 2021-12-16
2039 compute jump targets

fm-117:
Cancels and replaces #2052.

The cycle detection will be done outside of the parser but now we build a `JumpTarget` for each `PERFORM` or `GO TO` instruction. Targets for these jump instructions are stored in the resulting graph object and publicly accessible.

### DIFF
--- a/TypeCobol.Analysis/Graph/ControlFlowGraph.JumpTarget.cs
+++ b/TypeCobol.Analysis/Graph/ControlFlowGraph.JumpTarget.cs
@@ -1,0 +1,38 @@
+﻿using System.Collections.Generic;
+
+namespace TypeCobol.Analysis.Graph
+{
+	public partial class ControlFlowGraph<N, D>
+    {
+        /// <summary>
+        /// Represents the target of a jump instruction, that is either a PERFORM or a GO TO.
+        /// For a PERFORM instruction:
+        /// - the sentences are contiguous, it represents a single region of the Procedure Division
+        /// - if a section is included in procedures, it may not be fully targeted
+        ///   -> check the last procedure of the target against the last procedure of the section to see whether the section is fully covered or not
+        /// For a GO TO instruction:
+        /// - sentences may be non-contiguous if the GO TO has multiple targets which are non-contiguous.
+        /// </summary>
+        public class JumpTarget
+        {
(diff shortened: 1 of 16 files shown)


TypeCobol.Analysis/Graph/ControlFlowGraph.JumpTarget.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// For a PERFORM instruction:
+        /// - the sentences are contiguous, it represents a single region of the Procedure Division
+        /// - if a section is included in procedures, it may not be fully targeted
+        ///   -> check the last procedure of the target against the last procedure of the section to see whether the section is fully covered or not
+        /// For a GO TO instruction:
+        /// - sentences may be non-contiguous if the GO TO has multiple targets which are non-contiguous.
+        /// </summary>
+        public class JumpTarget
+        {
+            private readonly List<Sentence> _sentences;
```
smedilol, line 18: Why do we need to store `sentences` ?
The target of a `perform` or a `go to` is a paragraph, a section or a range of paragraph/section. So the references to `procedure(s)` should be enough.

smedilol (changes requested):

TypeCobol.Analysis/Graph/ControlFlowGraph.JumpTarget.cs
fm-117: The sentence accumulation mechanism in case of a range is not trivial, as we need to compute the sentence list for the perform block cloning, we compute the list here and store for re-use. Also the listing of procedure themselves is done through this accumulation process so we get both sentences and procedure in one go.

TypeCobol.Analysis/Graph/ControlFlowGraph.Procedure.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Collections.Generic;
+
+namespace TypeCobol.Analysis.Graph
+{
+    public partial class ControlFlowGraph<N, D>
```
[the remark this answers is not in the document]
fm-117: I tried at first to turn them into regular classes but I stopped because of the generic parameters. As nested classes they are implicitly generic.

PR #2103, opened 2021-12-24
2089 optimize cross checker

### LINKED ISSUE #2089: Optimize CrossChecker - visit of all nodes
The ASTVisitor is not the fastest way to iterate all node. Each node as to iterate over its children which means a lot of loop initialization.
It's actually faster to iterate over a simple list of all Node and call methods of the visitor directly (`BeginNode` and `VisitNode`).

In class `ProgramClassParserStep`, here's the method to optimize
```C#
public static void CrossCheckPrograms(SourceFile root, TemporarySemanticDocument temporarySemanticDocument)
        {
...
            foreach(var node in temporarySemanticDocument.**NewListOfNode**)
            {
                crossChecker.BeginNode(node);
                node.VisitNode(crossChecker);
            }
}
```
With this solution we lose the way to call `EndNode` method but as we don't use it, I think it's ok.

The gain of performance will be rather small, but given the small amount of work it's worth it.

fm-117:
Fix #2089

Here are the comparative results : [#2089_measurements.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/7773596/2089_measurements.xlsx)

Parsing times are not always improved but it seems better in incremental mode.


smedilol (commented):
Why change code in CGMV01-DeepVariables.cbl ?

fm-117:
@smedilol: CGMV01-DeepVariables.cbl had incorrect statements (not enough subscripts when accessing table elements). As the parser now checks this kind of error, the file would fail the performance test without change.

fm-117:
Here is a second performance report showing a greater improvement: [#2089_measurements_2.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/7808888/2089_measurements_2.xlsx)

fm-117:
Same data but including average parsing time:  [#2089.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/7814107/2089.xlsx).

smedilol:
Some measures seems inconsistent.
Your modifications only impact Incremental mode.
Test "Incremental_CGMV01-UseAlotOfTypes-WithProc-100Times" is slower with this PR, but the same test in FullParsing is faster. This should'nt be the case.

I have noticed that when you launch the same test twice, you can get different result (like 1 to 20ms).
Did you close any other application than VS before executing the tests ?
As we now have faster CPU, maybe we need to do more iteration of performance test.
Or maybe we need to also check the median.

smedilol (approved):

mayanje (dismissed):

fm-117:
I am keeping this PR open as we have not decided yet which optimizations have to be merged. The fixes on performance tests will be merged no matter what.

smedilol (approved):

smedilol:
With .Net 6 there is no gain anymore with this PR. Let's abandon it.

PR #2104, opened 2022-01-05
2090 optimize all diagnostics method

### LINKED ISSUE #2090: Optimize AllDiagnostic
Method AllDiagnostics is very slow compared to what it's supposed to do.

First it's because, `TokensDocument`, `ProcessedTokensDocument` and method OnlyCodeElementDiagnostics will each iterate over all lines to get differente kind diagnostics.
It's easy to optimize this by iterating only once over the lines and get directly all diagnostics.

Second and in link with  #2089, instead to use an ASTVisitor to iterate over all nodes, it's faster to iterate over a list of Node.

What to do:
1. integrate a performance measure of AllDiagnostics method in Performance tests (see `CompilationStats` class)
2. Optimize method AllDiagnostics with the 2 indications given above.

fm-117:
Fix #2090.

Follows #2103.

Performance report: [#2090.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/7814121/2090.xlsx)
On average, the parsing time is improved by 2%.


TypeCobol/Compiler/Nodes/Procedure.cs
```diff
@@ -153,11 +153,7 @@ public FunctionDeclaration(FunctionDeclarationHeader header) : base(header)
 
         public override bool VisitNode(IASTVisitor astVisitor)
         {
-            return astVisitor.Visit(this) && this.ContinueVisitToChildren(astVisitor, Profile.InputParameters)
```
smedilol, line 156: Why remove these lines ?

fm-117: To avoid duplicate visit of parameters. Parameters are created as nodes and now properly referenced as children of `FunctionDeclaration` so they are visited as part of the traversal of the tree and the parent don't need to call `ContinueVisitToChildren`. The `VisitNode` method is designed to visit the elements of the node itself not children nodes.

smedilol (approved):

mayanje (approved):

Issue #2112, opened 2022-01-18
Improve contextual data


fm-117:
To help identify bugs:
- add current `TypeCobolOptions` where available
- always add `Parser.Version`

smedilol:
@fm-117 Can we close this issue ?

fm-117:
Yes and no... Contextual data could be improved more but I think it is enough for now and we can always add info in other PRs. So ok to close it now.

PR #2114, opened 2022-01-18
WI #2112 Improve logging contextual data

fm-117:
- Ensure parser version is always traced in `LoggingSystem`
- Ensure command line args are traced
- Allow tracing of client info through the user-agent 'option'


smedilol (changes requested):

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -717,7 +717,7 @@ private void RefreshCustomSymbols()
                     LoadingIssueEvent(null, new LoadingIssueEvent() { Message = "An error occured while trying to load Intrinsics or Dependencies files." }); //Send notification to client
 
                 LoggingSystem.LogException(exception);
```
smedilol, line 719: Is it possible to pass CompilationProject.CompilationOptions here so the logger can also log the UserAgent?
```suggestion
                LoggingSystem.LogException(exception, CompilationProject.CompilationOptions);
```

smedilol (changes requested):
EDIT: you can ignore this comment.

TypeCobol.LanguageServer/Workspace.cs
fm-117: As discussed, user-agent is always logged in every trace as part of the command-line.

smedilol (approved):

mayanje (approved):

Issue #1874, opened 2021-01-28
New diagnostic CONFIGURATION SECTION in nested program


rooksdo:
### Description
Create a new diagnostic for CONFIGURATION SECTION declared in a nested program. 

**Code sample**
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Main.
       ENVIRONMENT DIVISION.
      *Ok
       CONFIGURATION SECTION.
       DATA DIVISION.
       PROCEDURE DIVISION.
      
           goback.
      
       IDENTIFICATION DIVISION.
       PROGRAM-ID. Nested.
       ENVIRONMENT DIVISION.
      *Ko CONFIGURATION SECTION is not allowed in nested programs
       CONFIGURATION SECTION.
       DATA DIVISION.
       PROCEDURE DIVISION.
           goback.
       END PROGRAM Nested.
      
       END PROGRAM Main.
```

fm-117:
Closed with #2131.

PR #2131, opened 2022-02-02
1874 new diag conf section in nested program

mayanje:
**What have been done**

A diagnostic is added to report an error in case presence of a CONFIGURATION SECTION in a nested program.
A test case is also adedd.

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
@@ -532,20 +532,20 @@ namespace TypeCobol.Compiler.Diagnostics
             CheckEndProgram(program);
             
             FormalizedCommentsChecker.CheckProgramComments(program);
-            if (program.IsNested)
-            {  // Check that a Nested Program cannot have a CONFIGURATION SECTION
-                IList<EnvironmentDivision> envDiv = program.GetChildren<EnvironmentDivision>();
-                if (envDiv.Count > 0)
-                {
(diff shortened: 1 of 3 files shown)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -532,6 +532,18 @@ public override bool Visit(Program program)
...
+                IList<EnvironmentDivision> envDiv = program.GetChildren<EnvironmentDivision>();
+                if (envDiv.Count > 0)
+                {
+                    IList<ConfigurationSection> confSection = envDiv[0].GetChildren<ConfigurationSection>();
+                    if (confSection.Count > 0)
+                    {
+                        DiagnosticUtils.AddError(confSection[0], "A Nested Program cannot have a CONFIGURATION SECTION.");
+                    }
+                }
+            }
```
fm-117, line 546: I'd rather override `Visit(ConfigurationSection configurationSection)` method and then use `GetProgramNode` to access the declaring program. This would avoid iterating over children and test type twice.

fm-117 (changes requested):

fm-117 (approved):

smedilol:
Hello,
Just a note about commit messages. Please check https://github.com/TypeCobolTeam/TypeCobol/wiki/CommitRules.
A commit message should help the reader the purpose of the change (the what) and not how it has been done.
When we create changelog for a release, if the commit message is not clear we have to check the associated issue and this can take times.
`WI #1874 Error diag if nested pgm contains configuration section`

smedilol (approved):

Issue #2127, opened 2022-01-31
Exception : REMARKS with parenthesis in comment


smedilol:
**Describe the bug**
A parenthesis in comment in a Remarks causes an Exception.

**To Reproduce**
(Type)Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
      *REMARKS. COPY=(
      *LM)            YMYCOPY 
      *                       ).
       END PROGRAM MyPGM.
      
```

**Expected behavior**
No error

**Solutions**
- Remove REMARKS parsing as we don't need it ?
- Fix this behavior and keep REMARKS parsing?

smedilol:
Fix with #1907 for Cobol.
For TypeCobol we expect to have only clean REMARKS directive.

Issue #2126, opened 2022-01-31
Exception with REMARKS that ends with * on col 72


smedilol:
**Describe the bug**
Copy remarks cause Exception when line ends with `*`.
![image](https://user-images.githubusercontent.com/13764482/151791402-1166693e-0b46-4667-9eac-6e854af308bb.png)

**To Reproduce**
Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPGM.
      *REMARKS. COPY=(                                                 *
      *).                                                              *
       END PROGRAM ABRSMRIU.
      
```
Expected (Type)Cobol code : 
No error

Solution:
- Remove REMARKS parsing as we don't need it ?
- Fix this behavior and keep REMARKS parsing?

Issue #2124, opened 2022-01-31
Comment-entry on multiples lines with REMARKS inside


smedilol:
**Describe the bug**
Comment-entry (like `Author` clause) can be on multiple lines. These lines can also contains comment line.
This works fine in our parser except when there is a REMARKS clause inside this comment-entry.

**To Reproduce**
Cobol code that cause the bug :
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       AUTHOR. Me.
      *REMARKS.  COPY=YMSG113
             .
       procedure division.
           goback
           .
       End Program DVZF0OSM.
```
Error on line before `procedure division`.

It works fine with:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       AUTHOR. Me.
      *Comment
             .
       procedure division.
           goback
           .
       End Program DVZF0OSM.
```

**Expected behavior**
No error.

**Solutions**
- Remove REMARKS parsing as we don't need it ?
- Fix this behavior and keep REMARKS parsing?

Issue #1907, opened 2021-03-08
Remove parsing of COPY REMARKS


rooksdo:
As a follow up to #1732, and as mentionned in that issue, parsing COPY REMARKS can be dropped.

AutoRemarks option can also be removed but the behavior must be kept.
The behavior must be like if option AutoRemarks is always true.

**Sample**
Here's a sample of EI Remarks
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
      *REMARKS. COPY=(
      *        YXXXTTT
      *        ).
       data division.
       working-storage section.
```

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
000200*REMARKS. COPY=(YXXXTTT).                               
      
       data division.
       working-storage section.
```

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
000480*REMARKS.  COPY=(YXXXTTT
000490*                YXXXTT2
000520*                ).                                               
       data division.
       working-storage section.
```

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
000370*REMARKS.                                                         
000380*         COPY=(                                                  
000390*              YXXXTTT
000440*              ).                                                 
       data division.
       working-storage section.
```

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
000370*REMARKS.                                                         
000420*         COPY=(YXXXTTT, YXXXTT2)    COPY DESCRIPTION                               
000430*         COPY=YXXXTT3                      COPY DESCRIPTION                                                                
       data division.
       working-storage section.
```

smedilol:
This is not mandatory for CobolEditor. It'd be better to postpone this issue.

PR #2132, opened 2022-02-03
WI #1907 Disable REMARKS directive parsing

fm-117:
This PR addresses:
- #2127
- #2126
- #2124
- #1639
- #1103

It is also an answer to #1907 however parsing of `REMARKS` directive could not be fully removed because it is required for Codegen when generating the clause into target Cobol85 source file.

**Changes:**
- `REMARKS` comments are parsed only in TypeCobol mode (meaning `IsCobolLanguage == false`)
- `AutoRemarks` option has been removed. When using Codegen, `REMARKS` clause is always generated using data from actual `COPY` clauses found in source
- `REMARKS` are not generated anymore for stacked programs

### DIFF
--- a/CLI/test/ressources/typedefCopy/output_expected/Client.rdz.tcbl
+++ b/CLI/test/ressources/typedefCopy/output_expected/Client.rdz.tcbl
@@ -1,6 +1,10 @@
-      *TypeCobol_Version:[[ParserVersion]]
+      *TypeCobol_Version:v0.0.0-local
        IDENTIFICATION DIVISION.
        PROGRAM-ID. Client.
+      *REMARKS. COPY=(
+      *        TESTCPY
+      *        ).
+                                                           
        data division.
        working-storage section.
       *01 MyType typedef strict public.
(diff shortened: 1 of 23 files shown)


CLI/test/ressources/typedefCopy/output_expected/Client.rdz.tcbl
```diff
@@ -1,6 +1,10 @@
-      *TypeCobol_Version:[[ParserVersion]]
+      *TypeCobol_Version:v0.0.0-local
```
smedilol, line 1: Error of replacing `[[ParserVersion]]` ?

smedilol (changes requested):

CLI/test/ressources/typedefCopy/output_expected/Client.rdz.tcbl
fm-117: Yes, caught in my own trap 🤦‍♂️ . Fixed in 5b1bdc5e987ef71bf14cc00bb63f28b8798f0e38.

smedilol (approved):

smedilol (dismissed):

mayanje (dismissed):

smedilol (approved):

PR #2158, opened 2022-03-07
WI #2151 Add support of Workspace Project concept.

mayanje:
**What have been done**

- A WorkspaceProject class has been added to represent the association between a client project and a Parser Compilation project
- A WorkspaceProjectStore class has been added to store all instances of WorkspaceProject, and to store all opened document instances.
- A custom LSP notification "DidOpenProjectTextDocumentNotification" whose parameters extends "DidOpenTextDocumentParams" is added in order to be able to directly opening a document knowing its target Workspace Project key.
- A custom notification "DidChangeProjectConfigurationNotification" is added to create or update a Workspace Project key instance.

PR #2197, opened 2022-04-11
2184 lsp implement refresh copy

### LINKED ISSUE #2184: New LSP notification refresh one COPY
Provide a new LSP notification to refresh only one COPY.
This new notification is also here to optimize the refresh mechanism from LSP (`TypeCobol.LanguageServer.Workspace#ScheduleRefresh`)
Instead of calling `Workspace#ScheduleRefresh`, a new method must be created.

This new method will 
- [ ] Remove only the COPY to refresh from the cache.
- [ ] Refresh only document(s) that use this copy
- [ ] Optimize re-parsing mechanism #2159
   - Not sure if this issue will provide a significative gain as the slower step of the parser is CodeElement.
   - #2159 can also apply to `TypeCobol.LanguageServer.Workspace#ScheduleRefresh`

**Technical**
Notif parameters:
- COPY to refresh
- projectKey

How to do the refresh of one copy:
- Remove only the COPY (+ its variants) to refresh from the cache.
   - Create a new class for COPY cache.
      - Move method `ClearImportedCompilationDocumentsCache` into this class.
      - Create a new method to remove only one COPY (+its variants) from the cache.
      - Think about variant. See `cacheKey` in `CompilationProject#Import`
      - Caller of `ClearImportedCompilationDocumentsCache` will now access the copy cache from `CompilationProject`.
- Refresh only document(s) that use this copy
   - In `Workspace` (or `WorkspaceProjectStore` I didn't check #2158), iterate over all documents and use `CompilationUnit.CopyTextNamesVariations` to detect these cases.

**How to test ?**
LanguageServerRobot is not able to create a change in a COPY while the test is running as the content of the COPY is not managed by LSP messages.
So what to do:
1. Keep LSR running between 2 tests.
   - Create a first test to parse a source
   - Update the COPY file locally
   - Play another test which send `refresh single copy` notif.
2. Create a manual dedicated test without LanguageServerRobot ?
3. Extend LanguageServerRobot to be able to insert local script between LSP messages?
   - This local script will change the content of the COPY before the new refresh notification will be sent

The test can use 2 copys:
- Parse the document with these 2 copys.
- Change the 2 copy locally.
- Send refresh single copy notification for only 1 of the 2 copy.
- Reparse the document.
- Check that the document use the new version of the refresed copy but not the other one.

fm-117:
Fixes #2184.

Using the `didChangeWatchedFiles` notification, the server can now react to changes made on copies.

This PR is based on #2158.


TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
```diff
@@ -36,5 +38,17 @@ public bool Contains(string name)
             return Array.BinarySearch(_cpyCopyNames, name.ToUpper()) >= 0;
         }
     }
+
+    /// <summary>
+    /// Helper methods
+    /// </summary>
+    public static class CopyNameMapFileExtensions
```
smedilol, line 45: Why not put these methods directly in the related classes ?

TypeCobol.LanguageServer/WorkspaceProjectStore.cs
```diff
@@ -115,51 +103,54 @@ private void AddProject(WorkspaceProject project)
...
-        }
-
-        /// <summary>
-        /// Refresh all documents in all WorkspaceProject instances
-        /// </summary>
-        internal void RefreshOpenedFiles()
-        {
-            this.DefaultWorkspaceProject.RefreshOpenedFiles();
-            foreach (var wksProject in _workspaceProjects.Values)
+            if (!_namedProjects.TryAdd(projectKey, workspaceProject))
```
smedilol, line 114: I think it's better to keep the dedicated method `AddProject(workspaceProject);` here.

smedilol (changes requested):

TypeCobol/Compiler/Preprocessor/CopyNameMapFile.cs
fm-117: This allow to factorize the `Contains` logic

TypeCobol.LanguageServer/WorkspaceProjectStore.cs
fm-117: But I kept the assertion and the exception. `AddProject` is not used elsewhere so I deleted it.

mayanje (approved):

mayanje (approved):

smedilol (approved):

PR #2204, opened 2022-04-19
2167 support procedure name in area a

mayanje:
**What have been done**

All the checks bellow are performed only if the format is the COBOL format and not the Freestyle format, and if the scanner is in a PROCEDURE DIVISION.

- The scanner has been adapted to reconized SectionParagraphName or a PartialCobolWord if the token is in Area A
- The scanner has been adapted to reconized UserDefinedWord or a PartialCobolWord if  the token is after a PERFORM in a position that expects a UserDefinedWord
- Position after a PERFORM keyword are tracked using a hard coded automata as property in a MultilineScanState instance.
- The PreprocessorStep is used to preprocess the fact that, if before the TIMES keyword in a PERFORM state the UserDefinedWord is in fact an integer, then the UserDefinedWord is replaced by IntegerLiteral.

### DIFF
--- a/TypeCobol/Compiler/Scanner/Scanner.cs
+++ b/TypeCobol/Compiler/Scanner/Scanner.cs
@@ -986,22 +986,22 @@ namespace TypeCobol.Compiler.Scanner
                         return ScanUntilDelimiter(startIndex, TokenType.FORMALIZED_COMMENTS_VALUE, "%>>>");
                 }
             }
-
-            // --- Main switch ---
+            
+            // Special treatment in case of Section or Paragraph name.
             // IN Cobol Format:
-            // IF we are in the PROCEDURE DIVISION and the current character is not a space and  THEN:
-            //      IF (we are inside AREA A or we are inside a PERFORM Statement state taht allow a procedure name)
-            //          and the next character can be the start of a UserDefineWord THEN
+            // IF we are in the PROCEDURE DIVISION and the current character is not a space THEN:
+            //      IF ((we are inside AREA A or we are inside a PERFORM Statement state that allows a procedure name)
+            //          and the next character can be the start of a UserDefineWord) THEN
             //              TRY to scan a UserDefinedWord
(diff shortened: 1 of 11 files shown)


TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -2213,7 +2276,7 @@ private Token ScanSqlCodeOrExecStatementTextOrExecSqlInclude(int startIndex)
             return new Token(TokenType.ExecStatementText, startIndex, endIndex, tokensLine);
         }
 
-        private Token ScanKeywordOrUserDefinedWord(int startIndex)
+        private Token ScanKeywordOrUserDefinedWord(int startIndex, bool tryProcedureName)
```
fm-117, line 2279: What `tryProcedureName` does ? As I understand it, it restricts to `SectionParagraphName` the returned token type (or partial cobol word). So the naming is not clear.

TypeCobol/Compiler/Scanner/ScannerUtils.cs
```diff
@@ -55,4 +55,22 @@ public static class ScannerUtils
 
         return false;
     }
+
+    /// <summary>
+    /// Get the CobolFormatArea in which the given index is into.
+    /// </summary>
+    /// <param name="index">The index to found the corresponding CobolFormatAreas</param>
+    /// <returns> the CobolFormatArea</returns>
+    public static CobolFormatAreas GetAreaFromIndex(int index)
```
fm-117, line 64: I think this should return `TextAreaType` instead of `CobolFormatAreas`. In fact `CobolFormatAreas` is not really an enmu but more of a collection of const int. Also the test for `Indicator` is incorrect, it should be

```csharp
        if (index < (int)CobolFormatAreas.Begin_A)
            return CobolFormatAreas.Indicator;
```

fm-117 (changes requested):
I did not understand the whole mechanism with the new automatas used in this PR, so I'll have to read it again. My remarks are on the code itself for now, not the solution.

smedilol:
See new PR #2215

Issue #2078, opened 2021-11-15
Replace : manage parenthesis in partial cobol word (simple case)


smedilol:
**Describe the bug**

**CHILD OF #2077**

Some of our users use this kind of replace:
```cobol
        replace ==(REJ)== by ==REJ==.
```

Cobol allow these characters as separators of partial Cobol word: `:` `(` and `)` (see #2077).
But in this issue we'll only focus on this case:

## 2 separators
| **Replace**                | **Text surrounded by spaces** | **Result** |
| -------------------------- | ---------------------------- | ---------- |
| **\==(REJ)== by ==REJ==**  | (REJ)-COD                    | REJ-COD    |
| **\==:REJ:== by ==REJ==**  | (REJ)-COD                    | No match   |
| **\==:REJ:== by ==REJ==**  | :REJ:-COD                    | REJ-COD    |
| **\==:REJ:== by ==REJ==** | A:REJ:-COD                 | AREJ-COD   |
| **\==:R EJ:== by ==REJ==** | A:R  EJ:-COD                 | AREJ-COD   |
| **\==(R EJ)== by ==REJ==** | A(R  EJ)-COD                 | AREJ-COD   |
| **\==(REJ)== by ==REJ==** | A(REJ)-COD                 | AREJ-COD   |

**Expected behavior**
Handle all these replace.

PR #2211, opened 2022-04-21
WI #2190 Handle Picture Character String replacement as Partial Cobol Word replacement

### LINKED ISSUE #2190: Bug in picture string based on a REPLACE operation
**Describe the bug**
Our parser does not handle replace operations over PictureCharacterString

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPgm.
       REPLACE ==toReplace== BY ==50==.
       data division.
       working-storage section.
       01 var PIC X(toReplace).
       END PROGRAM MyPgm.
```

The parser will fail and return `Line 6[19,30] <27, Error, Syntax> - Syntax error : Missing '(' or ')' in PICTURE string` because the REPLACE did not happen.

**Expected behavior**
No error and correct picture definition (check DataDef and VariableSymbol)

fm-117:
Fixes #2190.

Another hack in the REPLACE mechanism to handle simple case of PIC replacements which are now considered at text level instead of token level.

Note that the preprocessor test `PgmCopyReplacing` is now wrong on its last part. This is because the test was previously based on 2 bugs (#2078 and #2190). With this current fix, we now observe effects of #2078 and also the fact that in a multiple replace instruction the parser should not attempt to replace the same token twice (see #2212).


smedilol (dismissed):

mayanje (approved):

fm-117:
Replaced by #2465.

PR #2230, opened 2022-05-03
WI #2212 Do not apply subsequent REPLACE operations if one matched prior

### LINKED ISSUE #2212: Bug in REPLACE/REPLACING directives with multiple operations
**Describe the bug**
According to some results with IBM compiler, when a token is replaced by one replace operation, other following operations of the same directive should not be considered for this token (i.e. a word can be replaced zero or one time at most).

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPgm.
       data division.
       working-storage section.
       01 group1.
       REPLACE ==(01)== BY ==(01)==
                      == 01 == BY == 05 ==.
       01 var1 PIC X(01).
       END PROGRAM MyPgm.
```

On IBM compiler the resulting `var1` has a `X(01)` picture definition, however our parser will have a `X(05)` picture definition.
This test is already in our code, see [PgmCopyReplacing.cbl](https://github.com/TypeCobolTeam/TypeCobol/blob/e830129f01afff13cba87793e19e21c1a93e1728/TypeCobol.Test/Parser/Preprocessor/CopyTestFiles/PgmCopyReplacing.cbl)

**Expected behavior**
The second replace operaton should have no effect on the picture string.
- check Cobol specs to explain this behavior
- check for every 4 types of replace operations
  - for example test [CopyReplace4Colon](https://github.com/TypeCobolTeam/TypeCobol/blob/e830129f01afff13cba87793e19e21c1a93e1728/TypeCobol.Test/Parser/Programs/Cobol85/CopyReplace4Colon.rdz.cbl)  seems to contradict this behavior...

fm-117:
Fixes #2212.


fm-117:
Replaced by #2466.

PR #2238, opened 2022-05-23
chore: Fix build error when cloning for first time

gromatluidgi:
TypeCobol - Build error when cloning

When cloning the repository for the first time and try to build it; many errors occur because of the compiling process requiring unknown generated code insides TypeCobol.csproj.

How to reproduce error:
- Clone a fresh repository
- Try to build CLI.csproj first

The proposed solution adds a condition that ensures any generated sources are linked in the build process only if they exists. This fix should not have any impacts on:
- Other development environment
- CI/CD Build process (presumption)

Tests status: OK

### DIFF
--- a/TypeCobol/TypeCobol.csproj
+++ b/TypeCobol/TypeCobol.csproj
@@ -36,34 +36,34 @@
     </Reference>
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="$(TC_AntlrGeneratedFiles)\CodeElementsBaseListener.cs">
+    <Compile Include="$(TC_AntlrGeneratedFiles)\CodeElementsBaseListener.cs" Condition="Exists('$(TC_AntlrGeneratedFiles)\CodeElementsBaseListener.cs')">
       <Link>Compiler\Parser\Generated\CodeElementsBaseListener.cs</Link>
     </Compile>
-    <Compile Include="$(TC_AntlrGeneratedFiles)\CodeElementsBaseVisitor.cs">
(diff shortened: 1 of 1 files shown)


fm-117:
Hi @gromatluidgi. We'll look into your fix and integrate it after testing it in our own environment.
Thanks for the input.

fm-117:
Tested on VS 2109 after a fresh clone, I don't see any difference in behavior from our current csproj. How is it supposed to behave ?
The build errors stems from VS which keeps a model of the project in-memory, this model is built at project opening. When the project is opened for the first time, the generated files do not exist and consequently are not included in the model. After generating them (through a first build), we still have to unload/reload TypeCobol project or close/reopen the solution so as to update the in-memory project model.
Adding the condition should not change this, is it different on VS 2022 ?

gromatluidgi:
Hi @fm-117 ,

Thanks for your reply which let me confirm that's my solution is not relevant and does not offer the guarantes to suppress the issue.

**Scenario (24/05/22)**:
MinorIssueBuild

Development environment:
- Windows 11
- Visual Studio 2022

I've done more investigation and various tests, I would say this issue is not coming from the project configuration.  By doing many iterations of git clone and build, the issue i've describe first is not recurrent:
- Sometimes building solution a second time, seem to **fix the issues**. 
- Cleaning the solution and try to rebuild only TypeCobol, **trigger the issues**.

Other elements make me think that the problem is from my System/IDE :
- VBCSCompiler process can remain in the background even after closing/reopening/closing my IDE => This spawn many .NET Host process => I**nconsistent  IDE state**
- Reloading VS solution when the issue occurs, allow me to build the project successfully => **Inconsistent IDE state** 

However I can notice that's **compiling TypeCobol project first** after cloning the repository, never trigger build issues.

There are too many scenario that's may be caused by my environment, to guarantee the pertinence of the fix.

**I do apologise for the inconvenience, next time I will open issue before doing any proposal !**

Issue #2225, opened 2022-04-27
SQL Variables


mayanje:
**Variables**

A variable in an SQL statement specifies a value that can be changed when SQL statement is executed.
All types of DB2 variables are described here: https://www.ibm.com/docs/en/db2-for-zos/12?topic=elements-variables.

Usually variables are syntactically: simple identifiers, or qualified identifiers. But there are also special syntax like for Host variables and others variable types.

fm-117:
@smedilol @laabidiHend We should review the design first. I don't understand the purpose of `SqlSymbol` here. Also this is another case where we have to decide how storage areas will be handled. This includes the definition of new SymbolTypes for SQL. Currently we only have `SqlIdentifier` which is as generic as it could be so it certainly can be improved.

PR #2239, opened 2022-05-24
WI #2225 Define SQL Column Name

### LINKED ISSUE #2227: SQL Column names
**[Column names](https://www.ibm.com/docs/en/db2-for-zos/12?topic=elements-column-names)**
Column names can be seen as variables used to declare colum name, specify/identify the name of a column or specify the values of column in SQL statements or expressions.

laabidiHend:
Fixes #2227.


TypeCobol/Compiler/Sql/VisitorPattern.cs
```diff
@@ -29,6 +29,7 @@ public interface ISqlVisitor
         bool Visit([NotNull] UnpackedRowSelection unpackedRowSelection);
         bool Visit([NotNull] TableOrViewAllColumnsSelection tableOrViewAllColumnsSelection);
         bool Visit([NotNull] SqlExpression sqlExpression);
+        bool Visit([NotNull] SqlObject sqlObject);
```
fm-117, line 32: This is unnecessary. For `SqlObject`, the visit is handled through the `BeginSqlObject`, `EndSqlObject` and `VisitSqlObject` methods.
For classes that derive from `SqlObject`, a dedicated visit method is added.

TypeCobol/Compiler/Sql/Model/SqlColumnName.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace TypeCobol.Compiler.Sql.Model
+{
+    public class SqlColumnName : SqlExpression
+    {
+        public SqlColumnName(SymbolReference symbol)
+        {
+            this.Symbol = symbol;
+        }
+
+        private SymbolReference Symbol { get; }
```
fm-117, line 13: must be public

TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -8298,7 +8298,8 @@ table_reference: single_table_or_view_reference;
 single_table_or_view_reference: tableOrViewOrCorrelationName correlation_clause?;
 correlation_clause: SQL_AS? (correlation_name=UserDefinedWord) new_column_names?;
 new_column_names: LeftParenthesisSeparator new_column_name (SQL_CommaSeparator new_column_name)* RightParenthesisSeparator;
-new_column_name: UserDefinedWord;
+new_column_name: column_name;
+column_name:UserDefinedWord;
```
fm-117, line 8302: There is no need for a new definition. You can keep the previous definitions or rename `new_column_name` rule to `column_name` if you want to reuse 'column_name' in future rules.

TypeCobol/Compiler/CodeElements/Expressions/SymbolType.cs
```diff
@@ -41,7 +41,8 @@ public enum SymbolType
         // Used when the type of the symbol has not yet been resolved
         TO_BE_RESOLVED,
         // For SQL
-        SqlIdentifier
+        SqlIdentifier,
+        ColumnName
```
fm-117, line 45: Ok this anticipates for #2249, but we can keep it in this PR.

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -283,5 +281,15 @@ private DatetimeConstant CreateDatetimeConstant(CodeElementsParser.Datetime_cons
 
             return null;
         }
+        private SqlColumnName CreateSqlColumnName(CodeElementsParser.Column_nameContext context)
+        {
+            if (context.UserDefinedWord() != null)
```
fm-117, line 286: Why adding the test now ? In previous code there was no nullity check. To improve the code, you should add tests for the from clause with missing column names and see how ANTLR handles it. Then you will find how to adapt the builder. The goal is to have a syntax error from ANTLR and no exception in the builder.

For example, in ExecSqlWithFromClause.rdz.cbl, try this statement:

```
SELECT * FROM t_table as new_name (c1,,)
```

The syntax here is wrong, so we should one or more errors from ANTLR and the builder should create a list with one new column name (for 'c1').

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -157,12 +157,7 @@ private CorrelationClause CreateCorrelationClause(CodeElementsParser.Correlation
 
             if (context.new_column_names() != null)
             {
-                List<SqlColumnName> newColumnNamesList = new List<SqlColumnName>();
-                foreach (var columnName in context.new_column_names().new_column_name())
-                {
-                    SqlColumnName newColumnName = CreateSqlColumnName(columnName.column_name());
-                    newColumnNamesList.Add(newColumnName);
-                }
+                List<SqlColumnName> newColumnNamesList = (from columnName in context.new_column_names().column_name() where columnName.Diagnostics == null select CreateSqlColumnName(columnName)).ToList();
```
laabidiHend, line 160: I used " columnName.Diagnostics == null" to check  if there is an error from ANTLR .
I tried to use "columnName!=null" but it fails because with emty columnName (in the test )i found  with "<MissingUserDefinedWord>" as  userDefinedWord.

TypeCobol/Compiler/Sql/Scanner/SqlScanner.cs
```diff
@@ -42,7 +42,7 @@ private enum Sign
         /// <returns>True if yes, no otherwise</returns>
         private static bool IsSqlKeywordPart(char c)
         {
-            return c == '-' || char.IsLetter(c);
+            return c == '_' || c == '-' || char.IsLetterOrDigit(c);
```
laabidiHend, line 45: UserDefinedWord can contain digits and underscore '_'

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
fm-117: As discussed directly, I think testing on the literal is cleaner

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
fm-117: The test can now be safely removed as we check the absence of diagnostics before calling the method.

TypeCobol/Compiler/Sql/Model/SqlColumnName.cs
```diff
@@ -0,0 +1,25 @@
...
+        protected override void DumpContent(TextWriter output, int indentLevel)
+        {
+            DumpProperty(output, nameof(Symbol), Symbol, indentLevel);
+        }
+
+        protected override bool VisitSqlObject(ISqlVisitor visitor)
+        {
+            return visitor.Visit(this) ;
+        }
+    }
```
fm-117, line 24: Could you please move this into SqlExpression.cs file (and delete this file) ? We already have constants defined there, so I think we'll put all expressions in the same file.

fm-117 (changes requested):

fm-117 (approved):

smedilol (approved):

PR #2242, opened 2022-05-27
WI #2196 Integrate lockTableStatement

### LINKED ISSUE #2196: SQL LOCK TABLE
**Description**

Integrate the [LOCK TABLE](https://www.ibm.com/docs/en/db2-for-zos/12?topic=statements-lock-table) Sql statement in the parser.

This implies:

- Ensure that all Keywords are recognized by the sql scanner
- To write the ANTLR grammar
- To write the CUP grammar
- Building the LockTable Statement CodeElement
- Building the LockTable Node
- Creating the corresponding Sql Object
   - Each options of the statement must be a property (ex: Partition, IsShare,...)
   - Each token that the form the LockTable statement must be stored in the SqlObject.
   - The _table name_ must be a StorageArea that holds  a SymbolReference
- Making corresponding CodeElement, Node and SqlObject classes visitable
- Write test unit

Prerequisites:
by #2183 the _integer_ constant is produced by the scanner.

laabidiHend:
Fixes #2196


laabidiHend:
I changed :
        private static bool IsSqlKeywordPart(char c)
        {
            return c == '-' || char.IsLetterOrDigit(c);
        }
Because tableName which is userDefinedWord in sql can contain digits and i think that also the all userDefinedWord they can contain letters and digits .

Also in the test i put a wrong statement (without MODE) to verifie that the error message exists .

TypeCobol.Test/Utils/Comparator.cs
```diff
@@ -533,6 +533,14 @@ public override bool Visit(TruncateStatement truncateStatement)
 
                 return true;
             }
+            public override bool Visit(LockTableStatement lockTableStatement)
+            {
+                _writer.WriteLine($"line{lockTableStatement.Line}: {nameof(lockTableStatement)}");
```
fm-117, line 538: Please add a space between 'line' and the actual number, correct the nameof expression. We want the name of the type not the parameter. Check other methods as a guide.

TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -8300,6 +8301,10 @@ correlation_clause: SQL_AS? (correlation_name=UserDefinedWord) new_column_names?
 new_column_names: LeftParenthesisSeparator new_column_name (SQL_CommaSeparator new_column_name)* RightParenthesisSeparator;
 new_column_name: UserDefinedWord;
 
+lockTableStatement: SQL_LOCK SQL_TABLE tableOrViewOrCorrelationName (SQL_PARTITION IntegerLiteral)? SQL_IN (share | exclusive) sql_mode;
+share: ({ string.Equals(CurrentToken.Text, "SHARE", System.StringComparison.OrdinalIgnoreCase) }? KeywordSHARE=UserDefinedWord);
+exclusive: ({ string.Equals(CurrentToken.Text, "EXCLUSIVE", System.StringComparison.OrdinalIgnoreCase) }? KeywordEXCLUSIVE=UserDefinedWord);
+sql_mode: ({ string.Equals(CurrentToken.Text, "MODE", System.StringComparison.OrdinalIgnoreCase) }? KeywordMODE=UserDefinedWord);
```
fm-117, line 8307: Is 'mode' already defined in grammar ? If the name is available I'd rather have a `mode` rule instead of `sql_mode`.

TypeCobol/Compiler/Parser/CodeElementBuilder/CobolCodeElementBuilder.cs
```diff
@@ -2280,5 +2280,10 @@ public override void EnterTruncateStatement([NotNull] CodeElementsParser.Truncat
             Context = context;
             CodeElement = _sqlCodeElementBuilder.CreateTruncateStatement(context);
         }
-    }
+        public override void EnterLockTableStatement([NotNull] CodeElementsParser.LockTableStatementContext context)
+        {
+            Context = context;
+            CodeElement = _sqlCodeElementBuilder.CreateLockTableStatement(context);
+        }
+	}
```
fm-117, line 2288: Indent problem

TypeCobol/Compiler/Sql/CodeElements/Statements/LockTableStatement.cs
```diff
@@ -0,0 +1,43 @@
...
+            IsPartitionInteger = isPartitionInteger;
+        }
+
+        public override bool VisitCodeElement(IASTVisitor astVisitor)
+        {
+            return base.VisitCodeElement(astVisitor) && astVisitor.Visit(this)
+                                                     && this.ContinueVisitToChildren(astVisitor, Mode,
+                                                         IsPartitionInteger)
+                                                     && astVisitor.SqlVisitor != null
+                                                     && astVisitor.SqlVisitor.ContinueVisit(PartitionId,Table);
```
fm-117, line 40: spacing

TypeCobol/Compiler/Sql/CodeElements/Statements/LockTableStatement.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    public class LockTableStatement : SqlStatementElement
+    {
+        public TableViewCorrelationName Table { get; }
+
+        [CanBeNull] public SqlConstant PartitionId { get; }
+
+        public SyntaxProperty<LockMode> Mode { get; }
+
+        public SyntaxProperty<bool> IsPartitionInteger { get; }
```
fm-117, line 22: I don't understand this property. The absence of partition id is already visible when `PartitionId` is null. Compared to `IsImmediate` in `TruncateStatement` for example, this is different. In TRUNCATE statement, the keyword IMMEDIATE changes the behavior of the statement so we capture this info using a `SyntaxProperty<bool>`. Here the relevant information is the partition identifier (including its absence) and this is fully captured through the `PartitionId` property.
Please remove the `IsPartitionInteger` property and adapt code accordingly.

fm-117 (changes requested):

TypeCobol/Compiler/Sql/Model/SqlExpression.cs
```diff
@@ -68,6 +69,10 @@ public virtual SqlConstantType Type
 
         public Token Literal { get; }
 
+        protected override void DumpContent(TextWriter output, int indentLevel)
+        {
+            DumpProperty(output, nameof(Literal.LiteralValue), Literal.LiteralValue, indentLevel);
```
fm-117, line 74: Ok this means that for SqlConstant, we only output the literal value. It's incomplete (no info on the token positioning for example) but it's fine.

fm-117 (approved):

smedilol (approved):

PR #2268, opened 2022-06-22
2248 integrate alter sequence

### LINKED ISSUE #2248: Alter sequence
Description

The [Alter sequence](https://www.ibm.com/docs/en/db2-for-zos/12?topic=statements-alter-sequence) statement changes the attributes of a sequence at the current server. Only future values of the sequence are affected by the ALTER SEQUENCE statement.

This implies:

- To write the ANTLR grammar
- To write the CUP grammar
- Building the  Alter sequence Statement CodeElement
- Building the  Alter sequence Node
- Creating the corresponding Sql Object
- Making corresponding CodeElement, Node and SqlObject classes visitable
- Write test unit

laabidiHend:
Fixes #2248


laabidiHend:
I tested ```  ALTER SEQUENCE RESTART INCREMENT BY 5.2 ``` it doesn't contain SequenceName which shoud generate an error.
But here the ```RESTART``` which is a Keyword is considered a sequence name.
I think the problem can be resolved if we add ```RESTART``` in the list of TokenType so it will not be considered as a userDefinedWord.

fm-117:
Concerning the RESTART keyword: we could indeed add the keyword and this should solve the problem. Another solution is to instruct ANTLR to reject sequence names that could be mistaken with contextual keywords. You have an example of that in allocateStatement in Cobol2002CodeElement.g4:

https://github.com/TypeCobolTeam/TypeCobol/blob/84cdec76606463fb0bab24786acc48d6ec2820f2/TypeCobol/AntlrGrammar/Cobol2002CodeElements.g4#L56-L60

Regarding the non-duplicate clause, why do you say it cannot be verified in a checker ? A checker has basically the same amount of information than a builder. Remember it is a static method that receives the constructed CodeElement and the ANTLR context so it has access to all the data the builder has.

laabidiHend:
I thought in the checker we should just test with the constructed CodeElement received from the builder, that's why i proposed to verify the  non-duplicate clause in the builder.
So it's clear now , i can also check with the context in the checker.
I add the keyword to resolve the problem ? 

> Concerning the RESTART keyword: we could indeed add the keyword and this should solve the problem. Another solution is to instruct ANTLR to reject sequence names that could be mistaken with contextual keywords. You have an example of that in allocateStatement in Cobol2002CodeElement.g4:
> 
> https://github.com/TypeCobolTeam/TypeCobol/blob/84cdec76606463fb0bab24786acc48d6ec2820f2/TypeCobol/AntlrGrammar/Cobol2002CodeElements.g4#L56-L60
> 
> Regarding the non-duplicate clause, why do you say it cannot be verified in a checker ? A checker has basically the same amount of information than a builder. Remember it is a static method that receives the constructed CodeElement and the ANTLR context so it has access to all the data the builder has.

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -365,5 +365,74 @@ public LockTableStatement CreateLockTableStatement(CodeElementsParser.LockTableS
...
+            SyntaxProperty<bool> cycle = null;
+            SqlConstant cacheSize = null;
+            SyntaxProperty<bool> ordered = null;
+            if (context.sequence_name != null)
+            {
+                sequenceName = new SymbolReference(new AlphanumericValue((Token) context.sequence_name),
+                    SymbolType.SqlIdentifier);
+            }
+
+            foreach (var alterSequenceClause in context.alterSequenceClause())
```
fm-117, line 387: I see now what you mean about checking the non-duplicate clauses in the builder. I think we should test that for each clause the associated SqlConstant or SyntaxProperty is not null before defining it so as to keep only the first clause and identify the duplicates. We could then pass the duplicates to the checker so it can create the diagnostic.

fm-117 (changes requested):

fm-117:
@laabidiHend, usually all checks are performed in checkers while builder keeps creating objects no matter what. Here is a special case where it is more efficient to examine the structure in the builder and keep some info about the syntax used. At any rate, the diagnostics must be created in the checker.

I'll take care of adding the RESTART keyword.

TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -8312,6 +8313,22 @@ sqlGotoHostLabel: (SQL_GOTO | SQL_GO SQL_TO) ColonSeparator? hostLabel=UserDefin
...
+cache: ({ string.Equals(CurrentToken.Text, "CACHE", System.StringComparison.OrdinalIgnoreCase) }? KeywordCACHE=UserDefinedWord);
+
+alterSequenceStatement: SQL_ALTER SQL_SEQUENCE (sequence_name=UserDefinedWord) alterSequenceClause alterSequenceClause*;
+alterSequenceClause: restartClause | incrementClause | minValueClause | maxValueClause | cacheClause | cycle | SQL_ORDER | noClauses;
+restartClause: restart ( SQL_WITH (numeric_constant=DecimalLiteral))? ;
+incrementClause: sqlIncrement SQL_BY (numeric_constant=DecimalLiteral);
+minValueClause: minvalue (numeric_constant=DecimalLiteral);
+maxValueClause: maxvalue (numeric_constant=DecimalLiteral);
+cacheClause: cache (numeric_constant=IntegerLiteral);
+noClauses: SQL_NO (minvalue | maxvalue | SQL_ORDER | cycle | cache);
```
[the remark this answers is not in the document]
laabidiHend: But this one dosn't word .If i write a sentence with "no maxValue" i get an error .The grammar is always waiting for a minValue after the NO because MinValue is the first one . That's why i changed that .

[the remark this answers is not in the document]
fm-117: Ok maybe ANTLR does not work properly with this grammar using a lot of contextual keyword. However you have to test for the presence or absence of 'NO' keyword so as to build more precise SyntaxProperty objects. A 'NO' keyword will translate into a non-null false SyntaxProperty<bool>.

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -391,5 +391,150 @@ public LockTableStatement CreateLockTableStatement(CodeElementsParser.LockTableS
...
+            MinValue,
+            MaxValue,
+            Cycle,
+            Ordered,
+            Cache,
+            Restart,
+            Increment
+        }
+
+        public AlterSequenceStatement CreateAlterSequenceStatement(
```
fm-117, line 405: Ok this method is complex with the duplicate handling. Here is what I suggest

```csharp
        public AlterSequenceStatement CreateAlterSequenceStatement(CodeElementsParser.AlterSequenceStatementContext context)
        {
            var duplicates = new List<ParserRuleContextWithDiagnostics>();

            var clauseSet = new HashSet<AlterSequenceClauseType>();

            SyntaxProperty<bool> cycle = null;
            SqlConstant minValue = null;

            foreach (var alterSequenceClauseContext in context.alterSequenceClause())
            {
                if (alterSequenceClauseContext.minValueClause() != null)
                {
                    var minValueClause = alterSequenceClauseContext.minValueClause();
                    if (clauseSet.Add(AlterSequenceClauseType.MinValue))
                    {
                        System.Diagnostics.Debug.Assert(minValue == null);
                        minValue = new SqlConstant((Token)minValueClause.numeric_constant);
                    }
                    else
                    {
                        System.Diagnostics.Debug.Assert(minValue != null);
                        duplicates.Add(minValueClause);
                    }
                }
                else if (alterSequenceClauseContext.cycle() != null)
                {
                    var cycleClause = alterSequenceClauseContext.cycle();
                    if (clauseSet.Add(AlterSequenceClauseType.Cycle))
                    {
                        System.Diagnostics.Debug.Assert(cycle == null);
                        cycle = new SyntaxProperty<bool>(true, ParseTreeUtils.GetFirstToken(cycleClause));
                    }
                    else
                    {
                        System.Diagnostics.Debug.Assert(cycle != null);
                        duplicates.Add(cycleClause);
                    }
                }
                // else if (...)
                // {
                //     ...
                // }
                else if (alterSequenceClauseContext.noClauses() != null)
                {
                    if (alterSequenceClauseContext.noClauses().minvalue() != null)
                    {
                        var noMinValueClause = alterSequenceClauseContext.noClauses().minvalue();
                        if (clauseSet.Add(AlterSequenceClauseType.MinValue))
                        {
                            System.Diagnostics.Debug.Assert(minValue == null);
                        }
                        else
                        {
                            System.Diagnostics.Debug.Assert(minValue != null);
                            duplicates.Add(noMinValueClause);
                        }
                    }
                    else if (alterSequenceClauseContext.noClauses().cycle() != null)
                    {
                        var noCycleClause = alterSequenceClauseContext.noClauses().cycle();
                        if (clauseSet.Add(AlterSequenceClauseType.Cycle))
                        {
                            System.Diagnostics.Debug.Assert(cycle == null);
                            cycle = new SyntaxProperty<bool>(false, ParseTreeUtils.GetTokenFromTerminalNode(alterSequenceClauseContext.noClauses().SQL_NO()));
                        }
                        else
                        {
                            System.Diagnostics.Debug.Assert(cycle != null);
                            duplicates.Add(noCycleClause);
                        }
                    }
                    // else if (...)
                    // {
                    //    ...
                    // }
                }
            }
        }
```

Could you see how it looks with the whole set of options and try to look for factorizations  (use local function for examples)

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -7,19 +8,38 @@ namespace TypeCobol.Compiler.Sql.CodeElements
...
                 {
                     DiagnosticUtils.AddError(savepointStatement,
-                        "Invalid savepoint-name, it must not begin with 'SYS'.",
-                        context);
+                        "Invalid savepoint-name, it must not begin with 'SYS'.", context);
                 }
             }
         }
     }
+    public static class IntegrateAlterSequenceStatementChecker
```
fm-117, line 24: - naming `AlterSequenceStatementChecker`
- Checker classes should be internal, let's use this PR to change it for SavepointStatementChecker and AlterSequenceStatementChecker

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -7,19 +8,38 @@ namespace TypeCobol.Compiler.Sql.CodeElements
...
+            {
+                DiagnosticUtils.AddError(alterSequenceStatement, clauseType + " can not be specified more than once",
+                    context);
+            }
+
+            if (alterSequenceStatement != null && alterSequenceStatement.MinValue == null &&
+                alterSequenceStatement.MaxValue == null &&
+                alterSequenceStatement.Ordered == null && alterSequenceStatement.Restart == null &&
+                alterSequenceStatement.IncrementValue == null && alterSequenceStatement.Cycle == null &&
+                alterSequenceStatement.NoCache == false)
```
fm-117, line 39: Better test on CacheValue, if we decide to distinguish between explicit NO CACHE and implicit

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
fm-117: Ok for the visibility, now please change the name.

fm-117:
Here is a way to simplify the code for the builder:

```csharp
            foreach (var alterSequenceClauseContext in context.alterSequenceClause())
            {
                if (alterSequenceClauseContext.restartClause() != null)
                {
                    SetOption(alterSequenceClauseContext.restartClause(), AlterSequenceClauseType.Restart, c => restart = new SyntaxProperty<bool>(true, ParseTreeUtils.GetFirstToken(c)));
                    if (alterSequenceClauseContext.restartClause().numeric_constant != null)
                    {
                        restartValue = new SqlConstant((Token) alterSequenceClauseContext.restartClause().numeric_constant);
                    }
                }

                [...]

                if (alterSequenceClauseContext.noClauses() != null)
                {
                    var noKeyword = alterSequenceClauseContext.noClauses().SQL_NO();
                    if (alterSequenceClauseContext.noClauses().cycle() != null)
                    {
                        SetOption(noKeyword, AlterSequenceClauseType.Cycle, c => cycle = new SyntaxProperty<bool>(false, ParseTreeUtils.GetTokenFromTerminalNode(noKeyword)));
                    }

                    [...]
                }
            }
```

with the local function

```csharp
            void SetOption(IParseTree clause, AlterSequenceClauseType type, Action<IParseTree> set)
            {
                if (clauseSet.Add(type))
                {
                    set(clause);
                }
                else
                {
                    duplicates.Add(clause);
                }
            }
```

@laabidiHend ask me if you have difficultes when rewriting the method using this pattern.

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -474,5 +476,149 @@ private SymbolReference CreateTableOrAliasName(CodeElementsParser.TableOrAliasNa
...
+                    SetOption(alterSequenceClauseContext.restartClause(), AlterSequenceClauseType.Restart,
+                        c => restart = new SyntaxProperty<bool>(true, ParseTreeUtils.GetFirstToken(c)));
+                    if (alterSequenceClauseContext.restartClause().numeric_constant != null)
+                    {
+                        restartValue =
+                            new SqlConstant((Token) alterSequenceClauseContext.restartClause().numeric_constant);
+                    }
+                }
+
+                if (alterSequenceClauseContext.incrementClause() != null)
```
fm-117, line 524: No, this is inconsistent. The action supplied to SetOption does nothing.

```csharp
if (alterSequenceClauseContext.incrementClause() != null && alterSequenceClauseContext.incrementClause().numeric_constant != null)
{
    SetOption(alterSequenceClauseContext.incrementClause(), AlterSequenceClauseType.Increment, c => incrementValue =
                            new SqlConstant((Token) c.numeric_constant););
}
```

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -474,5 +476,149 @@ private SymbolReference CreateTableOrAliasName(CodeElementsParser.TableOrAliasNa
...
+                        SetOption(noKeyword, AlterSequenceClauseType.Cache,
+                            c => ParseTreeUtils.GetTokenFromTerminalNode(noKeyword));
+                    }
+
+                }
+            }
+
+            var alterSequenceStatement = new AlterSequenceStatement(sequenceName, restart, restartValue, incrementValue,
+                minValue, maxValue, cycle, cacheSize, ordered);
+            AlterSequenceStatementChecker.OnCodeElement(alterSequenceStatement, duplicates, !clauseSet.Any(),
```
fm-117, line 607: replace `!clauseSet.Any()` with `clauseSet.Count == 0` which is faster

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
fm-117: This is the same for the following options, so change them too

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -474,5 +476,149 @@ private SymbolReference CreateTableOrAliasName(CodeElementsParser.TableOrAliasNa
...
+                    }
+                }
+
+                if (alterSequenceClauseContext.SQL_ORDER() != null)
+                {
+                    SetOption(alterSequenceClauseContext.SQL_ORDER(), AlterSequenceClauseType.Ordered,
+                        c => ordered = new SyntaxProperty<bool>(true, ParseTreeUtils.GetFirstToken(c)));
+                }
+
+                if (alterSequenceClauseContext.noClauses() != null)
```
fm-117, line 579: NO MINVALUE and NO MAXVALUE are unhandled. You still have to call SetOption so as to add the clause in the set, just use an empty action

```csharp
SetOption(noKeyword, AlterSequenceClauseType.MinValue, c => {});
```

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +19,29 @@ public static class SavepointStatementChecker
...
+    internal static class AlterSequenceStatementChecker
+    {
+        public static void OnCodeElement(AlterSequenceStatement alterSequenceStatement,
+            List<IParseTree> duplicatedClauses, bool emptyClauseSet,
+            CodeElementsParser.AlterSequenceStatementContext context)
+        {
+            foreach (var clauseType in duplicatedClauses)
+            {
+                DiagnosticUtils.AddError(alterSequenceStatement,
+                    clauseType.GetChild(0).GetText() + " can not be specified more than once",
```
fm-117, line 37: `clauseType.GetChild(0).GetText()` won't work when NO clause is passed.
I think we need to change the grammar, keeping the noClauses branch but adding named variants for each

```
noClauses: noMinvalueClause | noMaxvalueClause | noOrderClause | noCycleClause | noCacheClause;
noMinvalueClause: SQL_NO minValue;
[...]
```

This allow to capture the whole clause in a single context object that can be passed to SetOption. The NO keyword will still be used for syntax properties and the keyword following it will be used in the diagnostic message.

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +19,29 @@ public static class SavepointStatementChecker
...
-                        context);
+                        "Invalid savepoint-name, it must not begin with 'SYS'.", context);
                 }
             }
         }
     }
+
+    internal static class AlterSequenceStatementChecker
+    {
+        public static void OnCodeElement(AlterSequenceStatement alterSequenceStatement,
```
fm-117, line 30: We only need to add checks on numeric literals (and corresponding unit test) and we'll be done

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -40,8 +39,77 @@ internal static class AlterSequenceStatementChecker
...
+                    DiagnosticUtils.AddError(alterSequenceStatement,
+                        "The IncrementValue decimal should be without non-zero digits in the right of the decimal point",
+                        context);
+                }
+            }
+
+            if (alterSequenceStatement.CacheSize != null)
+            {
+                var tokenValue = alterSequenceStatement.CacheSize.Literal.LiteralValue;
+                if (tokenValue.Type != LiteralTokenValueType.Integer)
```
laabidiHend, line 106: In the spec RestartValue, MinValue, MaxValue and IncrementValue should be decimal without non-zero digits in the right of the decimal point but CacheSize shouyld be declared as integer . It's already declared in the grammer as integer .

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +18,120 @@ public static class SavepointStatementChecker
...
+            }
+
+            if (emptyClauseSet)
+            {
+                DiagnosticUtils.AddError(alterSequenceStatement,
+                    "At least one option among RESTART, INCREMENT, MINVALUE, MAXVALUE, CYCLE, CACHE or ORDER must be specified",
+                    context);
+            }
+
+            if (alterSequenceStatement.MinValue != null)
```
fm-117, line 47: You have to factorize code.

```csharp
void CheckIsInteger(string optionName, Token optionValue)
{
    [...]
    - retrieve literal value
    - check type
    - check digits if decimal
}
```

you can also return the value as `double` if you have further checks to do.

instead of copy pasting

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +18,120 @@ public static class SavepointStatementChecker
...
+                        "The IncrementValue decimal should be without non-zero digits in the right of the decimal point",
+                        context);
+                }
+            }
+
+            if (alterSequenceStatement.CacheSize != null)
+            {
+                var tokenValue = alterSequenceStatement.CacheSize.Literal.LiteralValue;
+                var integerLiteral = (IntegerLiteralTokenValue) tokenValue;
+                if (integerLiteral.HasSign)
```
fm-117, line 120: No. `HasSign` property indicates whether the sign has been explicitly written, to check whether the number is negative or not, you have to test the value. That's why it will become redundant with the check below.

fm-117 (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/ExecSqlWithAlterSequence.rdz.cbl
```diff
@@ -0,0 +1,35 @@
...
+              INCREMENT BY 5.0           
+              ALTER SEQUENCE org_seq INCREMENT BY 5.8
+              ALTER SEQUENCE org_seq2 MINVALUE 6.0 MAXVALUE 5.0
+              ALTER SEQUENCE org_seq3
+             END-EXEC.
+       PROCEDURE DIVISION.
+         
+           GOBACK
+           .
+       END PROGRAM DVZZMFT3.
```
smedilol, line 35: I've added `SPECIAL-NAMES.  DECIMAL-POINT IS COMMA.` because it's the default on our system. 
It's easier to test SQL syntax this way.
```suggestion
       IDENTIFICATION DIVISION.
       program-id. DVZZMFT3.
       ENVIRONMENT DIVISION.
       CONFIGURATION SECTION.
       SOURCE-COMPUTER. IBM-370.
       SPECIAL-NAMES.  DECIMAL-POINT IS COMMA.
       data division.
       working-storage section.
           EXEC SQL
                INCLUDE SQLCA
           END-EXEC.

       procedure division.

           EXEC SQL ALTER SEQUENCE org_seq RESTART
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq RESTART WITH 100
           END-EXEC.
      *KO
      *DSNH336I DSNHSM5S THE SCALE OF THE DECIMAL NUMBER MUST BE ZERO
           EXEC SQL ALTER SEQUENCE org_seq RESTART WITH 100,1
           END-EXEC.

           EXEC SQL ALTER SEQUENCE org_seq INCREMENT BY 5
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq INCREMENT BY 5
           END-EXEC.

           EXEC SQL ALTER SEQUENCE org_seq NO MINVALUE
           END-EXEC.
      *KO
      *DSNH637I DSNHSM5S DUPLICATE "MINVALUE" KEYWORD OR CLAUSE.
           EXEC SQL ALTER SEQUENCE org_seq2 MINVALUE 6 MINVALUE 8
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq2 MINVALUE 6,0
           END-EXEC.
      *KO
      *DSNH336I DSNHSM5S THE SCALE OF THE DECIMAL NUMBER MUST BE ZERO
           EXEC SQL ALTER SEQUENCE org_seq2 MINVALUE 6,1
           END-EXEC.

           EXEC SQL ALTER SEQUENCE org_seq NO MAXVALUE
           END-EXEC.
      *KO
      *DSNH628I DSNHSM5S THE CLAUSES ARE MUTUALLY EXCLUSIVE
           EXEC SQL ALTER SEQUENCE org_seq2 MAXVALUE 25 NO MAXVALUE
           END-EXEC.
      *KO
      *DSNH336I DSNHSM5S THE SCALE OF THE DECIMAL NUMBER MUST BE ZERO
           EXEC SQL ALTER SEQUENCE org_seq2 MAXVALUE 7,1
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq2 MAXVALUE 7
           END-EXEC.


           EXEC SQL ALTER SEQUENCE org_seq2 NO CYCLE
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq2 CYCLE
           END-EXEC.

      *KO
      *DSNH846I DSNHSM5S LINE 83 COL 38  INVALID SPECIFICATION OF AN
      * IDENTITY COLUMN OR SEQUENCE OBJECT SEQUENCE ORG_SEQ2.
      * REASON CODE = 4
      *DSNH336I DSNHSM5S THE SCALE OF THE DECIMAL NUMBER MUST BE ZERO
           EXEC SQL ALTER SEQUENCE org_seq2 CACHE 100,1
           END-EXEC.
      *KO
      *The cache size should be positive
      *DSNH104I DSNHSM5S ILLEGAL SYMBOL"-". SOME SYMBOLS THAT MIGHT BE
      *LEGAL ARE: +<INTEGER>
           EXEC SQL ALTER SEQUENCE org_seq2 CACHE -20
           END-EXEC.
      *KO
      *The cache size should be greater than 2
      *DSNH846I DSNHSM5S  INVALID SPECIFICATION OF AN IDENTITY COLUMN
      * OR SEQUENCE OBJECT SEQUENCE ORG_SEQ2.
      * REASON CODE = 4
           EXEC SQL ALTER SEQUENCE org_seq2 CACHE 1
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq2 NO CACHE
           END-EXEC.


           EXEC SQL ALTER SEQUENCE org_seq2 ORDER
           END-EXEC.
           EXEC SQL ALTER SEQUENCE org_seq2 NO ORDER
           END-EXEC.


      *Combination
           EXEC SQL ALTER SEQUENCE org_seq2 NO ORDER RESTART
              INCREMENT BY 5
           END-EXEC.

      *KO
      *The maxValue must be greater than or equal to the minimum value
           EXEC SQL ALTER SEQUENCE org_seq2 MINVALUE 6 MAXVALUE 5
           END-EXEC.

      *KO
      *At least one option among RESTART, INCREMENT, MINVALUE, MAXVALUE,
      * CYCLE, CACHE or ORDER must be specified
      *DSNH104I DSNHPARS ILLEGAL
      * SYMBOL"<END-OF-STATEMEN". SOME SYMBOLS THAT MIGHT BE LEGAL ARE:
      *NOMAXVALUE NOMINVALUE NOORDER CYCLE NOCYCLE NOCACHE
           EXEC SQL ALTER SEQUENCE org_seq3
           END-EXEC.

           goback.
       end program DVZZMFT3.
```

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +18,120 @@ public static class SavepointStatementChecker
...
+
+            if (alterSequenceStatement.MinValue != null)
+            {
+                var tokenValue = alterSequenceStatement.MinValue.Literal.LiteralValue;
+                if (tokenValue.Type != LiteralTokenValueType.Decimal) return;
+                var decimalLiteral = (DecimalLiteralTokenValue) tokenValue;
+                bool isInteger = Math.Abs(decimalLiteral.Number % 1) < double.Epsilon;
+                if (!isInteger)
+                {
+                    DiagnosticUtils.AddError(alterSequenceStatement,
```
smedilol, line 55: The IBM error is `THE SCALE OF THE DECIMAL NUMBER MUST BE ZERO`.
I'd prefer `Only digits '0' are allowed after decimal point`.

This apply to other `decimal should be without non-zero digits in the right of the decimal point` messages.

TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -8317,11 +8318,28 @@ sqlGotoHostLabel: (SQL_GOTO | SQL_GO SQL_TO) ColonSeparator? hostLabel=UserDefin
 hostVariable: ColonSeparator mainVariable=UserDefinedWord ((indicator)? ColonSeparator indicatorVariable=UserDefinedWord)?;
 indicator: ({string.Equals(CurrentToken.Text, "INDICATOR", System.StringComparison.OrdinalIgnoreCase) }? KeywordINDICATOR=UserDefinedWord);
 
+sqlConstant: SQL_DecimalFloatingPointLiteral | SQL_BinaryStringLiteral | SQL_GraphicStringLiteral | AlphanumericLiteral | HexadecimalAlphanumericLiteral | IntegerLiteral | DecimalLiteral | FloatingPointLiteral | datetime_constant | SQL_NULL;
```
smedilol, line 8321: sqlConstant is not used in `alterSequenceStatement`. It was missing from a previous PR ?

smedilol (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -474,5 +476,143 @@ private SymbolReference CreateTableOrAliasName(CodeElementsParser.TableOrAliasNa
...
+                    if (alterSequenceClauseContext.noClauses().SQL_ORDER() != null)
+                    {
+                        SetOption(noKeyword, AlterSequenceClauseType.Order,
+                            c => ordered = new SyntaxProperty<bool>(false,
+                                ParseTreeUtils.GetTokenFromTerminalNode(noKeyword)));
+                    }
+
+                    if (alterSequenceClauseContext.noClauses().cache() != null)
+                    {
+                        SetOption(noKeyword, AlterSequenceClauseType.Cache,c=> {});
```
smedilol, line 583: We should also store the information about nocache, no minvalue and no maxvalue.
Can you add these properties to AlterSequenceStatement:
```
public SyntaxProperty<bool> Cache { get; }
public SyntaxProperty<bool> MinValue { get; }
public SyntaxProperty<bool> MaxValue { get; }
```
and se them here here.

smedilol (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
fm-117: Yes, here they would have to be set to false and you'll  have to set them to true in upper part of the method too.

TypeCobol/AntlrGrammar/CobolCodeElements.g4
laabidiHend: Cache size is an integer-constant  .

fm-117: @smedilol is right, since the 'sqlConstant' ANTLR rule is not used, it should be removed from this PR. However the builder will create `SqlConstant` instances for the different numeric literals used here.

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
```diff
@@ -474,5 +476,162 @@ private SymbolReference CreateTableOrAliasName(CodeElementsParser.TableOrAliasNa
...
+                if (alterSequenceClauseContext.minValueClause() != null)
+                {
+                    SetOption(alterSequenceClauseContext.minValueClause(), AlterSequenceClauseType.MinValue,
+                        c => hasMinValue = new SyntaxProperty<bool>(true, ParseTreeUtils.GetFirstToken(c)));
+                    if (alterSequenceClauseContext.minValueClause().numeric_constant != null)
+                    {
+                        minValue = new SqlConstant((Token) alterSequenceClauseContext.minValueClause()
+                            .numeric_constant);
+                    }
+                }
```
fm-117, line 543: For MINVALUE, MAXVALUE and CACHE, I would rather create both properties only if both keyword and value are present.

```csharp
if (alterSequenceClauseContext.minValueClause() != null && alterSequenceClauseContext.minValueClause().numeric_constant != null)
{
    SetOption([...]);
}
```

and in the action provided to SetOption, set both properties (boolean and constant).

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +18,109 @@ public static class SavepointStatementChecker
...
+                    context);
+            }
+
+            if (alterSequenceStatement.MinValue != null)
+            {
+                CheckIsInteger(alterSequenceStatement.MinValue.Literal);
+                if (!isInteger)
+                {
+                    DiagnosticUtils.AddError(alterSequenceStatement,
+                        " In minValue only digits '0' are allowed after decimal point",
```
fm-117, line 54: Factorize the diagnostic creation directly into `CheckIsInteger`, this will avoid problems like inconsistency in letter case or wrong messages.
For example:
- 'In minValue only digits '0' are allowed after decimal point' vs 'In RESTART value only digits '0' are allowed after decimal point', different letter case, always use uppercase for keywords appearing in diagnostic messages
- 'In RESTART value only digits '0' are allowed after decimal point' vs 'The INCREMENT value only digits '0' are allowed after decimal point', what happened ? The second message with 'The' is grammatically wrong.

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +18,109 @@ public static class SavepointStatementChecker
...
+                    DiagnosticUtils.AddError(alterSequenceStatement,
+                        "Minimum value for cache size is 2",
+                        context);
+                }
+
+            }
+            void CheckIsInteger(Token optionValue)
+            {
+                var tokenValue = optionValue.LiteralValue;
+                if (tokenValue.Type != LiteralTokenValueType.Decimal) return;
```
fm-117, line 119: Ok because this method expects only DecimalLiteral or IntegerLiteral. Please add an assert before the test.

`Debug.Assert(tokenValue.Type == LiteralTokenValue.Integer || tokenValue.Type == LiteralTokenValue.Decimal);`

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementCheckers.cs
```diff
@@ -16,10 +18,109 @@ public static class SavepointStatementChecker
...
+                var decimalLiteral = (DecimalLiteralTokenValue)tokenValue;
+                isInteger = Math.Abs(decimalLiteral.Number % 1) < double.Epsilon;
```
fm-117, line 121: <= double.Epsilon

fm-117 (changes requested):

TypeCobol/Compiler/Sql/CodeElements/SqlCodeElementBuilder.cs
laabidiHend: I didn't change that yet . I have a question .

TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/ExecSqlWithAlterSequence.rdz.cbl
fm-117: I commented DECIMAL POINT IS COMMA, tests will be changed globally when dealing with #2307.

TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/ExecSqlWithAlterSequence.rdzSQL.txt
```diff
@@ -0,0 +1,366 @@
+﻿--- Diagnostics ---
+Line 16[21,25] <27, Error, Syntax> - Syntax error : In RESTART value, only digits '0' are allowed after decimal point RuleStack=codeElement>alterSequenceStatement,  OffendingSymbol=[21,25:ALTER]<SQL_ALTER>
```
smedilol, line 2: RuleStack can be removed from these diagnostics

smedilol (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/ExecSqlWithAlterSequence.rdzSQL.txt
fm-117: The RuleStack does not appear in the message displayed in IDz. Besides, we already use this overload of `DiagnosticUtils.AddError` in other CodeElement checkers.

smedilol (approved):

PR #2287, opened 2022-07-12
WI #2213 Use correct ScanState when testing for debugging mode in continuation group

### LINKED ISSUE #2213: NullReferenceException in IsDebugLineActive
**Describe the bug**
A log showed that a NullReferenceException may be thrown by `Scanner.IsDebugLineActive`. Looking at the code it means that we encountered a line with either `SourceText` null or no `ScanState`.

`SourceText` may be null for an empty line, so we have to fix the method. Check also whether `ScanState` can be null or not.
- include `initialScanState` as new parameter for `TokensLine` constructor to ensure the scan state is always valued ?

fm-117:
Fixes #2213


TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -96,7 +96,7 @@ public class Scanner : AbstractScanner
 
             // Comment line => return only one token with type CommentLine
             // Debug line => treated as a comment line if debugging mode was not activated
-            if (textLine.Type == CobolTextLineType.Comment || (tokensLine.Type == CobolTextLineType.Debug && !IsDebugLineActive(tokensLine)))
+            if (textLine.Type == CobolTextLineType.Comment || (tokensLine.Type == CobolTextLineType.Debug && !IsDebuggingModeActive(tokensLine.ScanState, tokensLine.SourceText)))
```
smedilol, line 99: If I understand correctly, the scanState of following lines of a continuation group is null, that's why we need to use only the scanState of the first line of a continuation group.
If so, I'd rather make sure to always have a non null scanState on all lines of a continuation group to avoid this bug in other situations.

fm-117: No it's not that.

When scanning a continuation group, the scanState of the line is set after the test on the debugging mode. Here we can use directly the property of the line when calling the method but in the continuation group case, we pass the instance of scanState that would be set later on the tokens line.

fm-117: Added a comment for that special case

smedilol (approved):

Issue #2295, opened 2022-07-18
LSP : CopyFolders of default configuration must also be used for file opened in a project


smedilol:
**Describe the bug**
With LSP:
When you open a source that references a default copy (like SQLCA), then this default COPY is not found if the source is open with notification "typecobol/workspace/didOpenProjectTextDocument".

**To Reproduce**
Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       program-id. MyPGM.
       data division.
       working-storage section.
           EXEC SQL
                INCLUDE SQLCA
           END-EXEC.
       end program MyPGM.
```

**Expected behavior**
No error, the default copy must be found.

**Technical**
In `WorkspaceProject.Configure`, we need to add all `Workspace.Configuration.CopyFolders` if they are not present yet in the specified new `copyFolders` parameter.

**How to test automatically**
Can be done with LSR tests as we reference copy ?
Otherwise it will be a manual tests on our editor side.

fm-117:
Fixed by #2296.

PR #2296, opened 2022-07-18
WI #2295 LSP Add default copy folder for document opened in a project

smedilol:
Default copy folder is the the last element like document opened in the default project.

### DIFF
--- a/TypeCobol.LanguageServer/WorkspaceProject.cs
+++ b/TypeCobol.LanguageServer/WorkspaceProject.cs
@@ -176,7 +176,7 @@ namespace TypeCobol.LanguageServer
             var newCopyFolders = oldCopyFolders.ToList();
             if (copyFolders != null)
             {
-                newCopyFolders = copyFolders.Union(new []{ Workspace.DefaultCopyFolder }, StringComparer.OrdinalIgnoreCase).ToList();
+                newCopyFolders = copyFolders.Union(new []{ Workspace.DefaultCopyFolder }, StringComparer.InvariantCultureIgnoreCase).ToList();
                 if (AreCopyFoldersDifferent(oldCopyFolders, newCopyFolders))
                 {
                     updateCopyFolders = true;


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -615,6 +615,12 @@ private bool StopDocumentBackgroundCompilation(DocumentContext contextToStop)
             return false;
         }
 
+        public string GetDefaultCopyFolder()
```
fm-117, line 618: Method should be static, it could even be turned into a static property

TypeCobol.LanguageServer/WorkspaceProject.cs
```diff
@@ -174,10 +174,17 @@ public void Configure(DocumentFormat format, TypeCobolOptions options, AnalyzerP
...
-            if (copyFolders != null && AreCopyFoldersDifferent(oldCopyFolders, copyFolders))
+            if (copyFolders != null)
             {
-                updateCopyFolders = true;
-                newCopyFolders = copyFolders;
+                newCopyFolders = new List<string>
+                {
+                    this._workspace.GetDefaultCopyFolder()
+                };
+                newCopyFolders = copyFolders.Union(newCopyFolders).ToList();
```
fm-117, line 183: Maybe add a comparer in union operation to use case-insensitive comparison ?

fm-117 (changes requested):

smedilol:
@fm-117  As discussed, see #2298 for the better solution to handle this.

TypeCobol.LanguageServer/Workspace.cs
smedilol: Done 32ff807dc5b9cead7138edb14175dcbdd732b990

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -615,6 +615,15 @@ private bool StopDocumentBackgroundCompilation(DocumentContext contextToStop)
             return false;
         }
 
+        public static string GetDefaultCopyFolder
```
fm-117, line 618: naming DefaultCopyFolder, you can even turn it into a static readonly field

TypeCobol.LanguageServer/WorkspaceProject.cs
```diff
@@ -174,10 +174,13 @@ public void Configure(DocumentFormat format, TypeCobolOptions options, AnalyzerP
             var oldCopyFolders = GetCopyFolders();
             bool updateCopyFolders = false;
             var newCopyFolders = oldCopyFolders.ToList();
-            if (copyFolders != null && AreCopyFoldersDifferent(oldCopyFolders, copyFolders))
+            if (copyFolders != null)
             {
-                updateCopyFolders = true;
-                newCopyFolders = copyFolders;
+                newCopyFolders = copyFolders.Union(new []{ Workspace.GetDefaultCopyFolder }, StringComparer.OrdinalIgnoreCase).ToList();
```
fm-117, line 179: You have to use a local variable  before assigning `newCopyFolders` because the value is used even when the boolean `updateCopyFolders` is false.
Also I use InvariantCultureIgnoreCase for paths and OrdinalIgnoreCase for source code names but I'm not sure it is required.

fm-117 (changes requested):

TypeCobol.LanguageServer/WorkspaceProject.cs
smedilol: if `AreCopyFoldersDifferent(oldCopyFolders, newCopyFolders)` is false, then we can keep the value of newCopyFolders because it's the same as `oldCopyFolders.ToList()`.
Done for InvariantCultureIgnoreCase : ac4aca9eee24c42ad79ca7c90ff65fc5ddc8059a.

fm-117 (approved):

PR #2304, opened 2022-07-22
WI #2303 Always create CodeElement for SUBTRACT / DIVIDE / MULTIPLY

### LINKED ISSUE #2303: Always create CodeElement for SUBTRACT / DIVIDE / MULTIPLY
Child of #1835.

Single keywords or invalid statements won't give any CodeElement which is required for completion.
The goal of this issue is to ensure that CE are always created and that diagnostics are correctly reported for SUBTRACT, DIVIDE and MULTIPLY statements.

fm-117:
Fixes #2303.


TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
```diff
@@ -344,11 +342,36 @@ internal static IList<DocumentChange<ICodeElementsLine>> ParseProcessedTokensLin
...
+                            {
+                                if (codeElement.Diagnostics == null)
+                                {
+                                    codeElement.Diagnostics = new List<Diagnostic>(diagnosticsToReport);
+                                }
+                                else
+                                {
+                                    foreach (var diagnostic in diagnosticsToReport)
+                                    {
+                                        codeElement.Diagnostics.Add(diagnostic);
```
smedilol, line 356: Optional suggestion:
Maybe change `codeElement.Diagnostics` to List and use `AddRange` here.
I don't think we'll ever use another implementation of IList for `CodeElement.Diagnostics`.

smedilol (approved):

TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
fm-117: done

smedilol (approved):

PR #2324, opened 2022-08-16
WI #2318 Add diagnostic for Code Elements altering ScanState

### LINKED ISSUE #2318: Create error for syntax that change the ScanState inside a copy
**Describe the bug**
Temporary solution for #2316 to state that we do not support some syntax.

Create an error :
1. if `DataDivisionHeader` or `ProcedureDivisionHeader` CodeElement are inside a copy
2. if `SourceComputerParagraph` with `DebuggingMode=true` is inside a copy
3. if an `Exec SQL` is inside a COPY and one of its its children is not in the **same** copy, this also includes the `END-EXEC`.

Other syntaxes that can change the scan state inside a copy are considered too marginal or impossible, to be detected (formalized comment, special names, replace)

**To Reproduce**
Cobol code that cause the bug : (if any)
```cobol
TODO
```

**Technical**
Create diagnostic in CodeElementChecker for 1. and 2..

Create diagnostic in CrossChecker for 3. To detect that it's the same COPY, check the CopyDirective.

**How to test automatically**
Will be edited by a team member.
Only if our current automated tests doesn't handle this.

fm-117:
Fixes #2318


fm-117:
Unfortunately this new check is incompatible with SM106A from NIST test suite. We could:
- use Warning severity (currently implemented in this PR)
- exclude SM106A from tests

smedilol (approved):

fm-117:
@smedilol re-requesting your review after changing the severity and adapting the NIST test.

smedilol (approved):

PR #2325, opened 2022-08-17
WI #373 Add file name resolution

### LINKED ISSUE #373: File name resolution
**Describe the bug**
Resolve file name.

**To Reproduce**
Cobol code that cause the bug :
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.    MyPGM.
       ENVIRONMENT DIVISION.
       INPUT-OUTPUT SECTION.
       FILE-CONTROL.
           SELECT F-SOR003         ASSIGN TO   SOR003
      *OK can reuse same name as file for file status
                                   FILE STATUS F-SOR003.
       DATA DIVISION.
       FILE SECTION.

       FD  F-SOR003
           BLOCK CONTAINS 0
           RECORDING MODE F.
       01  Group1.
           05 Var1                   PIC X(06).

       PROCEDURE DIVISION.
      *OK qualify variable name with file name
           MOVE "A" TO Var1 IN F-SOR003
      *KO "F-SOR003"was defined as a type that was invalid in this
      *context.  The statement was discarded.
           move "A" to F-SOR003
      *Ok
           write Group1 in F-SOR003
      *Ok
           call "foo" using F-SOR003
           .
       END PROGRAM MyPGM.
```

**Expected behavior**
No error.

**Technical**
`FD` and `SD` files declared in `FILE SECTION` must be added to the `SymbolTable.Variables` of the appropriate scope.
TODO : check if we need a new SymbolType for SymbolReference.

Idea:
- `FD` and `SD` files declared in `FILE SECTION`  must be `SymbolDefinition` instead of `SymbolReference`
- FileName in SelectClause can be kept as `SymbolDefinition`
- Make FileDescriptionEntry inherits from DataDefinitionEntry or a sublass of it
- FileDescriptionEntryNode inherits from DataDefinition or a subclass of it
- In ProgramClassBuilder add FileDescriptionEntryNode to SymbolTable:  SymbolTable.AddVariable
- Check SemanticDomain. It must be a VariableSymbol.
- Level 01 following the file name is a child of it.

See 7cfc663138b03d323817cb5501c9c015c31314ff. Ignore `SymbolDefinition` changed to `SymbolInformation`.

**How to test automatically**
Standard pgm test.

fm-117:
Fixes #373.

However type checking is not supported yet. As the most important part of the issue is to resolve correctly file names, I suggest to merge as is and add  specific checks later in a separate issue.

Checking type compatibilty in MOVE instructions when dealing with file descriptions also has an impact on typing system so it may be more complex than it appears...


TypeCobol/AntlrGrammar/CobolCodeElements.g4
```diff
@@ -1227,7 +1227,7 @@ fileControlEntry:
 // p130: If the file connector referenced by file-name-1 is an external file connector, all file-control entries in the run unit that reference this file connector must have the same specification for the OPTIONAL phrase.
 
 selectClause:
-    SELECT OPTIONAL? fileNameDefinition;
+    SELECT OPTIONAL? fileNameReference;
```
smedilol, line 1230: After second thought I think we can keep `fileNameDefinition` here.
Because, if later we want a SymbolTable for file declarations we will need a "SymbolDefinition".

Then for `fileDescriptionEntry` we could use a combination of reference and definition:
```suggestion
    (FD | SD) fileNameReferenceAndDataNameDefinition
```
Defined like that:
`fileNameReferenceAndDataNameDefinition: UserDefinedWord;`

Class `FileDescriptionEntry` still inherits from `DataDefinitionEntry` and has 2 properties:
- public SymbolReference FileName { get; set; }
- public SymbolDefinition DataName { get; set; };
What do you think ?

fm-117: I swapped between reference and definition because I think we should have a single instance of `SymbolDefinition` for the same thing. If we split into 2 definitions, one for the `FileName` and one for the `DataName` it is consistent. However, in order to compile, those two names must be iddentical right ? Where should this be checked ?

fm-117: Done

Issue #2185, opened 2022-03-30
NullReferenceException in SymbolTable.GetParagraph


fm-117:
**Describe the bug**
Exception in `SymbolTable.GetParagraph`, more precisely NullReferenceException is thrown in predicate defined here:

https://github.com/TypeCobolTeam/TypeCobol/blob/4b256f395eb8ac6a3a7fd3f70a52ba5d6d0483a9/TypeCobol/Compiler/CodeModel/SymbolTable.cs#L774

Full StackTrace is:
```
   à TypeCobol.Compiler.CodeModel.SymbolTable.<>c.<GetParagraph>b__37_2(Paragraph p)
   à System.Collections.Generic.List`1.FindAll(Predicate`1 match)
   à TypeCobol.Compiler.CodeModel.SymbolTable.GetParagraph(SymbolReference symbolRef, Section sectionNode)
   à TypeCobol.Compiler.CodeModel.SymbolTable.GetSectionOrParagraph(SymbolReference target, Section callerNodeSection)
   à TypeCobol.Compiler.Diagnostics.SectionOrParagraphUsageChecker.ResolveParagraphOrSection(Node callerNode, SymbolReference target, Section currentSection)
   à TypeCobol.Compiler.Diagnostics.CrossCompleteChecker.Visit(PerformProcedure performProcedureNode)
   à TypeCobol.Compiler.Nodes.PerformProcedure.VisitNode(IASTVisitor astVisitor)
   à TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor(IASTVisitor astVisitor)
   à TypeCobol.Compiler.Nodes.Node.VisitChildrens(IASTVisitor astVisitor, Boolean continueVisit, IEnumerable`1 childrenNodes)
   à TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor(IASTVisitor astVisitor)
   à TypeCobol.Compiler.Nodes.Node.VisitChildrens(IASTVisitor astVisitor, Boolean continueVisit, IEnumerable`1 childrenNodes)
   à TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor(IASTVisitor astVisitor)
   à TypeCobol.Compiler.Nodes.Node.VisitChildrens(IASTVisitor astVisitor, Boolean continueVisit, IEnumerable`1 childrenNodes)
   à TypeCobol.Compiler.Nodes.Node.AcceptASTVisitor(IASTVisitor astVisitor)
[...]

PR #2336, opened 2022-09-14
WI #2185 Relocate only statements when new Sentence is created

fm-117:
Instead of changing the `ProgramClassBuilder` workflow, I'd like to try this solution which ensure a `Paragraph` won't be relocated as a child of a `Sentence`.

### DIFF
--- a/TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
+++ b/TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
@@ -18,6 +18,44 @@ namespace TypeCobol.Compiler.CupParser.NodeBuilder
     /// </summary>
     public class ProgramClassBuilder : IProgramClassBuilder
     {
+        private static bool _SourceCodeDumped;
+
+        private void LogErrorIncludingFullSourceCode(string message)
+        {
+            if (_SourceCodeDumped) return; //Dump only once to avoid huge logs
+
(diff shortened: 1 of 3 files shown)


TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -914,7 +914,7 @@ public virtual void EndSentence(SentenceEnd end, bool bCheck)
         /// </summary>
         public virtual void CheckStartSentenceLastStatement()
         {
-            if (LastEnteredNode != null)
+            if (LastEnteredNode is Statement)
```
smedilol, line 917: ok with this change.
Can you also add a check in `StartParagraph(ParagraphHeader header)`:
```C#
       public static bool NeedDump = true;
       public virtual void StartParagraph(ParagraphHeader header)
        {
           //pseudo code
            if (CurrentNode is Sentence)
            { 
                  Exit(CurrentNode);
                  //Dump only once to avoid too much logs in our system
                  if (NeedDump)
                  {
                      Dump all sources with _codeElementsLines;
                      NeedDump = false;
                  }
            }
        }
```

fm-117: Done

fm-117:
Code updated to trace all invalid parents for paragraphs, not only sentences.

smedilol (approved):

PR #2350, opened 2022-10-12
WI #2217 Use original columns layout when rescanning tokens after a REPLACE

### LINKED ISSUE #2217: During a replace the Scanner always use ColumnsLayout.FreeTextFormat
**Describe the bug**
During a replace the `Scanner` always use `ColumnsLayout.FreeTextFormat`.
See method `ScanIsolatedToken` in the `Scanner`.
And the line `TokensLine.CreateVirtualLineForInsertedToken(0, tokenText);`

The `ScanState` used by `ScanIsolatedToken` is correct and is the `ScanState` of the original token.
But the `ScanState` doesn't have the information of the `ColumnsLayout`.

**To Reproduce**
Cobol code that cause the bug :
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPgm.
       procedure division.
      
       REPLACE ==:-MAJ:== BY ====.
           PERFORM EACGAPEL-C0001:-MAJ:.
           .
       END PROGRAM MyPgm.
```

**Expected behavior**
ScanIsolatedToken must use the correct ColumnsLayout.

**Technical**
Suggestion:
Pass argument of type `ColumnsLayout` to method `ScanIsolatedToken`.
`ColumnsLayout` could be exposed by `ICobolTextLine`, so the value could be retrieve by the caller of `ScanIsolatedToken`.

**How to test automatically**
We need a test in the Scanner that rely on the ColumnsLayout.
Maybe a simple test like `tokensLine.ColumnsLayout == ColumnsLayout.CobolReferenceFormat && tokensLine.Text.Length > 80` already in the Scanner but to check a whole line.

fm-117:
Fixes #2217.


TypeCobol/Compiler/CodeElements/Entry/DataType.cs
```diff
@@ -201,15 +201,6 @@ private static Nodes.TypeDefinition CreateBuiltIn(DataType type)
                     false,
                     0,
                     3);
-                var tokenLine = TokensLine.CreateVirtualLineForInsertedToken(dataTypeDescriptionEntry.Line, "01 CURRENCY TYPEDEF STRICT PUBLIC PIC X(03).");
```
smedilol, line 204: Why remove this code?

fm-117: While tracking usages of `CreateVirtualLineForInsertedToken` to update the param list, I noticed that CURRENCY is the only built-in type with 'fake' tokens. So I removed it for sake of consistency. I think ConsumedTokens for CURRENCY may have been required in the past but are no longer used now (?)

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -364,7 +364,7 @@ public static void ScanTokensLineContinuationGroup(IList<TokensLine> continuatio
                 if (concatenatedLine.Length > 0)
                 {
                     // Scan the continuation text, and get its last token so far
-                    TokensLine temporaryTokensLine = TokensLine.CreateVirtualLineForInsertedToken(firstSourceLine.LineIndex, concatenatedLine);
+                    TokensLine temporaryTokensLine = TokensLine.CreateVirtualLineForInsertedToken(firstSourceLine.LineIndex, concatenatedLine, ColumnsLayout.FreeTextFormat);
```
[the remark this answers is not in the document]
smedilol: Ok I get it, a continuation line can be longer than 80 chars.
But it means, that currently continuation line are scanned as FreeTextFormat. So all our tests on `ColumnsLayout` and column of char/token cannot be done.
I suggest to create a new `ColumnsLayout`: `CobolReferenceFormatForContinuationLine` (or a shorter name ;) ).
This layout will only be used in the Scanner step to scan continuation line.
The format is like `CobolReferenceFormat` but with no end limit for SourceText and no CommentText:
- Chars 1 to 6 : Sequence Number
- Char 7 : Indicator
- Char 8 to end of line : SourceText

There are only 11 real test of `ColumnsLayout` in TypeCobol project. Others projects like CodeGen will not see this ColumnLayout.

What do you think ?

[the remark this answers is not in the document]
smedilol: I've created issue #2357 about that.

TypeCobol/Compiler/CodeElements/Entry/DataType.cs
fm-117: yes done

smedilol (approved):

Issue #2235, opened 2022-05-16
Optimize CodeElementParserStep/TokensLinesTokenStream


smedilol:
## What is the current problem
In case of incremental parsing, `CodeElementParserStep` will only scan a portion of CodeElements.
See `// --- PREPARATION PHASE : identify all parse sections where code elements need to be refreshed ---` in `CodeElementParserStep.cs`.
This is based on a `IList<DocumentChange<IProcessedTokensLine>> processedTokensLinesChanges`.
After that we have the first line and last line to reparse.

Then we need to initialize the tokenStream to the first token of the portion to reparse.
That's where we can optimize.
Currently, we start at the first token of the document and we iterate on each token until we find the token of the first line to reparse.
See `// TO DO : optimize this naive implementation` in `TokensLinesTokenStream`.

## Describe the solution you'd like
Directly position the tokenStream on the first line to reparse or only a few lines before the line to reparse.

## Technical
If the previous token was in a COPY, then we can simply use `ImportedToken.CopyDirective.COPYToken`.
If the previous token was the target of a Replace, then we can simply use `ReplacedToken.OriginalToken`.
Think about a mix of COPY and Replace.

## How to test automatically
Use Performance test of the parser or CFG/DFA
The gain is about 5% for the parser.

Write dedicated Incremental tests to check that the portion to reparse is calculated correctly.
These tests will also help to detect bug in incremental mode.
See `CompilationUnit.CodeElementsDocumentSnapshot.CurrentVersion.changes` to detect if changes made to CodeElementLines are correct.

We need a lot of tests.

A few examples of test to create:

### After COPY, case 1
```cobol
        COPY XXX.
```
Update to:
```cobol
        COPY XXX.
		01 toto pic X.
```

### After COPY, case 2
```cobol
        01 Group1. COPY XXX.
```
Update to:
```cobol
        01 Group1. COPY XXX.  
		01 toto pic X.
```

### After COPY, case 3
```cobol
        COPY XXXCode.
```
Update to:
```cobol
        COPY XXXCode.
		move a to b
```

### After COPY, case 4
```cobol
        display "aaa". COPY XXXCode.
```
Update to:
```cobol
        display "aaa". COPY XXXCode.
		move a to b
```

### After Replace, case 1
```cobol
        replace ==A== by ==01==.
		A Group1.
```
Update to:
```cobol
		replace ==A== by ==01==.
		A Group1.
		   05 Var 1 pic X.
```

[...]

PR #2370, opened 2022-10-31
2235 Optimize code element parser step in incremental

smedilol:
WI #2235 Optimize code element parser step in incremental mode only.
In incremental mode, tokenIterator will now start at a more appropriate line to avoid iteration over all tokens of the document.

What's left to do:
- Specific incremental tests. We already have some incremental tests. I need to be sure that they cover the case of this implementation.
I've tested in our private IDE and it works fine so far.
- Create a performance test with a lot of COPY to reflect our real use case. 
   - One test with a modification in the data division
   - One test with a modification in the procedure division

Performance gain:
[Gain-2235.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/9899396/Gain-2235.xlsx)
As always with performance gain, if you run the same test twice, you won't gain exactly the same results. You can still see that CodeElementParserStep is close to 0ms.

### DIFF
--- a/TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
+++ b/TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
@@ -110,10 +110,11 @@ namespace TypeCobol.Compiler.Parser
             //Start line to seek token in the stream of token (stream of token=tokenIterator) 
             int startLineToSeekToken = 0;
 
+            //For incremental mode, calculate best value for startLineToSeekToken
             if (largestRefreshParseSection != null)
             {
                 //No need to check for ReplacedToken or ImportedToken here, as we directly interact with documentLines
-                //which are the lines of the original document
+                //which are the lines of the original document.
 
                 //TODO find all ReplaceDirective that can target the current line.
                 //Count the largest number of tokens that a Replace directive can target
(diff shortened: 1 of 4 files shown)


TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
```diff
@@ -107,8 +107,36 @@ internal static IList<DocumentChange<ICodeElementsLine>> ParseProcessedTokensLin
...
+
+                //TODO find all ReplaceDirective that can target the current line.
+                //Count the largest number of tokens that a Replace directive can target
+                //Rollback by this number of tokens
+
+                //Temporary: for now rollback X lines as it's unlikely that a Replace target more than that
+                var line = Math.Max(0, largestRefreshParseSection.StartLineIndex - 40);
+
+
+                //Take previous line has long as there are tokens continued on this line
```
fm-117, line 127: as long as

TypeCobol/Compiler/Preprocessor/CopyTokensLinesIterator.cs
```diff
@@ -54,26 +54,35 @@ private struct CopyTokensLinesIteratorPosition
...
         }
 
         /// <summary>
-        /// Resets the iterator position : before the first token of the document
+        /// Resets the iterator position :
+        /// - Before the first token of the document if startLine = 0
+        /// - To the specified startLine otherwise
         /// </summary>
-        public void Reset()
+        public void Reset(int startLine = 0)
```
fm-117, line 72: The method can be private and no need for a default value

TypeCobol/Compiler/Preprocessor/CopyTokensLinesIterator.cs
```diff
@@ -54,26 +54,35 @@ private struct CopyTokensLinesIteratorPosition
         /// <summary>
         /// Set the initial position of the iterator with startToken.
         /// </summary>
-        public CopyTokensLinesIterator(string sourceFileName, IReadOnlyList<IProcessedTokensLine> tokensLines, int channelFilter)
+        public CopyTokensLinesIterator(string sourceFileName, IReadOnlyList<IProcessedTokensLine> tokensLines, int channelFilter, int startLine = 0)
```
fm-117, line 57: Used only once so better remove the default value

fm-117 (changes requested):

smedilol:
Replaced by #2433

Issue #2236, opened 2022-05-16
Optimize access to our ISearchableReadOnlyList / ImmutableList


smedilol:
## What is the current problem
If you look at `CodeElementTokenizer` with a Profiler, you'll see that iteration of `ImmutableList` is slow.
This is the same if you use direct index access to an element ot the `ImmutableList`.
The `ImmutableList` is iterated multiple times by the paser. Most parser step and the diagnostics can iterate it.

The parser should use ISearchableReadOnlyList interface which wraps our ImmutableList class. But there are several cast of ISearchableReadOnlyList to ImmutableList.

## Describe the solution you'd like
As we need to update our list of `CodeElementLine` : insert, update or remove at a specified index, the `ImmutableList` is the better choice.
A LinkedList is not accessible directly with a index. At least not in a fast way.

After a simple test, it's faster to copy the `ImmutableList` to another list and then iterate over this list.
The solution is to copy the `ImmutableList` to a simple implementation of `ISearchableReadOnlyList` backed by a simple `List`. Of course this implementation should prevent any modification to the list.

**Technical**
- Replace all usage of `ImmutableList` by `ISearchableReadOnlyList` if only the interface is needed.
- Restrict the usage of `ImmutableList` to `CompilationDocument` to manage text update step. 
- Create a simple implementation of `ISearchableReadOnlyList` backed by a simple `List` and use it in all step of the parser after `text update` step.

**How to test automatically**
We can use the same tests made for #2235 and checks the previous version of the Document Snapshot to be sure that each version is immutable.

fm-117:
The .NET framework has its own implementation for [ImmutableList](https://docs.microsoft.com/fr-fr/dotnet/api/system.collections.immutable.immutablelist-1?view=net-6.0), unfortunately it is available starting with .NET 5...

smedilol:
Depending on the result of #2442, this issue might not be necessary.

PR #2371, opened 2022-10-31
2236 Optimize acces to document lines in incremental mode

smedilol:
For WI #2236.
Optimize access to our document lines (`ISearchableReadOnlyList`) in incremental mode.

`ISearchableReadOnlyList` is currently implemented with an `ImmutableList`.
`ImmutableList` is great to manage our `CodeElementLines` (see `TypeCobol.Compiler.CompilationDocument.compilationDocumentLines`) and update them in incremental mode. But the problem is `ImmutableList` are very slow to iterate. A direct element access is also very slow.
After tests, it's faster to iterate our `ImmutableList` only once and copy it into a standard List then iterate this List multiple times, than to iterate the `ImmutableList` multiple times.

This PR is divided in 3 steps:
- Replace cast to ImmutableList with cast to ISearchableReadOnlyList (0b05c937e6146842b89e0da358b94fbeabf805d5).
  - This PR don't try to replace all usage of CodeElementsLine by ICodeElementsLine
- Code cleanup in  ISearchableReadOnlyList to help following step (c8607ada6275307656f0b285fa8473fc54c41912).
- Create implementation of ISearchableReadOnlyList with a simple List and use it in `TypeCobol.Compiler.CompilationDocument.RefreshTokensDocumentSnapshot` (168a3ca77b907f0f6655c07b36bd71f992716097).

## What's left to do
- Specific incremental tests. We already have some incremental tests. I need to be sure that they cover the case of this implementation.
I've tested in our private IDE and it works fine so far.
   - Check line in CompilationDocument that used `.ToBuilder()` then `.ToImmutable()`.
   - Check `TypeCobol.Compiler.CompilationDocument.PrepareDocumentLineForUpdate` and direct access to `compilationDocumentLines[index]`

## Bonus
We could add a new parameter to know if the parser will be used in incremental mode or not and then directly use `SearchableReadOnlyList` for batch mode to handle `TypeCobol.Compiler.CompilationDocument.compilationDocumentLines`.
Or always start with `SearchableReadOnlyList` and if incremental mode is triggered, switch to `ImmutableList`.

## Performance gain
Small gain only in incremental mode.
[Perf-Gain.xlsx](https://github.com/TypeCobolTeam/TypeCobol/files/9899632/Perf-Gain.xlsx)
This Excel also include gain combined with PR #2370.

As always with performance gain, if you run the same test twice, you won't gain exactly the same results.


TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -597,7 +597,9 @@ public void RefreshTokensDocumentSnapshot()
                 // Create a new snapshot only if things changed since last snapshot
                 if (TokensDocumentSnapshot == null || TokensDocumentSnapshot.CurrentVersion != currentTokensLinesVersion)
                 {
-                    TokensDocumentSnapshot = new TokensDocument(TextSourceInfo, textLinesVersionForCurrentTokensLines, currentTokensLinesVersion, compilationDocumentLines.ToImmutable());
+                    //Copy our compilationDocumentLines into a new List.
+                    //As compilationDocumentLines is an ImmutableList, it's faster in next step to copy the ImmutableList once in a new list and then iterate over it multiples times.
```
fm-117, line 601: To be precise, an `ImmutableList.Builder`

TypeCobol/Compiler/Concurrency/SearchableReadOnlyList.cs
```diff
@@ -0,0 +1,169 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TypeCobol.Compiler.Concurrency
+{
+
+
+    /// <summary>
+    /// TODO it's not a real ReadOnyList, try to inherits from ReadOnlyCollection or override all *modification* method and a throw an exception.
```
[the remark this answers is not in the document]
fm-117: We can also wrap a List ?

TypeCobol/Compiler/Concurrency/SearchableReadOnlyList.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (this._index < 0 || (uint)this._index >= (uint)list.Count)
+                    return this.MoveNextRare();
+                this._current = list[this._index];
+                this._index += _reverse ? -1 : 1;
+                return true;
+            }
+
+            private bool MoveNextRare()
+            {
+                this._index = this._list.Count + 1;
```
fm-117, line 137: `+ 1` ?

fm-117 (changes requested):

smedilol:
@fm-117 @efr15 This PR must be tested again with .Net 6.
I'm not sure it's still worth it.

smedilol:
@fm-117 @efr15 Maybe just keep 2 commits:
- Replace cast to ImmutableList with cast to ISearchableReadOnlyList (https://github.com/TypeCobolTeam/TypeCobol/commit/0b05c937e6146842b89e0da358b94fbeabf805d5).
    - This PR don't try to replace all usage of CodeElementsLine by ICodeElementsLine
- Code cleanup in ISearchableReadOnlyList to help following step (https://github.com/TypeCobolTeam/TypeCobol/commit/c8607ada6275307656f0b285fa8473fc54c41912).

fm-117:
Replaced by #2575.

PR #2375, opened 2022-11-02
WI #2367 Force uppercase on suffix and pre-suffix in EI Legacy replacing syntax

### LINKED ISSUE #2367: EI Suffixing only apply on uppercase text
**Describe the bug**
The text inside the COPY must be in uppercase to be the target of the suffixing mechanism.
The copy statement can be written in any case. It doesn't matter.
The target of the suffixing mechanism are the last 3 letters of the COPY name without suffix followed by character `-`.

For copy statement `COPY YxxxaaaS` or `COPY YXXXAAAS`, here are how the suffixing is applied.
| **Text inside COPY** | **Target of suffixing** | **Result**     |
|------------------|---------------------|------------|
| AAA-var1         | Yes                 | AAAS-var1  |
| aaa-var1         | No                  | aaa-var1   |
| AAA-             | Yes                 | AAAS-      |
| aaa-             | No                  | aaa-       |
| AAAvar1          | No                  | AAAvar1    |
| aaavar1          | No                  | aaavar1    |
| AAA              | No                  | AAA        |
| aaa              | No                  | aaa        |
| AAA-AAA          | Yes                 | AAAS-AAA   |

**Technical**
See `ReplaceTokensLinesIterator`.

**How to test automatically**
Standard pgm test with EI rules activated.

fm-117:
Fixes #2367.


TypeCobol/Compiler/Directives/CompilerDirective.cs
```diff
@@ -455,7 +453,7 @@ public CopyDirective(CompilerDirectiveType type, Token copyOrExecToken) : base(t
         public bool InsertSuffixChar { get; set; }
```
smedilol, line 453: Optional : could be changed to :
```suggestion
        public bool InsertSuffixChar => Suffix != null;
```

TypeCobol/Compiler/Directives/CompilerDirective.cs
```diff
@@ -867,15 +865,15 @@ public TextNameVariation(string textNameWithSuffix)
             /// <summary>
             /// Text name without suffix
             /// </summary>
-            public string TextName { get { return HasSuffix ? TextNameWithSuffix.Substring(0, 7) : TextNameWithSuffix; } }
+            public string TextName { get { return HasSuffix ? TextNameWithSuffix.Substring(0, 7).ToUpper() : TextNameWithSuffix; } }
```
smedilol, line 868: This change, means that the copy will now be written in uppercase in error message.
I'd rather set only properties `Suffix`and `PreSuffix` in uppercase. What do you think ?

smedilol (changes requested):

TypeCobol/Compiler/Directives/CompilerDirective.cs
fm-117: It works, it relies on the fact that the `Suffix` property of `CopyDirective` is written only in case of a variation having a suffix.

smedilol (approved):

Issue #2357, opened 2022-10-20
Continuation lines are always scanned with ColumnsLayout.FreeTextFormat


smedilol:
**Describe the bug**
Continuation lines are always scanned with layout `ColumnsLayout.FreeTextFormat`.

**To Reproduce**
Any continuation line.

**Expected behavior**
Continuation line scanned with the same format of the document.

**Technical**
I suggest to create a new ColumnsLayout: `CobolReferenceFormatForContinuationLine` (or a shorter name ;) ).
This layout will only be used in the Scanner step to scan continuation line.
The format is like `CobolReferenceFormat` but with no end limit for `SourceText` and no `CommentText`:
- Chars 1 to 6 : Sequence Number
- Char 7 : Indicator
- Char 8 to end of line : SourceText

There are only 11 real test of `ColumnsLayout` in TypeCobol project. Others projects like CodeGen will not see this `ColumnLayout`.

**How to test automatically**
Code a rule in the Scanner where the `ColumnsLayout.CobolReferenceFormat` is used.

PR #2380, opened 2022-11-07
WI #2356 Ensure "literal not correctly delimited" diagnostic is reported for continuation lines

### LINKED ISSUE #2356: Scanner diagnostics on continuation line are not reported
**Describe the bug**
Diagnostics added on a continuation line and not associated with a specific token are not reported.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.     MyPgm.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 Var1 pic X(08).
       PROCEDURE DIVISION.
      *    Error reported:
      *    "Syntax error: Literal is not correctly delimited"
           move
              "a  to Var1

      *    Error not repoted
           move
      -       "a  to Var1

           goback
           .
       end program MyPgm.
```

**Expected behavior**
Error reported correctly on continuation line.

**Technical**
In Scanner, see comment:
`// Then attribute each token and diagnostic to its corresponding tokens line`
We report diagnostics associated with tokens but not diagnostics associated with the `virtualContinuationTokensLine`.

**How to test automatically**
Standard pgm test.

fm-117:
Fixes #2356 

### About this PR:
The problem resides in the `TokensLine.AddDiagnostic` overload accepting direct positioning of diagnostic instead of token positioning. Diagnostics created using direct positioning cannot be copied from the virtual line onto the real line.

Out of the 19 usages of this 'unsafe' method:
- 8 are applied onto a real continuation line so we won't risk losing diags
- 6 are related to multiline or formalized comments which are TypeCobol specific
- 4 are potentially applied on virtual lines and then lost but do not actually concern virtual lines (indicator checks, max line size, both of which are non relevant in FreeFormat, this may evolve considering #2217 and #2357)
- the remaining usage is for "Literal not correctly delimited" and has been updated in this PR

### DIFF
--- a/TypeCobol/Compiler/Scanner/Scanner.cs
+++ b/TypeCobol/Compiler/Scanner/Scanner.cs
@@ -1307,15 +1307,18 @@ namespace TypeCobol.Compiler.Scanner
                             }
 
                             delimiterToken = new Token(TokenType.PseudoTextDelimiter, startIndex, startIndex + 1, usesVirtualSpaceAtEndOfLine, tokensLine);
-                            if (followingChar == ',' || followingChar == ';')
+                            if (currentState.BeforeLastSignificantToken?.TokenType != TokenType.BY)
                             {
-                                // Error
-                                tokensLine.AddDiagnostic(MessageCode.InvalidCharAfterPseudoTextDelimiter, delimiterToken, followingChar);
-                            }
-                            else if (followingChar != ' ' && followingChar != '.')
-                            {
-                                // Warning
-                                tokensLine.AddDiagnostic(MessageCode.DotShouldBeFollowedBySpace, delimiterToken, followingChar, currentIndex + 1);
+                                if (followingChar == ',' || followingChar == ';')
+                                {
+                                    // Error
+                                    tokensLine.AddDiagnostic(MessageCode.InvalidCharAfterPseudoTextDelimiter, delimiterToken, followingChar);
+                                }
+                                else if (followingChar != ' ' && followingChar != '.')
+                                {
+                                    // Warning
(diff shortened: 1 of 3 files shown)


smedilol:
On top of this PR, isn't it possible to copy diagnostics on the virtual line into the original line.
I mean something like this in the method `void InitLine()`:
```C#
             originalLine.Diagnostics.AddRange(virtualContinuationTokensLine.Diagnostics);
```

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -1050,46 +1060,29 @@ public override Token GetTokenStartingFrom(int startIndex)
                         return new Token(TokenType.MultiplyOperator, startIndex, startIndex, tokensLine);
                     }
                 case '%':
-                    if (line.Length >= currentIndex + 2 && line[currentIndex + 1] == '>' && line[currentIndex + 2] == '>')
```
smedilol, line 1053: This code is not useful anymore ?

fm-117: Yes and bugged.
The diagnostics for `WrongFormalizedCommentMarckupPosition` and `WrongMultilineCommentMarckupPosition` are created elsewhere:

https://github.com/TypeCobolTeam/TypeCobol/blob/ee6b01617bc4155372a9429b7c786729e4f7ebbd/TypeCobol/Compiler/Scanner/Scanner.cs#L874-L883

https://github.com/TypeCobolTeam/TypeCobol/blob/ee6b01617bc4155372a9429b7c786729e4f7ebbd/TypeCobol/Compiler/Scanner/Scanner.cs#L916-L925

And it does not concern starting tags but closing ones (and the check on index value is wrong too).

smedilol (approved):

PR #2404, opened 2022-12-19
WI #2389 Check last fetched token to fix StopToken lookup

### LINKED ISSUE #2389: Incremental bug: Change first line of a program which contains more than 1 line before ID DIVISION
**Describe the bug**
When a program contains 2 lines or more before the `identification division`, when you modifiy the first with an incremental change, then you get the error:
```
Runtime error: System.ArgumentException: Un élément avec la même clé a déjà été ajouté.
   à System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   à System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   à TypeCobol.Compiler.CupParser.NodeBuilder.ProgramClassBuilder.Enter(Node node, CodeElement context, SymbolTable table)
   à TypeCobol.Compiler.CupParser.NodeBuilder.ProgramClassBuilder.StartDataDescriptionEntry(DataDescriptionEntry entry)
   à TypeCobol.Compiler.CupParser.CUP_TypeCobolProgramParser_actions.CUP_TypeCobolProgramParser_do_action(Int32 CUP_TypeCobolProgramParser_act_num, lr_parser CUP_TypeCobolProgramParser_parser, StackList`1 xstack1, Int32 CUP_TypeCobolProgramParser_top)
   à TypeCobol.Compiler.CupParser.TypeCobolProgramParser.do_action(Int32 act_num, lr_parser parser, StackList`1 CUP_parser_stack, Int32 top)
   à TUVienna.CS_CUP.Runtime.lr_parser.parse()
   à TypeCobol.Compiler.Parser.ProgramClassParserStep.CupParseProgramOrClass(TextSourceInfo textSourceInfo, ISearchableReadOnlyList`1 codeElementsLines, TypeCobolOptions compilerOptions, SymbolTable customSymbols, PerfStatsForParserInvocation perfStatsForParserInvocation, ISyntaxDrivenAnalyzer[] customAnalyzers, SourceFile& root, List`1& diagnostics, Dictionary`2& nodeCodeElementLinkers, List`1& typedVariablesOutsideTypedef, List`1& typeThatNeedTypeLinking)
```
On top of that, if the program contains a COPY, you can also get the error:
```
Error in copy 'XXXXX' at line Y : Syntax error : A group item cannot be empty.
```

**To Reproduce**
Cobol code that cause the bug : (if any)
```cobol
      * Comment
      * Comment 2
       IDENTIFICATION DIVISION.
       PROGRAM-ID. CCTGAPEL.
       DATA DIVISION.
       working-storage  SECTION.
       COPY YDVZOSM REPLACING ==:DVZOSM:== BY ==DVZOSM==.

       END PROGRAM CCTGAPEL.
```
And `YDVZOSM.copy` :
```cobol
       01 :DVZOSM:.
           05 :DVZOSM:-Var1 pic X.
```

Add one character on the first line to get the errors.

You can also reproduce the bug with this program and the same copy:
```cobol
        
        
       IDENTIFICATION DIVISION.
       PROGRAM-ID. CCTGAPEL.
       DATA DIVISION.
       working-storage  SECTION.
       COPY YDVZOSM REPLACING ==:DVZOSM:== BY ==DVZOSM==.

       END PROGRAM CCTGAPEL.
```
And then also add 1 char on the first line.

**Expected behavior**
No error

**Technical**
Will be edited by a team member.
Implementation details with the chosen solution.

**How to test automatically**
Will be edited by a team member.
Only if our current automated tests doesn't handle this.

fm-117:
Fixes #2389.


smedilol:
I suggest to handle all possibles cases in TokensLinesTokenStream.cs.
See this patch
[Proposed-TokensLinesTokenStream.txt](https://github.com/TypeCobolTeam/TypeCobol/files/10342519/Proposed-TokensLinesTokenStream.txt)

The idea is to delete `indexOfStopTokenReplacedByEOF` as others properties are enough to detect stopToken.
Handle the case where multiple tokens could be in the buffer. It's currently not supposed to be useful as we only fetch token one by one, but if one day we decide to change that then this class will already handle it.
Reset stop token properly.

fm-117:
I think the `indexOfStopTokenReplacedByEOF` is still useful. Since the `Fetch` method auto-replaces the first stop token seen by EOF, we'll always alter the stream one token at a time. So it is more efficient to keep the index instead of iterating again in the `ResetStopTokenLookup` method. Am I missing something ?

I've pushed the change to check all fetched tokens (+ some renaming).

smedilol (approved):

Issue #2412, opened 2023-01-03
Add incremental tests to validate future changes


fm-117:
Part of #2086.

We need more incremental tests that cover basic and advanced uses of the parser inside an IDE.
To add
- variable declaration
- MOVE editing
- IF editing
- EVALUATE editing
- paragraph declaration and PERFORM editing
- structural changes
  - adding a section into a division

See #2235 for more advanced test cases.

PR #2413, opened 2023-01-03
WI #2412 Add basic edits and structural edits tests

fm-117:
New incremental tests as described in #2412.


smedilol (changes requested):
Ok with the tests but to check correctly what happened during the test, I'd rather use on each step:
- Nodes instead of CodeElements comparator to be sure that our AST is correct
- Add INC comparator but limit the history to a depth of 1. Otherwise it's too complicated to read.

smedilol:
After our discussion, I've pushed the changes:
- INC and Nodes comparator for all tests
- Limit incremental changes history to a depth 1 in the context of unit test

smedilol (approved):

PR #2419, opened 2023-01-13
WI #2418 Add new incremental performance tests

### LINKED ISSUE #2418: Add new incremental performance test
The goal is to have new performance tests addressing the incremental mode.

All tests will operate on a large file, see CNAF files to find a good candidate.

4 test cases to create:
- update a single line at the beginning of the file (like a hundred lines after the start of the `DATA DIVISION`)
- update a single line at the end of the file (like few thousands lines before the end, should still be in `PROCEDURE DIVISION`)
- add a line at the beginning of the file (same position as first case)
- add a line at the end of the file (same position as second case)

As for the code, copy existing performance test.

fm-117:
Fixes #2418.

### TODO
- run the tests to get a baseline


TypeCobol.Test/Parser/Performance/Performance.cs
```diff
@@ -213,7 +223,65 @@ public void Part3_Incremental_Cobol85_DeepVariables()
...
+        //[Ignore]
+        public void Part4_Incremental_LargeFile_LineUpdateAtBeginning()
+        {
+            /*
+             * Line 352, we are renaming 88-level 'CEA-TP-RETURN-DUPKEY' to 'CEA-TP-RETURN-DUPLICATE-KEY'
+             * and then readjust the end of the line to match 80 chars.
+             */
+            var rename = new RangeUpdate(351, 34, 351, 34, "LICATE-");
+            var adjust = new RangeUpdate(351, 49, 351, 56, string.Empty);
+            IncrementalPerformance2(LargeFile, rename, adjust);
```
smedilol, line 243: To be equivalent of others tests, I'd rather create only 1 RangeUpdate.
The current code gives me an error.
I'd prefter to create a RangeUpdate that doesn't insert new characters to be sure that we are always in an update situation.
```suggestion
/*
             * Line 352, we are renaming 88-level 'CEA-TP-RETURN-DUPKEY' to 'CEA-TP-RETURN-DUP-EY'
             */
            var rename = new RangeUpdate(351, 34, 351, 35, "-");
            IncrementalPerformance2(LargeFile, rename);
```

smedilol (changes requested):

smedilol (approved):

PR #2423, opened 2023-01-24
WI #2416 Optimize text update by using line update mechanism when possible

### LINKED ISSUE #2416: Optimize TextUpdate by using TextChangeType.LineUpdated when possible
**What is the current problem**
1. When a rangeUpdate update a line, we currently delete the line and insert a new one. This is not ideal, because each lines positionned after the deleted or inserted ones will be updated (update their `LineIndex`).

The goal is to optimize this by using TextChangeType.LineUpdated.

2. In CompilationDocument.cs, replace de Skip (LinQ) call on `compilationDocumentLines` by a call to `GetEnumerator(argument of Skip method, -1, false)` 

**Technical**
In TypeCobol/Compiler/CompilationDocument.cs, search usage of ApplyTextChange. There should be at least 2 usages:
- Remove all the old lines
- Insert the updated lines

**How to test automatically**
- Create a performance test that update only one character one a line near the beginning of the file.
The Update Text Lines step must be faster after this issue is implemented.
- Create a performance test that update only one character one a line near the end of the file.
The Update Text Lines step shouldn't be faster

fm-117:
Fixes #2416.

This is incomplete as one LSR test fails in the current state of the branch.


TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -340,36 +336,57 @@ public void UpdateTextLines(RangeUpdate[] updates)
...
+                            else
+                            {
+                                throw new InvalidOperationException("Unable to process RangeUpdate !");
+                            }
+
+                            void InsertOrUpdate(TextChangeType type)
+                            {
+                                string newLine = (lineUpdates != null && lineUpdates.Count > 0)
+                                    ? lineUpdates[i]
+                                    : string.Empty;
```
smedilol, line 375: Optional:
```suggestion
                                string newLine = lineUpdates?.Count > 0 ? lineUpdates[i] : string.Empty;
```

TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -340,36 +336,57 @@ public void UpdateTextLines(RangeUpdate[] updates)
...
-                        //Mark the index line to be removed. The index will remains the same for each line delete, because text change are applied one after another
-                    }
-
-                    // Insert the updated lines
+                    // Update/Remove or Insert the updated lines
                     if (!(startOfFirstLine == null && lineUpdates == null && endOfLastLine == null))
                     {
                         int lineUpdatesCount = (lineUpdates != null && lineUpdates.Count > 0)
                             ? lineUpdates.Count
                             : 1;
```
smedilol, line 344: Optional:
```suggestion
                        int lineUpdatesCount = lineUpdates?.Count > 0 ? lineUpdates.Count : 1;
```

TypeCobol/Compiler/CompilationDocument.cs
```diff
@@ -513,6 +506,28 @@ private void ApplyTextChange(TextChange textChange, IList<DocumentChange<ICobolT
...
+                    //Remove generated diagnostics for the target line and below.
+                    if (!encounteredCodeElement)
+                        lineToUpdate.ResetDiagnostics(); //Reset diags when on the same zone
+
+                    if (offset != 0)
+                        lineToUpdate.Shift(offset);
+
+                    if (lineToUpdate.CodeElements != null)
+                        encounteredCodeElement = true;
+                }
```
smedilol, line 527: No need to iterate over all lines if there is nothing more to do
```suggestion
                    if (offset == 0 && encounteredCodeElement)
                        break; //Nothing more to do
                }
```

smedilol (changes requested):

smedilol:
Here's a proposal for others changes in `CompilationDocument`. 
See this diff :
[CompilationDocumentProposal.txt](https://github.com/TypeCobolTeam/TypeCobol/files/10663156/CompilationDocumentProposal.txt)

TypeCobol/Compiler/CompilationDocument.cs
fm-117: done

TypeCobol/Compiler/CompilationDocument.cs
fm-117: done

smedilol (approved):

Issue #2439, opened 2023-02-10
Detect/Fix anomalous line index order


fm-117:
Some of our debug traces show that `CobolTextLine.LineIndex` property ends up totally unordered.

There are multiple goals in this issues:
- determine whether `LineIndex` is expected to be accurate or not
  - multiple comments in code suggest that `LineIndex` cannot be used to find the line in the code elements lines collection and that `IndexOf` method should be used intead
- detect when line indexes get mixed up and add a specific debug log
- depending on whether `LineIndex` is reliable or not, fix the ordering problem

PR #2443, opened 2023-02-13
WI #2439 Detect anomalous line index

fm-117:


TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -20,6 +20,21 @@ public class ProgramClassBuilder : IProgramClassBuilder
     {
         private static bool _SourceCodeDumped;
 
+        internal void LogAnomalousLineIndex()
+        {
+            const string message = "Anomalous line index found !";
+            if (_SourceCodeDumped)
+            {
+                // Full source with changes has already been dumped, just add a simple trace to help notice the line index problem
+                LoggingSystem.LogMessage(LogLevel.Error, message);
```
smedilol, line 29: I don't think we need a log once the full source has been dumped.
If there is a problem with LineIndex, it will remain till the source is fully parsed again. We could then end up with a lot of logs.

smedilol (changes requested):

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
fm-117: You're right but what about adding a second flag to add this log just once ?

fm-117: I've kept the trace because the invalid AST problems are detected _during_ the AST building and the line index problem are reported _after_ the AST building.
If we don't add a specific trace for wrong line indices we'll likely only have traces for 'invalid parent for paragraph' and we may not remark the lines.

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
```diff
@@ -19,6 +19,26 @@ namespace TypeCobol.Compiler.CupParser.NodeBuilder
...
+
+        internal void LogAnomalousLineIndex()
+        {
+            const string message = "Anomalous line index found !";
+            if (_SourceCodeDumped)
+            {
+                // Full source with changes has already been dumped, just add a trace (once) to help notice the line index problem
+                if (!_AnomalousLineIndexTraced)
+                {
+                    LoggingSystem.LogMessage(LogLevel.Error, message);
```
smedilol, line 32: We need a full dump when a problem is found with index.
We cannot be sure that the dump made for invalid parent is done in the same instance of ProgramClassBuilder than line index problem.

smedilol (changes requested):

TypeCobol/Compiler/CupParser/NodeBuilder/ProgramClassBuilder.cs
fm-117: Now separating dumps on each cause

smedilol (approved):

PR #2458, opened 2023-02-21
WI #2444 Defer diagnostic reset during incremental parsing

### LINKED ISSUE #2444: Fix bugs found with generator AddEmptyLineAtBeginningThenRemove
Child of #2364.

The incremental changes generator has revealed several bugs (see [result file](https://github.com/TypeCobolTeam/TypeCobol/files/10661996/AddEmptyLineAtBeginningThenRemove.txt)).
Among the 4 exceptions;
- one is already fixed as it comes from invalid test case with INC comparator
- one is a duplicate

The goal of this issue is to fix the remaining 2.

TODO:
- add test case for each bug

fm-117:
Fixes #2444.


smedilol (changes requested):
Can you create a dedicated incremental test in order to understand what was not working before this PR ?

fm-117:
> Can you create a dedicated incremental test in order to understand what was not working before this PR ?

@smedilol It's done. 2 tests were failing the "add empty line at beginning then remove" change, I've reproduced them in a shorter form in the new test files.

smedilol (approved):

fm-117:
Some explanations for the new unit tests:

### CEAlteringScanState
#### Before the fix
The diagnostic `Error in copy 'DataDivision' at line 1 : Syntax error : Syntax not supported: 'DATA DIVISION' has been found inside COPY.` would disappear after adding an empty line at the beginning. The other 2 diagnostics would remain and be correctly shifted down.
#### After the fix
All 3 diagnostics are there and move up or down correctly, according to the lines added or removed

### MissingSpaceAfterProgId
#### Before the fix
The diagnostic `A blank was missing before character "P" in column 19. A blank was assumed.` would disappear after adding an empty line at the beginning.
#### After the fix
The diagnostic is there after each step (after initial parsing, after adding line, after removing line)

PR #2459, opened 2023-02-23
WI #2455 Fix backward navigation when trying to find previous CodeElement

### LINKED ISSUE #2455: Fix bugs found with generator AddEmptyLineAtEndThenRemove
Child of #2364.

Fix anomalies found when using `AddEmptyLineAtEndThenRemove` on all files in `TypeCobol.Test\Parser\Programs`

fm-117:
Fixes #2455. But...

...there is a problem with fixes used in #2458 and here. On a line having a COPY directive the booleans `previousCodeElementStartsAtBeginningOfTheLine` / `nextCodeElementStartsAtTheBeginningOfTheLine` are `false` meaning we'll include those lines in reparse section. This induces a performance hit when the COPY is big.

Problem is we currently reset diagnostics and code elements at line level but considering a COPY line, we will reset a lot of code elements (and diagnostics) in one go and we are then forced to reparse the whole copy to get those code element and diagnostics back.

Ideas:
- iterate at code element level instead of code elements line level
  - this should allow to reset only codeelements that really need to be reparsed
  - it does not work for `ParserDiagnostics` though, as this collection is still defined at line level
- Remove `ParserDiagnostics` collection and create fake CodeElements to hold the diagnostics !


TypeCobol.Test/Parser/Incremental/BasicEdits/AddEmptyLineAtEndThenRemove/Typedef2.tcbl
```diff
@@ -0,0 +1,18 @@
+﻿IDENTIFICATION DIVISION.
```
smedilol, line 1: Can you rather use this Cobol code and change the file name to MyPgm.rdz.cbl:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPgm.












       PROCEDURE DIVISION.
       01 OutOfPerimeter2.
          02 OutOfPerimeter2-var1 PIC X.
          02 OutOfPerimeter2-var2 PIC X.

```
inc should be the same than before.

I prefer to avoid usage of TypeCobol reference if the bug is related to Cobol.

smedilol (changes requested):

smedilol (approved):

fm-117:
Some explanations for the new unit tests:

### MyPgm
#### Before the fix
The diagnostic at line 6 `Syntax error : extraneous input '02' expecting {separator, numeric literal, character string, symbol, statement starting keyword, keyword, Formalized Comments elements, Sql statement starting keyword} RuleStack=,  OffendingSymbol=[11,12:02]<IntegerLiteral>{2}` would disappear after adding a blank line at the end.
#### After the fix
All 3 diagnostics stay during the edits

Issue #2463, opened 2023-03-02
Fix  the COPY / REPLACING / REPLACE mechanism


fm-117:
**Describe the bug**
Current algorithm used to apply `COPY` directives, `REPLACE` directives and `REPLACING` clauses does not match IBM compiler behavior.

**Expected behavior**
- first apply all `COPY` directives recursively by importing Tokens
  - the `COPY` directive should not import ProcessedTokens but Tokens (from scanner phase)
  - if the `COPY` directive has a `REPLACING` clause, the `REPLACING` clause applies to all imported tokens
    - it means that `REPLACE` directives written in the copy may be modified by the `REPLACING` clause
    - the `REPLACING` clause is not canceled by a `REPLACE` directive, it works at another level
  - the `REPLACING` clause applies to imported tokens from nested copies
  - nested `REPLACING` clauses are forbidden (IBM error is `A second "COPY" statement with "REPLACING" phrase was found within a nested "COPY".  The second "COPY" statement was discarded.`)
- finally, after the whole document is assembled, apply successive `REPLACE` directives
  - as usual, a `REPLACE` directive is in effect till the next `REPLACE` directive or till the end of the document if none found

**Technical**
The `CopyTokensLinesIterator` must be rewritten, it should include the code to handle the REPLACING mechanism. And this REPLACING mechanism should be taken out from existing `ReplaceTokensLinesIterator`.

**How to test automatically**
### TODO
Create tests
Will be edited by a team member.
Only if our current automated tests doesn't handle this.

fm-117:
Adapt this test, previoulsy used in `TypeCobol.Test\Parser\Preprocessor`

PgmReplaceNested.cbl
```cobol
000001 REPLACE == PIC X == BY == PICTURE N ==.
000002 COPY CpyNested REPLACING == :TAG: == BY == TITI ==
000003                          == X(10) == BY == X(20) ==.
```

CpyNested.cpy
```cobol
000001 REPLACE == TOTO == BY == :TAG: ==.
000002 01 TOTO 
000003 REPLACE == UNKNOWN == BY == ==
000004         == KEYWORD == BY == PIC X ==.
000005 UNKNOWN KEYWORD (10).
```

PR #2474, opened 2023-03-14
WI #2463 Introduce base class for REPLACE and REPLACING iterators

fm-117:
Still incomplete, we need the error message for REPLACE modified by REPLACING


TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
```diff
@@ -187,13 +187,41 @@ internal static IList<DocumentChange<ICodeElementsLine>> ParseProcessedTokensLin
...
+                    if (largestRefreshParseSection.StartLineIndex > 0)
+                    {
+                        var lineBefore = documentLines[largestRefreshParseSection.StartLineIndex - 1];
+                        if (lineBefore.ImportedDocuments != null)
+                        {
+                            // The preceding line imports copies, use last line of the last resolved copy
+                            var lastLine = lineBefore.ImportedDocuments.Values
+                                .Where(d => d != null)
+                                .SelectMany(d => d.SourceDocument.Lines)
+                                .LastOrDefault();
```
smedilol, line 201: For performance reason we should avoid SelectMany.
```suggestion
                            var lastLine = lineBefore.ImportedDocuments.Values.Last(d=>d!=null).SourceDocument.Lines.LastOrDefault();
```

smedilol (changes requested):

TypeCobol/Compiler/Parser/CodeElementsParserStep.cs
fm-117: Reverted.
As long as the CodeElementsParserStep reparses all lines affected by a modified REPLACE, using the value of `ActiveReplaceDirective` coming from a previous parsing round should be fine. So I suggest to keep the previous code.

TypeCobol/Compiler/Parser/CodeElementsLine.cs
```diff
@@ -74,7 +74,6 @@ internal void ResetCodeElements()
         {
             CodeElements = null;
             _ParserDiagnostics = null;
```
smedilol, line 76: Can you just add a comment about `ActiveReplaceDirective`:
```C#
//Don't reset ActiveReplaceDirective. 
//This property is always updated during CodeElementParserStep and we need the original value to start an incremental reparsing of CodeElementParserStep.
```

smedilol (approved):

PR #2497, opened 2023-04-13
WI #2488 Migrate to .NET 6.0

### LINKED ISSUE #2488: Migrate to .NET 6.0
Migrate whole TypeCobol to .NET 6.0 Runtime and framework.

Use `net6.0` for all projects in order to use C# 10 language.

fm-117:
Fixes #2488.

### What has been done ?
- Recreated all projects from scratch using new simplified csproj model
- Fixed path-related operations in unit tests because of new output dirs
  - Instead of a shared bin/obj, we now have separate bin/obj per project
- Removed all `AssemblyInfo.cs`, now using defaults from .NET or customizations directly written in csproj
- Updated test results because of loss of precision in `double` type (?)
- Update test results because of new display name for assemblys (now without extension)
- Removed all `*.config` files
  - `App.config` were not used
  - `packages.config` are now integrated into csproj thanks to `<PackageReference>`
- Removed some unused test files, added one cpy file that was previously linked
- Rewritten Performance tests with CFG or DFA using composition instead of inheritance which seems cleaner for MSTest...
- Fixed some namespace conflicts due to the new implicit usings feature
- Updated use of `GetEncoding` which now require codepage provider registration
- Rewritten CSCup code generation prebuild as a Powershell script
  - Removed generated files from source control
- Migrated from AppVeyor CI To GitHub Actions
  - Removed old build script
  - Added new workflow files
- Added helper project to retrieve LanguageServerRobot and LanguageServerRobot Monitor as NuGet packages
- Added script to prepare debug of LanguageServer


TypeCobol/TypeCobol.ruleset
smedilol, line 1: What is the equivalent of this file with .Net 6 ?

fm-117: After looking up MSDN, I think it's that:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig

I removed completely the analyzer config at first because the CodeAnalysis feature has changed with new project modeland the option to link a ruleset is not available anymore in VS. It looks like the analyzers can still be configured through some config files but the rules applied have been updated too. Translating our old ruleset to the new format may require some work to figure out which rules and severity need a config value.

fm-117: Reproduced the previous configuration with new format, and activated the whole ruleset.
In EI_Debug, the Rebuild action takes around 1min and in EI_Release, the Rebuild action takes a bit more than 2min.

Codegen/test/resources/input/TypeCobol/CopyReplace4Colon/YxxxENT.cpy
smedilol, line 1: Is it a missing file in our current tests (develop branch) ?

fm-117: It was previously linked from another project. I copied it but we can keep the link if you prefer. I preserved links in CLI but I duplicated the file for Codegen as it was the only one being linked.

efr15 (approved):

Codegen/test/resources/input/TypeCobol/CopyReplace4Colon/YxxxENT.cpy
smedilol: Ha ok. Yes it makes more sense to copy the file if we now have a `bin` folder per projet.

PR #2507, opened 2023-04-27
WI #2471 check allowed characters in partial cobol word

### LINKED ISSUE #2471: Check allowed characters in partial cobol word
# Describe the bug
IBM Compiler check characters allowed in replace clause.
See "Chapter 1. Characters" page 3 of IBM PDF "Enterprise COBOL for z/OS 6.3 Language Reference".

Allowed characters are :
```
(space)
+
-
*
/
=
$
,
;
.
'
"
(
)
<
>
:
_
A-Z
a-z
0-9
```
So for example we must forbid character `@` in replace like IBM compiler.

# Out of scope
This issue is only about Partial Cobol word because a few programs in our company are not parsed correctly.
Allowed characters in replace directive out of partial cobol word will not be checked in this issue.

# To Reproduce
CheckPseudoText1.rdz.cbl:
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.    CheckPseudoText1.
       DATA DIVISION.
       working-storage section.

       01  Group1.
       replace ==:_:== by ==FOO==.
           05 :_:-Var pic X.
       replace ==:>:== by ==FOO==.
           05 :>:-Var pic X.
       replace ==:=:== by ==FOO==.
           05 :=:-Var pic X.
       replace ==:<:== by ==FOO==.
           05 :<:-Var pic X.
       replace ==:;:== by ==FOO==.
           05 :;:-Var pic X.
       replace ==:/:== by ==FOO==.
           05 :/:-Var pic X.
       replace ==:.:== by ==FOO==.
           05 :.:-Var pic X.
       replace ==:-:== by ==FOO==.
           05 :-:-Var pic X.
       replace ==:,:== by ==FOO==.
           05 :,:-Var pic X.
       replace ==:+:== by ==FOO==.
           05 :+:-Var pic X.
       replace ==:*:== by ==FOO==.
           05 :*:-Var pic X.
       replace ==:):== by ==FOO==.
           05 :):-Var pic X.
       replace ==:(:== by ==FOO==.
           05 :(:-Var pic X.
       replace ==:$:== by ==FOO==.
           05 :$:-Var pic X.
       replace ==: :== by ==FOO==.
           05 : :-Var pic X.

      *Ok
       replace ==:'':== by ==FOO==.
           05 :'':-Var pic X.
      *Ok
       replace ==:"":== by ==FOO==.
           05 :"":-Var pic X.


      *NOT OK because string is not terminated correctly ?
      *An invalid"REPLACE"statement was found.  Scanning was resumed at
      *  the period terminating the"REPLACE"statement.
       replace ==:':== by ==FOO==.
           05 :':-Var pic X.
      *NOT OK because string is not terminated correctly ?
      *An invalid"REPLACE"statement was found.  Scanning was resumed at
      *  the period terminating the"REPLACE"statement.
       replace ==:":== by ==FOO==.
           05 :":-Var pic X.


      *Non-COBOL character"@"was found in column 19.
      *  The character was accepted.
       replace ==:@:== by ==FOO==.
      *Non-COBOL character(s) were found starting with"@"in column 16.
      *  The characters were discarded.
           05 :@:-Var pic X.


      *Ko Non-COBOL character"€"was found in column 19.
      *The character was accepted.
       replace ==:€:== by ==FOO==.
           05 :€:-Var pic X.

      *Ko Non-COBOL character"£"was found in column 19.
      *The character was accepted.
       replace ==:£:== by ==FOO==.
           05 :£:-Var pic X.

      *Ko Non-COBOL character"#"was found in column 19.
      *The character was accepted.
       replace ==:#:== by ==FOO==.
           05 :#:-Var pic X.

      *Ko Non-COBOL character"é"was found in column 19.
      *The character was accepted.
       replace ==:é:== by ==FOO==.
           05 :é:-Var pic X.

      *Ko Non-COBOL character"&"was found in column 19.
      *The character was accepted.
       replace ==:&:== by ==FOO==.
           05 :&:-Var pic X.

      *Ko Non-COBOL character"["was found in column 19.
      *The character was accepted.
       replace ==:[:== by ==FOO==.
           05 :[:-Var pic X.
      *Ko Non-COBOL character"]"was found in column 19.
      *The character was accepted.
       replace ==:]:== by ==FOO==.
           05 :]:-Var pic X.
      *Ko Non-COBOL character"^"was found in column 19.
      *The character was accepted.
       replace ==:^:== by ==FOO==.
           05 :^ :-Var pic X.
      *Ko Non-COBOL character"{"was found in column 19.
      *The character was accepted.
       replace ==:{:== by ==FOO==.
           05 :{:-Var pic X.
      *Ko Non-COBOL character"}"was found in column 19.
      *The character was accepted.
       replace ==:}:== by ==FOO==.
           05 :}:-Var pic X.

       END PROGRAM CheckPseudoText1.
```

# Expected behavior
Mandatory: check that characters inside a partial cobol word are valid.

Bonus : Same error as IBM compiler.
IBM compiler sometimes seems to stop its parsing or try to find the end of the replace clause. We don't need to reproduce this behavior of course. It would be better to detect that the replace clause was intented to finish at a certain point and report that the replace is invalid.

# Technical
1. In `CobolChar`, add a new method `IsAllowedInsidePartialCobolWord(char)` which must return true if char is accepted by `IsCobolWordChar`, otherwise check if the character is in following list:
```
(space)
+
*
/
=
$
,
;
.
'
"
(
)
<
>
```
Note that `:` will not be accepted for now because we first need to fix #2309 to handle such cases.

2. In `ScannerUtils.CheckForPartialCobolWordPattern`, replace call to `CobolChar.IsCobolWordChar` by `CobolChar.IsAllowedInsidePartialCobolWord`.
3. Check impact on RegexReplace in ReplaceIterator.

# How to test automatically
Standard unit test with Token comparator.

efr15:
Fixes #2471

(space)
:
(
)

are too problematic to allow them


TypeCobol/Compiler/Diagnostics/MessageCode.cs
```diff
@@ -62,6 +62,8 @@ public enum MessageCode
         InvalidExponentInDecimalFloatingPointLiteral = 51,
         InvalidMantissaInDecimalFloatingPointLiteral = 52,
         InvalidCharBeforePseudoTextDelimiter = 53,
-        ShouldBePrecededBySpace = 54
+        ShouldBePrecededBySpace = 54,
+        // End for SQL
```
fm-117, line 66: Well, actually `InvalidCharBeforePseudoTextDelimiter` and `ShouldBePrecededBySpace` are for Cobol but I did not add any comment to signal the end of SQL-related members. Could you move your comment up 2 lines ? Thanks.

TypeCobol/Compiler/Text/CobolChar.cs
```diff
@@ -88,6 +105,30 @@ static CobolChar()
             isCobolWordSeparator[(byte)'='] = true;
             isCobolWordSeparator[(byte)'"'] = true;
             isCobolWordSeparator[(byte)'\''] = true;
+
+            // --- isAdditionalCharAllowedInPartialCobolWord ---
+
+            for (int i = 0; i < 256; i++)
+            {
+                isAdditionalCharAllowedInPartialCobolWord[i] = false;
+            }
```
fm-117, line 114: I think we should remove those unneeded inits in this file (and add a comment stating that bool default value is `false`)

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/MessageCode.cs
efr15: Done

TypeCobol/Compiler/Text/CobolChar.cs
```diff
@@ -35,6 +36,22 @@ public static bool IsCobolWordSeparator(char chr)
...
+            return IsCobolWordChar(chr) || isAdditionalCharAllowedInPartialCobolWord[(byte)chr];
+        }
+
+        /// <summary>
+        /// True if the current char is allowed in a PseudoText
+        /// </summary>
+        public static bool IsAllowedInsidePseudoText(char chr)
+        {
+            return IsAllowedInsidePartialCobolWord(chr) || (chr == ' ') || (chr == ':') || (chr == '(') || (chr == ')');
+        }
```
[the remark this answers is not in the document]
efr15: Done

TypeCobol/Compiler/Text/CobolChar.cs
efr15: Done

TypeCobol.Test/Parser/Programs/Cobol85/ReplaceCheckAllowedCharacters.PGM.txt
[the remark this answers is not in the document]
efr15: Done

TypeCobol/Compiler/Text/CobolChar.cs
```diff
@@ -35,14 +36,33 @@ public static bool IsCobolWordSeparator(char chr)
...
+        /// </summary>
+        public static bool IsAllowedInsidePartialCobolWord(char chr)
+        {
+            if ((chr == ' ') || (chr == ':') || (chr == '(') || (chr == ')'))
+            {
+                // Currently too problematic to allow them
+                return false;
+            }
+
+            return IsAllowedInsidePseudoText(chr);
```
fm-117, line 50:
```csharp
switch (chr)
{
    case ' ':
    case ':':
    case '(':
    case ')':
        // Currently not supported by the parser
        return false;
    default:
        return IsAllowedInsidePseudoText(chr);
}
```

fm-117 (changes requested):

fm-117 (approved):

smedilol (approved):

PR #2522, opened 2023-05-05
WI #2503 Support pointer-32

### LINKED ISSUE #2503: Cobol 6.3 : Support pointer-32
**What is the current problem**
Support `pointer-32`.
As we will not use 64 bits mode in our company we can consider that `pointer-32` works exactly the same as `pointer`.
TODO create an issue to accept LP(64) / AMODE 64 syntax but create an error that says that our parser don't support it. 

Cobol code to illustrate.
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. DVZF0OSM.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 Var1 pointer.
       01 Var2 usage is pointer.

       01 Var3 pointer-32.
       01 Var4 usage is pointer-32.

      *Ok on a group
       01 Group1 pointer-32.
      *    Ok because group has a valid usage that apply to its children
           05 Var5.
      *    Ok same usage as parent
           05 Var6 pointer-32.

      *    Ko usage different than parent
      *    TODO currently not check by our parser
      *The specified"USAGE"was different from the "USAGE"specified at
      *the group level.  The group "USAGE" was assumed for this item.
           05 Var7 pointer.

       01 Var10 pic X.

       LINKAGE SECTION.
       01 Link1 PIC 9(8).

       procedure division.
           set Var1 to address of Var10
           set Var2 to address of Var10
           set Var3 to address of Var10
           set Var4 to address of Var10

           set address of Link1 to Var1
           set address of Link1 to Var2
           set address of Link1 to Var3
           set address of Link1 to Var4

           goback
           .
       END PROGRAM DVZF0OSM .
```

**Technical**
Check every usage of `pointer` and also accept `pointer-32`.

**How to test automatically**
Standard program test with Mix comparator and Memory comparator to check physical lenght.

efr15:
Fixes #2503

1) POINTER-32 is now supported in USAGE statement.
2) STRING and UNSTRING statements accept only POINTER keyword and are consequently not impacted.
3) We choose not to adapt the specific TypeCobol syntax DECLARE function/procedure. So it does not support POINTER-32 either.

Most of the line codes handling DataUsage.POINTER are modified to support POINTER-32.
However 3 of them are not modified because they are related to TypeCobol syntax DECLARE function/procedure.
Here is the list:
- ParameterEntry#Lines (line 134)
- TypeCobolChecker#Check (line 297)
- TypeCobolCodeElementBuilder#CreateFunctionDataParameter (line 139)

Codegen does not handle correctly this new token. However there is no error and nothing is broken.
So we will need to report that TypeCobol does not support POINTER-32.


TypeCobol.Test/Parser/Programs/Cobol85/UsageIsPointer.rdz.cbl
```diff
@@ -0,0 +1,42 @@
...
+       01 Group1 pointer-32.
+      *    Ok because group has a valid usage that apply to its children
+           05 Var5.
+      *    Ok same usage as parent
+           05 Var6 pointer-32.
+
+      *    Ko usage different than parent
+      *    TODO currently not check by our parser
+      *The specified"USAGE"was different from the "USAGE"specified at
+      *the group level.  The group "USAGE" was assumed for this item.
```
fm-117, line 21: Is there an issue for that ? If not, please create it.

fm-117 (changes requested):
- did you test how codegen reacts to this new token ? This is not required to add a test for that but I'd like to know if it breaks something
- no change in `TypeCobolCodeElements.g4` for the functionDataParameter definition ?
- missing check in CrossChecker line 297 ?

TypeCobol.Test/Parser/Programs/Cobol85/UsageIsPointer.rdz.cbl
smedilol: Issue #2524 created

smedilol (approved):

efr15:
@smedilol @fm-117
Here is a codegen test with `var2` declared as `POINTER-32` (compared to `var1 `declared as `POINTER`).

image

Result is:

image

So considering POINTER-32:

1. Declaration is not altered by TypeCobol
2. Syntax SET UP BY is not supported.

Issue #2527, opened 2023-05-17
Activate #nullable enable


smedilol:
Activate `#nullable enable` file by file

Start with:
1. Token (all class hierarchy)
2. La hiérarchie TokenLine
3. DataType
4. SymbolInformation (all class hierarchy)
5. StorageArea
6. Variable
7. CodeElement (root class only)
8. DataDefinitionEntry (all class hierarchy)
9. PictureValidator (all partial class)
10. Node (root class only)
11. DataDefinition (all class hierarchy)

PR #2533, opened 2023-05-23
WI #2527 Add nullability annotations for Scanner-related objects

fm-117:


TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -236,17 +236,11 @@ private static bool IsDebuggingModeActive(MultilineScanState lineScanState, stri
                  * So an inactive debug line has to be parsed as regular source if it participates
                  * in a REPLACE directive.
                  */
-                return lineScanState.InsideReplaceDirective || StartsWithReplace();
+                return lineScanState.InsideReplaceDirective || lineSourceText.StartsWith(nameof(TokenType.REPLACE), StringComparison.OrdinalIgnoreCase);
```
smedilol, line 239: It works because REPLACE token doesn't contains an `_` but I'd rather keep the original code in order to cover further cases with token that contains a `_`.

TypeCobol/Compiler/Scanner/MultilineScanState.SpecialNamesContext.cs
```diff
@@ -156,7 +157,7 @@ private void CreateCurrencyDescriptor()
                 }
 
                 //Helper local functions
-                string Text(Token alphanumericLiteralToken) => ((AlphanumericLiteralTokenValue) alphanumericLiteralToken.LiteralValue).Text;
+                string Text(Token alphanumericLiteralToken) => ((AlphanumericLiteralTokenValue) alphanumericLiteralToken.LiteralValue!).Text; //Using null-forgiving operator here as AlphanumericLiteralToken always has a non-null LiteralValue.
```
smedilol, line 160: How do you know that `LiteralValue` is never null here ?

smedilol (changes requested):

TypeCobol/Compiler/Scanner/MultilineScanState.SpecialNamesContext.cs
fm-117: The token received here is guaranteed to be non-null of type `AlphanumericLiteral`, `HexadecimalAlphanumericLiteral`, `NullTerminatedAlphanumericLiteral`. All three are creatd by the `AbstractScanner.ScanAlphanumericLiteral` method which always sets the `LiteralValue` property.

TypeCobol/Compiler/Scanner/Scanner.cs
fm-117: I've restored the original code but this forces us to check for non-null of the `string replaceKeyword` which is a bit odd since we know at compile-time that this is always true. That's why i went for the `nameof` approach instead.

smedilol (changes requested):

TypeCobol/Compiler/Scanner/MultilineScanState.SpecialNamesContext.cs
fm-117: done

TypeCobol/Compiler/Scanner/Scanner.cs
fm-117: null-forgive on result

smedilol (approved):

efr15 (approved):

PR #2559, opened 2023-08-07
WI #2170 Remove multiple LINQ enumerations to optimize completion

### LINKED ISSUE #2170: Improve completion performance
This is a global issue to list possible performance improvements for completion requests:
- remove all usages of `System.Linq.Expressions` as we don't use any dynamic code features
- check for unnecessary `IEnumerable` enumerations and redundant list copies
  - Methods `CompletionFactory.CompletionFactory.GetCompletion*` could return a new `List `instead of `IEnumerable`. 
  These methods must guarantee that a new List is created every time they are called.
  - Then method `TypeCobolServer.OnCompletion(TextDocumentPosition parameters)` which use AddRange to retrieve could directly reuse the List.
- Check all usages of Node.*QualifiedName as it can be very expensive.
  - See `TypeCobol.LanguageServer.CompletionFactoryHelpers.CreateCompletionItemForSingleVariable`

To help measure improvement, some performance counter for completion have to be added, either into the server or the client.

# How to test ?
Ask completion after a sentence like:
```
move xxxx to 
```
Cancel the completion with Escape.
Ask the completion again without any modification in the source to avoid to trigger a Refresh Node step.
Completion popup should appear immediatly.

fm-117:
Fixes #2170.

No need to do anything on the `*QualifiedName*` properties as they are backed up by a field to avoid multiple computations.

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
@@ -138,7 +138,7 @@ namespace TypeCobol.Compiler.Diagnostics
                         //If there is more than one variable with the same name, it's ambiguous
                         message = string.Format("Call to '{0}'(no arguments) is ambigous. '{0}' is defined {1} times",
                             functionName,
-                            potentialVariables.Count() + functionDeclarations.Count);
+                            potentialVariablesCount + functionDeclarations.Count);
                         DiagnosticUtils.AddError(node, message);
                         return;
                     }
--- a/TypeCobol.LanguageServer/Completion Factory/CodeElementMatcher.cs
+++ b/TypeCobol.LanguageServer/Completion Factory/CodeElementMatcher.cs
@@ -1,9 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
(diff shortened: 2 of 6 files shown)


TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
```diff
@@ -138,7 +138,7 @@ public static void OnNode(Node node)
                         //If there is more than one variable with the same name, it's ambiguous
                         message = string.Format("Call to '{0}'(no arguments) is ambigous. '{0}' is defined {1} times",
                             functionName,
-                            potentialVariables.Count() + functionDeclarations.Count);
+                            potentialVariablesCount + functionDeclarations.Count);
```
efr15, line 141: `functionDeclarations.Count` is evaluated many times, it could be stored in a local variable like `potentialVariablesCount`

efr15 (changes requested):

TypeCobol/Compiler/Diagnostics/TypeCobolChecker.cs
fm-117: `functionDeclarations` is a `List`, calling the `Count` property is inexpensive. Contrary to LINQ `Count()` method which triggers an enumeration.

efr15 (approved):

smedilol (approved):

Issue #2218, opened 2022-04-22
During a replace the Scanner use a text with original column lost


smedilol:
**Describe the bug**
**LINK WITH #2217**
Method `ScanIsolatedToken` in the `Scanner` receive a string which contains the text without the original spaces before the text. It means the new token generated is align to the left. The original column is lost. It means we cannot check the Area of declaration.

**To Reproduce**
Cobol code that cause the bug : (if any)
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. MyPgm.
       procedure division.
      
       REPLACE ==:-MAJ:== BY ====.
           PERFORM EACGAPEL-C0001:-MAJ:.
           .
       END PROGRAM MyPgm.
```
`ScanIsolatedToken` will receive string 
`EACGAPEL-C0001`
not
`                   EACGAPEL-C0001`

**Expected behavior**
The original column of the text to scan is not lost.

**Technical**
Add spaces before the text to rescan so the column is preserved.

**How to test automatically**
We need a test in the Scanner that rely on the column.
We could do that to detect `paragraph` and section `token`.

fm-117:
The sample provided does not show any problem on develop. We'll have to find a new test case where the bug is shown or close this issue (?)

PR #2570, opened 2023-08-30
WI #2218 Preserve column and line while rescanning during a REPLACE

smedilol:
WI #2218 Scan during Replace preserve original column and line

Temporary Virtual Line used to rescan replaced token:
- Always use the lineIndex of the original line
- The line preserve space characters before first token
- Keep the ColumnLayout of the original line

Rewrite AutoReplaceIterator with AbstractReplaceTokensLinesIterator.

### DIFF
--- a/TypeCobol/Compiler/Preprocessor/AutoReplacePartialWordsTokensDocument.cs
+++ b/TypeCobol/Compiler/Preprocessor/AutoReplacePartialWordsTokensDocument.cs
@@ -22,15 +22,15 @@ namespace TypeCobol.Compiler.Preprocessor
             {
                 var nextToken = SourceIteratorNextToken();
 
-
                 // Reset previous replace operations
                 List<ReplaceOperation> updatedReplaceOperations = null;
                 if (nextToken.TokenType == TokenType.PartialCobolWord)
                 {
                     //TODO Don't reset replace but first try to check if it's the same than before
                     updatedReplaceOperations = new List<ReplaceOperation>();
-                    var replacementText = new string(' ', nextToken.StartIndex)  //Keep original spaces before first token
-                                          + nextToken.NormalizedText.Replace(":", string.Empty);
+                    var replacementText =
+                        new string(' ', nextToken.StartIndex) //Keep original spaces before first token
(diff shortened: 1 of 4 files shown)


efr15 (dismissed):

TypeCobol/Compiler/Preprocessor/AutoReplacePartialWordsTokensDocument.cs
```diff
@@ -12,77 +12,36 @@ namespace TypeCobol.Compiler.Preprocessor
...
-                    var scanState = originalToken.TokensLine.InitialScanState;
-                    var generatedReplacementToken = ReplaceTokensLinesIterator.GenerateReplacementToken(originalToken, replacedTokenText, scanState, _compilerOptions);
-
-                    nextToken = new ReplacedPartialCobolWord(generatedReplacementToken, null, originalToken);
+                    //TODO Don't reset replace but first try to check if it's the same than before
+                    updatedReplaceOperations = new List<ReplaceOperation>();
+                    var replacementText = new string(' ', nextToken.StartIndex)  //Keep original spaces before first token
+                                          + nextToken.NormalizedText.Replace(":", string.Empty);
+                    TokensLine tempTokensLine = TokensLine.CreateVirtualLineForInsertedToken(nextToken.TokensLine.LineIndex, replacementText, nextToken.TokensLine.ColumnsLayout);
+                    var replacementToken = new Token(TokenType.UserDefinedWord, nextToken.StartIndex, tempTokensLine.Length - 1, tempTokensLine);
```
fm-117, line 35: Why not keeping using `GenerateReplacementToken` ?

fm-117 (changes requested):
no test ?

TypeCobol/Compiler/Preprocessor/AutoReplacePartialWordsTokensDocument.cs
fm-117: Yeah ok I see it, no need to rescan as we already know the token type

fm-117 (approved):

PR #2571, opened 2023-08-30
WI #2357 Preserve original column of virtual concatenated line

smedilol:
WI #2357 Virtual line used by continuation line preserve original column of the first line

Fix column of diagnostics on continued lines.
Introduce new ColumnLayout CobolReferenceFormatWithoutComment which is like CobolReferenceFormat but with unlimited line length. 
This column layout should only be used by continuation line.

### DIFF
--- a/TypeCobol/Compiler/Scanner/Scanner.cs
+++ b/TypeCobol/Compiler/Scanner/Scanner.cs
@@ -381,7 +381,7 @@ namespace TypeCobol.Compiler.Scanner
                 if (concatenatedLine.Length > 0)
                 {
                     // Scan the continuation text, and get its last token so far
-                    TokensLine temporaryTokensLine = TokensLine.CreateVirtualLineForInsertedToken(firstSourceLine.LineIndex, concatenatedLine, firstSourceLine.ColumnsLayout == ColumnsLayout.FreeTextFormat ? ColumnsLayout.FreeTextFormat : ColumnsLayout.CobolReferenceFormatWithoutCommentText);
+                    TokensLine temporaryTokensLine = TokensLine.CreateVirtualLineForInsertedToken(firstSourceLine.LineIndex, concatenatedLine, firstSourceLine.ColumnsLayout == ColumnsLayout.FreeTextFormat ? ColumnsLayout.FreeTextFormat : ColumnsLayout.CobolReferenceFormatUnlimitedLength);
                     Scanner.ScanTokensLine(temporaryTokensLine, initialScanState, compilerOptions, copyTextNameVariations);
                     Token lastTokenOfConcatenatedLineSoFar = temporaryTokensLine.SourceTokens[temporaryTokensLine.SourceTokens.Count - 1];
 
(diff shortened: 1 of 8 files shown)


TypeCobol/Compiler/Text/ColumnsLayout.cs
```diff
@@ -15,6 +15,7 @@ public enum ColumnsLayout
         /// Columns 73-> : Comment
         /// </summary>
         CobolReferenceFormat,
+        CobolReferenceFormatWithoutCommentText,
```
efr15, line 18: If it is only dedicated to ContinuationLine, maybe we could rename it into `ContinuationLineFormat `?
I will discuss this with @fm-117

TypeCobol/Compiler/Scanner/Scanner.cs
```diff
@@ -470,7 +471,7 @@ public static void ScanTokensLineContinuationGroup(IList<TokensLine> continuatio
                             if (lastTokenOfConcatenatedLineSoFar.HasClosingDelimiter)
                             {
                                 bool continuationStartsWithTwoDelimiters = ((startOfContinuationIndex + 1) <= lastIndex && line[startOfContinuationIndex + 1] == lastTokenOfConcatenatedLineSoFar.ExpectedClosingDelimiter);
-                                bool previousDelimiterIsNotAtEnd = (lastTokenOfConcatenatedLineSoFar.EndColumn + CobolFormatAreas.Indicator) != CobolFormatAreas.End_B;
+                                bool previousDelimiterIsNotAtEnd = lastTokenOfConcatenatedLineSoFar.EndColumn != (int) CobolFormatAreas.End_B;
```
efr15, line 474: Could you please explain this modification?

efr15 (commented):
Ok with the modifications.
See my question + suggestion

TypeCobol/Compiler/Scanner/Scanner.cs
smedilol: Before:
The continuation line was the content of column 8 to 72 from the original line.
So the last column of the continuation line was 65.
That's why `CobolFormatAreas.Indicator` (value =7) was added to the the column of the last token to check if it ended at column 72.

After:
The continuation line is the content of column 1 to 72 from the original line.
So the last column of the continuation line is now also 72.
There is no need to add `CobolFormatAreas.Indicator` anymore.

TypeCobol/Compiler/Text/ColumnsLayout.cs
smedilol: @efr15 I suggest `CobolReferenceFormatForContinuationLine`

fm-117: `CobolReferenceFormatUnlimitedLength` ? I'm also ok with `CobolReferenceFormatForContinuationLine`. At any rate, add a comment on this special enum value.

fm-117 (changes requested):

TypeCobol/Compiler/Text/ColumnsLayout.cs
efr15: CobolReferenceFormatUnlimitedLength + comment added

TypeCobol/Compiler/Text/CobolTextLine.cs
```diff
@@ -262,6 +266,39 @@ private void MapVariableLengthLineWithReferenceFormat(out TextArea indicator, ou
                 }
             }
         }
+        private void MapVariableLengthLineWithReferenceFormatWithoutCommentText(out TextArea indicator, out TextArea source)
```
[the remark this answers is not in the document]
efr15: Done

fm-117 (approved):

PR #2572, opened 2023-09-01
WI #2569 Support REPLACE with LEADING/TRAILING

efr15:
Fixes #2569

Some additional blanks are sometimes added by the new Comparator `ProcessedTokens`.
Maybe related to a recent refactoring in [ScanStateTracker.GetCurrentScanState](url).

Another known bug in our parser explains that REPLACE with comma are not handled properly.

Issue #2569, opened 2023-08-30
Support REPLACE with LEADING/TRAILING


smedilol:
**What is the current problem**
TODO

Cobol code to illustrate.
```cobol

```

**Describe the solution you'd like**

Cobol code to illustrate.
```cobol

```

**Technical**
Will be edited by a team member.
Implementation details with the chosen solution.

**How to test automatically**
Will be edited by a team member.
Only if our current automated tests doesn't handle this.

PR #2568, opened 2023-08-24
WI #1942 Add sections in completion after PERFORM

efr15:
Fixes #1942

Modifications in **PerformCompletion.tlsp**:

- Section **Sec1** inserted before paragrap **Parag5**
- Section **Sec2** with paragraph **Parag6**  inserted after paragrap **Parag5**

Consequences:

- Diagnostic for **Parag5** is shifted by 2 lines (39 instead of 37)
- 3 additional completion items: **Parag6** and the 2 sections **Sec1** and **Sec2**

PR #2567, opened 2023-08-24
WI #2477 Check for 'replace separator' around token to replace

smedilol:
WI #2477 Check for 'replace separator' around token to replace
'replace separator' can be around token to replace or there can be at the beginning and end of the replacement token.

New ICompilationResultFormatter/Comparisons : ProcessedTokens.
This allow to see the source code with replaced token applied.

PR #2579, opened 2023-09-21
v2.1.0

fm-117:
# v2.1.0 changelog
This is a minor release focused on bug fix in `REPLACE` operations. It also contains support for Cobol 6.3 `REPOSITORY` paragraph and its simplified intrinsic function call syntax.

## New features
- WI #2536 Support REPOSITORY paragraph and optional keyword on intrinsic function calls (#2554)
- WI #2187 Check alignment of statements (#2566)
- WI #2569 Support REPLACE with LEADING/TRAILING (#2572)

## Bug fix
- WI #1942 Add sections in completion after PERFORM (#2568)
- WI #2316 Allow parsing of COPYs containing a PROCEDURE DIVISION (#2577)
- WI #2477 Check for 'replace separator' around token to replace (#2567)
- WI #2218 Preserve column and line while rescanning during a REPLACE (#2570)
- WI #2357 Preserve original column of virtual concatenated line (#2571)

## Internal
- WI #2548 Create a new LSP custom request for DataLayout view (#2563)
- WI #2236 Limit uses of ImmutableList<T> (#2575)


fm-117:
@Aubertov the purpose of this PR is to update `master` branch with commits from develop and then create a new release.
No need to review the code as everythig has already been validated, however the changelog needs to be reviewed to check that everything is easy to understand.

**These PRs must NOT be squashed as we want to keep the original commits from develop.**

Aubertov (approved):

PR #2601, opened 2023-12-19
WI #2600 Fix NullReferenceException in Workspace.CollectDataLayoutNodes

### LINKED ISSUE #2600: NullReferenceException in Workspace.CollectDataLayoutNodes
**Describe the bug**
A `NullReferenceException` may be thrown by `Workspace.CollectDataLayoutNodes` method.

**To Reproduce**
No source code to reproduce.

**Expected behavior**
No exception.

**Technical**
See
https://github.com/TypeCobolTeam/TypeCobol/blob/c1fdd30b2ef12a68a93d2e5ea205dd19212702fd/TypeCobol.LanguageServer/Workspace.cs#L1103
This line should be protected against null values.

**How to test automatically**
TODO

efr15:
Fixes #2600

To reproduce the crash in a Cobol program: remove the . after IDENTIFICATION DIVISION

### DIFF
--- a/TypeCobol.LanguageServer/Workspace.cs
+++ b/TypeCobol.LanguageServer/Workspace.cs
@@ -1100,7 +1100,7 @@ namespace TypeCobol.LanguageServer
         private List<Tuple<int, DataDefinition, int>> CollectDataLayoutNodes(CompilationUnit compilationUnit)
         {
             var dataLayoutNodes = new List<Tuple<int, DataDefinition, int>>();
-            Node dataDivision = compilationUnit.TemporaryProgramClassDocumentSnapshot.Root.MainProgram.GetChildren<DataDivision>().FirstOrDefault();
+            Node dataDivision = compilationUnit?.TemporaryProgramClassDocumentSnapshot?.Root?.MainProgram?.GetChildren<DataDivision>()?.FirstOrDefault();
             if (dataDivision != null)
             {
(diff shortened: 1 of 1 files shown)


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -1100,7 +1100,7 @@ static object GetEnd(long start, long length)
         private List<Tuple<int, DataDefinition, int>> CollectDataLayoutNodes(CompilationUnit compilationUnit)
         {
             var dataLayoutNodes = new List<Tuple<int, DataDefinition, int>>();
-            Node dataDivision = compilationUnit.TemporaryProgramClassDocumentSnapshot.Root.MainProgram.GetChildren<DataDivision>().FirstOrDefault();
+            Node dataDivision = compilationUnit?.TemporaryProgramClassDocumentSnapshot?.Root?.MainProgram?.GetChildren<DataDivision>()?.FirstOrDefault();
```
fm-117, line 1103: `compilationUnit`, `Root` and the result of `GetChildren<>` cannot be `null`. `TemporaryProgramClassDocumentSnapshot` should not be null either as we force node rebuilding. 

I'm ok to keep it like this safe approach but this will be changed if we move forward with the `#nullable` contexts.

fm-117 (approved):

PR #2607, opened 2024-01-09
WI #2605 Fix wrong URI management

### LINKED ISSUE #2605: Incorrect URI management in notifications PublishDiagnostics and MissingCopies
**Describe the bug**
When sending notifications `PublishDiagnostics `and `MissingCopies `to the client, the server does not provide escaped URI (for instance, blanks are not replaced by `%20`).
The client is consequently not able to retrieve the related document when receiving the notification.

**To Reproduce**
Open a document with Cobol syntax errors from a Repository Workspace containing some blanks: no diagnostics are shown.

**Expected behavior**
The diagnostics should be visible.

**Technical**
When creating the URI parameter in the notification, the server uses `ToString `on the original URI which may result in a wrong value (blanks are not escaped by `ToString`). 

**How to test automatically**
Not relevant

See also #2606

efr15:
Fixes #2605

Side effects: LSR tests must be adapted accordingly (URI 'file:///...' changed to 'file:/...')


TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -239,6 +239,18 @@ protected DocumentContext GetDocumentContextFromStringUri(string uri, Workspace.
...
+        /// Gets the original URI (which was set by the client)
+        /// DON'T use ToString() as it returns the canonically unescaped form of the URI
+        /// (it may cause issue if the path contains some blanks which need to be escaped)
+        /// </summary>
+        /// <param name="uri"></param>
+        /// <returns></returns>
+        private static string GetOriginalUri(Uri uri)
+        {
+            return uri.OriginalString;
+        }
```
fm-117, line 252: Please move that to https://github.com/TypeCobolTeam/TypeCobol/blob/develop/TypeCobol.LanguageServer/Utilities/Extensions.cs and make it extension an extension method.

fm-117: I think it's better to create a new constructor for `TextDocumentIdentifier` accepting an `Uri` parameter and call `OriginalString` in it. Duplicate the call to `OriginalString` for the `diagParameter.uri` assignment and remove this method.
You can keep the comment about using the original value sent by the client in the new constructor.

fm-117 (changes requested):

efr15:
Adding a new constructor in `TextDocumentIdentifier `makes all LSR tests crash.

So I have implemented the extension method as proposed first.

fm-117:
> Adding a new constructor in `TextDocumentIdentifier `makes all LSR tests crash.
> 
> So I have implemented the extension method as proposed first.

It works but you have to declare the constructor as `internal`.

fm-117 (changes requested):
See my remark on the constructor accessibility

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -11,6 +11,7 @@
 using TypeCobol.Compiler.Preprocessor;
 using TypeCobol.LanguageServer.Context;
 using TypeCobol.LanguageServer.SignatureHelper;
+using TypeCobol.LanguageServer.Utilities;
```
fm-117, line 14: Not needed

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/TextDocumentIdentifier.cs
```diff
@@ -23,5 +23,17 @@ public TextDocumentIdentifier(string uri)
         {
             this.uri = uri;
         }
+
+        /// <summary>
+        /// Creates a new TextDocumentIdentifier literal.
+        /// @param uri The document's uri.
```
fm-117, line 29: Java syntax, we're speaking C# here 💥

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolServer.cs
efr15: Removed

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/TextDocumentIdentifier.cs
efr15: Changed here (but there are many @param in other classes)

fm-117 (approved):

PR #2638, opened 2024-07-30
WI #2636 Implement refactor/adjustFillers command

### LINKED ISSUE #2636: Implement new command AdjustFillers
The command `refactor/adjustFillers` traverse the AST to find and adjust FILLERs that are inside REDEFINES. The goal is to change each FILLER so as the REDEFINES size matches the redefined area size.

This will be the first of our COBOL refactorings.

The issue #2634 is a more general issue about having all the necessary messages in the protocol to support refactorings properly.

fm-117:
Fixes #2636

As draft for now as we need to discuss about LSR testing for this new requests.


TypeCobol.LanguageServer/Commands/AdjustFillers.cs
```diff
@@ -0,0 +1,156 @@
...
+            var target = Server.GetDocumentContextFromStringUri(textDocumentIdentifier.uri, Workspace.SyntaxTreeRefreshLevel.RebuildNodes);
+            var programClassDocument = target.FileCompiler?.CompilationResultsForProgram?.ProgramClassDocumentSnapshot;
+            if (programClassDocument == null)
+            {
+                throw new InvalidOperationException($"Could not get AST for document '{target.Uri}'.");
+            }
+
+            // Compute edits asynchronously: create a message for ourselves
+            Server.Workspace.MessagesActionsQueue.Enqueue(new MessageActionWrapper(() => ComputeTextEdits(target.Uri, programClassDocument)));
+            return null;
```
efr15, line 138: As discussed, it would be better to return `new object()` so that the client is able to make a difference between a successful command and an error.

efr15 (changes requested):

efr15 (dismissed):

TypeCobol.LanguageServer/Commands/AdjustFillers.cs
fm-117: done

efr15 (approved):

PR #2649, opened 2024-08-30
2647 adjust filler may insert filler at bad location

### LINKED ISSUE #2647: Adjust FILLER may insert FILLER at bad location
**Describe the bug**
When the data after which the FILLER will be inserted has children the FILLER is inserted before the data's children.
The FILLER should be inserted after the data's last child.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOZADJ.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 target-zone PIC X(100).
       01 redef REDEFINES target-zone.
          05 group1.
             10 group11.
                15 group111.
                   20 var1 PIC X(25).

       PROCEDURE DIVISION.
            GOBACK
            .
       END PROGRAM TCOZADJ.
```
When applying the refactoring on this document, FILLER is inserted between `redef `and `group1`.

**Expected behavior**
FILLER should be inserted after `var1`.

**Technical**
`AdjustFillerVisitor `should consider possible children when determining where to insert.
Warning: INDEX children should be ignored.

**How to test automatically**
LSR tests.

efr15:
Fixes #2647

Also ignore INDEX among children to avoid a crash when retrieving the CodeElement (which is null for an Index node)

### DIFF
--- a/TypeCobol.LanguageServer/Commands/AdjustFillers.cs
+++ b/TypeCobol.LanguageServer/Commands/AdjustFillers.cs
@@ -18,7 +18,7 @@ namespace TypeCobol.LanguageServer.Commands
                     return true;
                 }
 
-                if (HasNoChildrenExcludingIndex(dataRedefines))
+                if (!dataRedefines.HasChildrenExcludingIndex())
                 {
                     // Inline REDEFINES, nothing to do
                     return true;
@@ -93,7 +93,7 @@ namespace TypeCobol.LanguageServer.Commands
                         string newText = $"{Environment.NewLine}{new string(' ', indent)}{level} FILLER PIC X({adjustedFillerSize}).";
 
                         // Look for the last node (last child itself or the last node of its descendants for a group)
-                        var lastNode = GetLastDescendant(lastChild);
+                        var lastNode = lastChild.GetLastNode();
 
                         // Insert at the end of last node line
                         Debug.Assert(lastNode.CodeElement != null);
@@ -109,17 +109,6 @@ namespace TypeCobol.LanguageServer.Commands
                     {
                         textEdits.Add(textEdit);
                     }
-
-                    DataDefinition GetLastDescendant(DataDefinition dataDefinition)
-                    {
-                        if (HasNoChildrenExcludingIndex(dataDefinition))
-                        {
-                            return dataDefinition;
-                        }
-
-                        Debug.Assert(dataDefinition.Children[^1] is DataDefinition);
-                        return GetLastDescendant((DataDefinition)dataDefinition.Children[^1]);
-                    }
                 }
 
                 void Remove()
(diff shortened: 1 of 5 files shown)


TypeCobol.LanguageServer/Commands/AdjustFillers.cs
```diff
@@ -105,6 +109,17 @@ void AddOrAdjust()
...
+                    DataDefinition GetLastDescendant(DataDefinition dataDefinition)
+                    {
+                        if (HasNoChildrenExcludingIndex(dataDefinition))
+                        {
+                            return dataDefinition;
+                        }
+
+                        Debug.Assert(dataDefinition.Children[^1] is DataDefinition);
+                        return GetLastDescendant((DataDefinition)dataDefinition.Children[^1]);
+                    }
```
fm-117, line 122: No need to bother with the node's type here. I agree that it should be data definitions all the way but I'm worried with EXEC nodes and some ill-formed AST.

```csharp
Node GetLastNode(Node node)
{
    if (HasNoChildrenExcludingIndex(node))
    {
        return node;
    }

    return GetLastNode(node.Children[^1]);
}
```

should work as well

TypeCobol.LanguageServer/Commands/AdjustFillers.cs
```diff
@@ -115,6 +130,11 @@ void Remove()
                     var textEdit = new TextEdit(new VsCodeProtocol.Range() { start = start, end = end }, string.Empty);
                     textEdits.Add(textEdit);
                 }
+
+                bool HasNoChildrenExcludingIndex(Node node)
+                {
+                    return node.ChildrenCount == node.Children.OfType<IndexDefinition>().Count();
```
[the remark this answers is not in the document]
fm-117: Will move with `GetLastNode`, so turn into a property on Node object

TypeCobol.LanguageServer/Commands/AdjustFillers.cs
fm-117: Also there is a `GetNextNode` method on `Node` object, so this new helper could be moved there for future uses.

fm-117 (changes requested):

TypeCobol.LanguageServer/Commands/AdjustFillers.cs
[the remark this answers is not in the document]
efr15: Done

TypeCobol.LanguageServer/Commands/AdjustFillers.cs
efr15: Done

TypeCobol/Compiler/Nodes/Node.cs
```diff
@@ -941,6 +941,26 @@ public virtual Node GetNextNode()
...
+        }
+
+        /// <summary>
+        /// Indicate whether this node has 'real' children (from a semantic point of view) by excluding indexes.
+        /// Note: indexes are created as children nodes for technical reasons (especially Symbol Table management).
+        /// </summary>
+        /// <returns></returns>
+        public bool HasChildrenExcludingIndex()
+        {
+            return Children.Any(child => child is not IndexDefinition);
```
fm-117, line 962: Ok this is consistent with the name however this does not guarantee that the `CodeElement` will be non-null. `IndeDefinition`, `GeneratedDefinition`, `Sentence`, `Then` and `WhenGroup` are special nodes without corresponding code element.

fm-117 (approved):

PR #2657, opened 2024-09-09
WI #2656 Use ConsoleLogger as last resort for pending logs on exit

### LINKED ISSUE #2656: Fix random build errors
Here on GitHub Actions, our build may randomly fail because of pending logging operations.

See https://github.com/TypeCobolTeam/TypeCobol/blob/1e76f900cd118f90fa7874b190a832fb67d24103/TypeCobol/Logging/LoggingSystem.cs#L78

The assertion here was designed to signal possible log drops but it seems to be frequently failing on GitHub CI, so as the code is called when application is exiting and nothing can be done, we can just replace the assertion with a console log.

fm-117:
Fixes #2656

### DIFF
--- a/TypeCobol/Logging/LoggingSystem.cs
+++ b/TypeCobol/Logging/LoggingSystem.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Threading;
+﻿using System.Collections.Concurrent;
 
 namespace TypeCobol.Logging
 {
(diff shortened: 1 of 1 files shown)


TypeCobol/Logging/LoggingSystem.cs
```diff
@@ -66,16 +63,25 @@ public void Stop()
...
 
-                System.Diagnostics.Debug.Assert(_work.Count == 0);
+                // Last chance ! Log errors into the console if we did not have enough time to use configured loggers
+                if (!_work.IsEmpty)
+                {
+                    var remainingWork = _work.ToArray();
+                    var consoleLogger = new ConsoleLogger(LogLevel.Error); // Drop infos and warnings, keep errors and exceptions
+                    foreach (var action in remainingWork)
+                    {
+                        action(consoleLogger);
```
efr15, line 82: Maybe the action call should be surrounded with a try/catch (doing a Console.Error of the exception)

efr15 (changes requested):

TypeCobol/Logging/LoggingSystem.cs
fm-117: You're right. Done.

efr15 (approved):

PR #2660, opened 2024-09-20
WI #2659 Test LSP objects

### LINKED ISSUE #2659: Test LSP objects
Our implementation of LSP is quite old now and more importantly is embedded instead of relying on a well-known externally-provided package.

Migrating to `Microsoft.VisualStudio.LanguageServer.Protocol` NuGet package is the target however this will be difficult as we have accumulated a lot of gaps over time.

The goal of this issue is to create tests covering the form of protocol objects. This will allow:
- to identify discrepancies between our private client definitions and server definitions
- to identify discrepancies between multiple version of the protocol objects (when migrating to VS package)

In a nutshell the protocol test should:
- automatically discover the messages known by this server
- verify that each message is actually covered by a test
- test every message using predefined JSON test files

fm-117:
Fixes #2659


TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -187,8 +168,13 @@ private void MissingCopiesDetected(object fileUri, MissingCopiesEvent missingCop
             //Warning the main file could not be opened
             //This event can be used when a dependency have not been loaded
 
+            // Get the original URI (which was set by the client)
+            // DON'T use ToString() as it returns the canonically unescaped form of the URI
+            // (it may cause issue if the path contains some blanks which need to be escaped)
+            string uri = ((Uri)fileUri).OriginalString;
```
efr15, line 174: In my opininon, the previous implementation with the dedicated constructor was safer (because the usage of `URI.OriginalString` was centralized)
I think we should keep something similar.
If it's not possible to restore the constructor, maybe a builder `BuildFrom(URI)`

TypeCobol.LanguageServer/VsCodeProtocol/Initialize Method/InitializeParams.cs
```diff
@@ -26,5 +26,10 @@ public class InitializeParams
         /// The capabilities provided by the client (editor)
         /// </summary>
         public ClientCapabilities capabilities { get; set; }
+
+        /// <summary>
+        /// The initial trace setting. If omitted trace is disabled ('off').
+        /// </summary>
+        public string trace { get; set; }
```
efr15, line 33: Should it be `nullable `to comply with LSP specification?

TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -13,6 +13,7 @@
 using System.Collections.Concurrent;
 using TypeCobol.Compiler.CodeElements;
 using TypeCobol.Compiler.Nodes;
+using TypeCobol.LanguageServer.TypeCobolCustomLanguageServerProtocol;
```
efr15, line 16: From my point of view (regarding its description), it is a little bit strange that `Workspace `refers to `TypeCobolCustomLanguageServerProtocol `(by the way that introduces a circular reference).
What do you think?

efr15 (changes requested):

TypeCobol.LanguageServer.Test/ProtocolTests/TestSet.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        public void Validate(Dictionary<string, LspMethodDefinition> methods)
+        {
+            // Detect untested methods: methods that are known by this server but without any associated test file
+            // and unknown methods: methods having a test file but not known by this server
+            var untested = new Dictionary<string, LspMethodDefinition>(methods);
+            var unknown = new List<string>();
+            var messagesByMethod = this.ToLookup(testMessage => testMessage.Method);
```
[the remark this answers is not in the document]
fm-117: You're right. At first I wanted to control more things by also iterating over the messages but in this version we don't need them.

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/TextEdit.cs
```diff
@@ -22,40 +22,34 @@ public class TextEdit
...
-            this.newText = newText;
-        }
-
         /// <summary>
         /// Creates a replace text edit.
         /// @param range The range of text to be replaced.
         /// @param newText The new text.
         /// </summary>
-        public static TextEdit replace(Range range, string newText)
+        public static TextEdit Replace(Range range, string newText)
```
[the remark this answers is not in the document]
fm-117: Ok

TypeCobol.LanguageServer/JsonRPC/NotificationType.cs
```diff
@@ -1,26 +1,29 @@
...
             this.ParamsType = paramsType;
         }
+
+        public override MethodKind Kind => MethodKind.NotificationMethod;
+
+        public override IEnumerable<Type> Types
+        {
+            get
+            {
+                yield return ParamsType;
```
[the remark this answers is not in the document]
fm-117: Not only useful but required. As `Types` property should return an `IEnumerable<Type>` the body of the getter should either return an instance of an Array or List or use `the `yield return` statement. Did you mean `return new[] { ParamsType }; `?

TypeCobol.LanguageServer/TypeCobolCustomLanguageServer/SyntaxColoring/SyntaxColoringParams.cs
```diff
@@ -16,7 +10,7 @@ public class SyntaxColoringParams
         /// <summary>
         /// The Text Document Identifier
         /// </summary>
-        public TextDocumentItem textDocument;
+        public TextDocumentIdentifier textDocument;
```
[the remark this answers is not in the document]
fm-117: `TextDocumentItem` is used by the client to describe a document not known yet, it contains the source code (property `text`). After receiving a `TextDocumentItem`, the server will know the document by its uri and both client and server will use `TextDocumentIdentifier` to refer to it. Here we are in the parameters for the `SyntaxColoring` notification which is sent by the server to the client. The source code should not be sent back to the client, only the uri. The use of `TextDocumentItem` was a mistake in original code.

TypeCobol.LanguageServer/TypeCobolLanguageServer.cs
```diff
@@ -184,13 +187,19 @@ public RefreshOutlineParams UpdateOutline(ProgramClassDocument programClassDocum
...
                 }
 
                 if (bForced || _rootOutlineNode.Update(programClassDocument.Root))
-                    return new RefreshOutlineParams(new TextDocumentIdentifier(this.LspTextDocument.uri),
-                        _rootOutlineNode);
+                {
+                    return new RefreshOutlineParams()
+                    {
+                        textDocument = new TextDocumentIdentifier() { uri = this.LspTextDocument.uri },
+                        outlineNodes = _rootOutlineNode.childNodes
```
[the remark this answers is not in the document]
fm-117: As I removed the constructor for `RefreshOutlineParams`, the code here does the same thing. This is because the params expect a list of nodes whereas the class works with a 'fake' root node.

TypeCobol.LanguageServer/TypeCobolServer.cs
fm-117: Ok for the builder method

TypeCobol.LanguageServer/VsCodeProtocol/Initialize Method/InitializeParams.cs
fm-117: Well it is. The nullable context is deactivated here so this `string` accepts `null` value.

TypeCobol.LanguageServer/Workspace.cs
fm-117: You're right, I'll refactor that a bit.

TypeCobol.LanguageServer/JsonRPC/NotificationType.cs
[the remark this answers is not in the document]
efr15: OK, no need to change

efr15 (approved):

efr15 (dismissed):

TypeCobol.LanguageServer.Test/RefactoringTest.cs
```diff
@@ -87,30 +87,30 @@ public void CheckTarget(CompilationUnit compilationUnit)
...
-                var deleteMultiLines = TextEdit.del(new Range(10, 22, 12, 22));
+                var deleteMultiLines = TextEdit.Del(Range.FromPositions(10, 22, 12, 22));
                 textEdits.Add(deleteMultiLines);
 
                 builder.Clear();
                 builder.AppendLine("has been");
                 builder.AppendLine("      * modified and the modification spans");
                 builder.Append("      * over multiple lines ");
-                var modifyMultiLines = TextEdit.replace(new Range(14, 22, 16, 22), builder.ToString());
+                var modifyMultiLines = TextEdit.Replace(new Position() { line = 14, character = 22 }, new Position() { line = 16, character = 22 }, builder.ToString());
```
efr15, line 113: `Range.FromPositions` could be used here

efr15 (changes requested):

efr15 (approved):

PR #2661, opened 2024-09-23
WI #2659 Fix LSP incompatibilities

efr15:
Some LSP classes are not compatible with the specifications of LSP 3.17

### DIFF
--- a/TypeCobol.LanguageServer/TypeCobolServer.cs
+++ b/TypeCobol.LanguageServer/TypeCobolServer.cs
@@ -435,7 +435,7 @@ namespace TypeCobol.LanguageServer
                 var vtdi = new VersionedTextDocumentIdentifier(parameters.textDocument.uri, 0);
                 TextDocumentContentChangeEvent tdcce = new TextDocumentContentChangeEvent();
                 tdcce.text = parameters.text;
-                DidChangeTextDocumentParams dctdp = new DidChangeTextDocumentParams(vtdi, new[] { tdcce });
+                DidChangeTextDocumentParams dctdp = new DidChangeTextDocumentParams { textDocument = vtdi, contentChanges = new[] { tdcce } };
                 OnDidChangeTextDocument(dctdp);
             }
         }
--- a/TypeCobol.LanguageServer/TypeCobolServer.cs
+++ b/TypeCobol.LanguageServer/TypeCobolServer.cs
@@ -370,11 +370,12 @@ namespace TypeCobol.LanguageServer
 
         protected override void OnDidChangeTextDocument(DidChangeTextDocumentParams parameters)
         {
-            var docContext = GetDocumentContextFromStringUri(parameters.uri, Workspace.SyntaxTreeRefreshLevel.NoRefresh); //Text Change do not have to trigger node phase, it's only another event that will do it
+            var uri = parameters.textDocument.uri;
+            var docContext = GetDocumentContextFromStringUri(uri, Workspace.SyntaxTreeRefreshLevel.NoRefresh); //Text Change do not have to trigger node phase, it's only another event that will do it
             if (docContext == null)
                 return;
 
-            Uri objUri = new Uri(parameters.uri);
+            Uri objUri = new Uri(uri);
             var updates = new RangeUpdate[parameters.contentChanges.Length];
             
             // Convert change events into updates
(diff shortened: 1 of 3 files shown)


fm-117:
Removed the closing keyword as there is more to fix for #2659

TypeCobol.LanguageServer/VsCodeProtocol/Text Document Notifications/DidChangeTextDocumentParams.cs
```diff
@@ -8,10 +8,21 @@ namespace TypeCobol.LanguageServer.VsCodeProtocol
...
+
         public TextDocumentContentChangeEvent[] contentChanges { get; set; }
 
-        public DidChangeTextDocumentParams(string uri) : base(uri) { }
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="textDocument">The document that did change.</param>
+        /// <param name="contentChanges">The actual content changes.</param>
+        public DidChangeTextDocumentParams(VersionedTextDocumentIdentifier textDocument, TextDocumentContentChangeEvent[] contentChanges)
```
fm-117, line 22: For protocol objects I prefer not to define any constructor, this ensure the deserializer will always find a way to create an instance. Use object initializer to create and define properties in one go.

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/VersionedTextDocumentIdentifier.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// The version number of this document.
+        /// </summary>
+        public int? version { get; set; }
+
+        /// <summary>
+        /// Creates a new versioned TextDocumentIdentifier.
+        /// </summary>
+        /// <param name="uri">The document's uri.</param>
+        /// <param name="version">The version number of this document.</param>
+        public VersionedTextDocumentIdentifier(string uri, int? version) : base(uri) { this.version = version; }
```
fm-117, line 18: Same remark as above

TypeCobol.LanguageServer/TypeCobolServer.cs
```diff
@@ -431,10 +432,10 @@ protected override void OnDidSaveTextDocument(DidSaveTextDocumentParams paramete
         {
             if (parameters.text != null)
             {
-                DidChangeTextDocumentParams dctdp = new DidChangeTextDocumentParams(parameters.textDocument.uri);
+                var vtdi = new VersionedTextDocumentIdentifier(parameters.textDocument.uri, 0);
```
fm-117, line 435: using `null` for version seems more appropriate as our server does not handle this field yet. Which value is sent by the client on a DidChangeText ?

fm-117 (changes requested):

TypeCobol.LanguageServer/TypeCobolServer.cs
efr15: The client aways sends 0 for version.

fm-117: Ok we can keep 0 here too then

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/VersionedTextDocumentIdentifier.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// An identifier to denote a specific version of a text document.
+    /// </summary>
+    public class VersionedTextDocumentIdentifier : TextDocumentIdentifier
+    {
+        /// <summary>
+        /// The version number of this document.
+        /// </summary>
+        public int? version { get; set; }
```
fm-117, line 11: Could you remove the `?` ? This is not nullable in LSP specs.

fm-117 (changes requested):

TypeCobol.LanguageServer/VsCodeProtocol/Text Document Notifications/DidChangeTextDocumentParams.cs
efr15: Done

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/VersionedTextDocumentIdentifier.cs
efr15: It's currently not possible to have no constructor because `TextDocumentIdentifier` has one.

TypeCobol.LanguageServer/VsCodeProtocol/Text Document/VersionedTextDocumentIdentifier.cs
efr15: Done

fm-117 (approved):

PR #2665, opened 2024-10-07
WI #2664 Ignore change notifications for closed projects

### LINKED ISSUE #2664: Limit "Copy to WorkspaceProject mismatch" warnings
In our logs we observe many "Copy to WorkspaceProject mismatch" warnings. As the `workspace/didChangeWatchedFiles` notification is sent by our client reacting to file system changes, the change notifications may arrive after the corresponding `WorkspaceProject` has been closed.

While it does not fully explain the cause of the out-of-sync messages, it is safe to ignore notifications for files we are not tracking anymore server-side. We can still keep a log trace for unknown projects, meaning projects whose key has never been seen by the server.

fm-117:
Fixes #2664

- Keep names of all opened `WorkspaceProject` instances
- Remove useless `TryAdd` call when creating neww `WorkspaceProject`

### DIFF
--- a/TypeCobol.LanguageServer/Workspace.cs
+++ b/TypeCobol.LanguageServer/Workspace.cs
@@ -788,7 +788,8 @@ namespace TypeCobol.LanguageServer
                 else if (!WorkspaceProjectStore.IsKnownProject(changedCopy.OwnerProject))
                 {
                     // Inconsistent notification from client, the target project is unknown
-                    LoggingSystem.LogMessage(LogLevel.Error, $"Copy to WorkspaceProject mismatch: project '{changedCopy.OwnerProject}' is unknown.");
+                    var contextData = new Dictionary<string, object>() { { "WorkspaceProjectId", changedCopy.OwnerProject } };
+                    LoggingSystem.LogMessage(LogLevel.Error, "Copy to WorkspaceProject mismatch: unknown project.", contextData);
                 }
                 // else this is a known project but already closed: ignore
             }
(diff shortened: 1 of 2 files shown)


TypeCobol.LanguageServer/Workspace.cs
```diff
@@ -785,11 +785,12 @@ internal void AcknowledgeCopyChanges(List<ChangedCopy> changedCopies)
                     // Evict from target project cache only
                     evictions[workspaceProject.Project.CopyCache].Add(changedCopy.CopyName);
                 }
-                else
+                else if (!WorkspaceProjectStore.IsKnownProject(changedCopy.OwnerProject))
                 {
-                    // Inconsistent notification from client, the target project could not be found
-                    LoggingSystem.LogMessage(LogLevel.Warning, $"Copy to WorkspaceProject mismatch: could not find project '{changedCopy.OwnerProject}'.");
+                    // Inconsistent notification from client, the target project is unknown
+                    LoggingSystem.LogMessage(LogLevel.Error, $"Copy to WorkspaceProject mismatch: project '{changedCopy.OwnerProject}' is unknown.");
```
efr15, line 791: Maybe we could have a fixed error message and the variable part (with project key) in the context data
It will be easier then to filter on this message when analyzing logs.

efr15 (changes requested):

TypeCobol.LanguageServer/Workspace.cs
fm-117: Yes ! Done.

efr15 (approved):

PR #2673, opened 2024-10-30
WI #2670 Use telemetry/event to send server exceptions to client

### LINKED ISSUE #2670: Convert window/showMessage notifications to telemetry/event
Child of #2668.

This is a minimalistic approach to LS exception tracking :
- in `TypeCobolServerHost`, as soon as the `TypeCobolServer` instance is available , add a new `ILogger`.
  - this new logger would be responsible for sending a `telemetry/event` as described in #2668 when the `LogException` method is called. It will be naturally dependent from the `TypeCobolServer` as it requires to send JSON to the client.
- remove the call to `ShowErrorMessage` from `NotifyException` and move it to `UnhandledExceptionHandler`, to ensure a message is still displayed in the event of a server crash.

fm-117:
Fixes #2670

No new LSP tests:
- since `telemetry/event` is not tied to any document, I suspect LSR will drop the message
- the send happens on the logger thread so the message order is not guaranteed

I manually tested on our private client and the event is correctly sent and received.
Capture

### DIFF
--- a/TypeCobol.LanguageServer/VsCodeProtocol/Telemetry Notification/TelemetryEvent.cs
+++ b/TypeCobol.LanguageServer/VsCodeProtocol/Telemetry Notification/TelemetryEvent.cs
@@ -1,0 +1,57 @@
+﻿using System.Text;
+
+namespace TypeCobol.LanguageServer.VsCodeProtocol
+{
+    /// <summary>
+    /// Class used to describe an event that occured on this server.
+    /// </summary>
+    public class TelemetryEvent
+    {
+        /// <summary>
+        /// Extract data from an exception to produce a meaningful TelemetryEvent.
+        /// The event will contain :
+        /// - the exception's .NET type
+        /// - the exception's message
+        /// - the exception's target site (which is the method that threw the exception, top of the StackTrace)
+        /// </summary>
+        /// <param name="exception">Non-null exception instance to describe.</param>
+        /// <returns>Non-null instance of TelemetryEvent.</returns>
+        public static TelemetryEvent CreateFrom(Exception exception)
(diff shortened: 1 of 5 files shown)


TypeCobol.LanguageServer/VsCodeProtocol/Telemetry Notification/TelemetryEvent.cs
```diff
@@ -0,0 +1,57 @@
...
+            if (exception.TargetSite != null)
+            {
+                var targetSite = exception.TargetSite;
+                exceptionTargetSite.Append(targetSite.DeclaringType?.FullName); // Fullname of the type declaring the method
+                exceptionTargetSite.Append('.');
+                exceptionTargetSite.Append(targetSite.Name);
+                exceptionTargetSite.Append('(');
+                exceptionTargetSite.AppendJoin(", ", targetSite.GetParameters().Select(p => p.ParameterType.Name)); // For each param, simple name of its type
+                exceptionTargetSite.Append(')');
+            }
```
efr15, line 33: Maybe the `targetSite.Module` could be interesting as well?

efr15 (changes requested):

TypeCobol.LanguageServer/VsCodeProtocol/Telemetry Notification/TelemetryEvent.cs
fm-117: Yes this will help us group exceptions. However I prefer using the `Exception.Source` property instead of adding somtheing in the `TargetSite` string description

efr15 (approved):

PR #2680, opened 2024-11-21
WI #2679 Use platform-independent code to start processes

### LINKED ISSUE #2679: Use platform-independent code to start processes
- When launching a .NET executable, always use `dotnet <path to dll>`
- Always use ArgumentsList, instead of concatenating arguments
- Always use `Path.Combine` and avoid use of `'\'`

fm-117:
Fixes #2679

For CLI:
- the arguments file now separates each arg on its own line

For Transform:
- same thing but as we have multiple runs in a single file, use an empty line to separate arguments sets

For LanguageServer
- convert the `-ro` option to a file path because there is no proper syntax to pass arguments _inside_ argument


CLI/test/CLITest.cs
```diff
@@ -477,22 +471,25 @@ internal static string Test(string workingDirectory, string arguments, ReturnCod
...
             startInfo.WorkingDirectory = workingDirectory;
-            // Find TypeCobol.CLI.exe location
+            // Find TypeCobol.CLI.dll location
             string currentDirectory = Environment.CurrentDirectory;
             string configuration = Path.GetFileName(currentDirectory);
             string pathToExe = Path.Combine(currentDirectory, "..", "..", "..", "..");
             pathToExe = Path.GetFullPath(pathToExe);
-            pathToExe = Path.Combine(pathToExe, "CLI", "src", "bin", configuration, "TypeCobol.CLI.exe");
-            startInfo.Arguments = @"/c " + pathToExe + " " + arguments;
+            pathToExe = Path.Combine(pathToExe, "CLI", "src", "bin", configuration, "TypeCobol.CLI.dll");
```
efr15, line 481: Maybe we can rename `pathToExe `to `cliPath`?

TypeCobol.LanguageServer/TypeCobolServerHost.cs
```diff
@@ -191,7 +208,7 @@ static int Main(string[] args)
                 "",
                 "DESCRIPTION:",
                 "  Run the Language Server Robot.",
-                { "l|loglevel=",  "Logging level (1=Lifecycle, 2=Message, 3=Protocol).", (string v) =>
+                { "l|loglevel=",  "Logging level (0=Lifecycle, 1=Message, 2=Protocol).", (string v) =>
```
efr15, line 211: Are we sure there is no other impact?
For instance, what about Language Server Robot Monitor?

TypeCobol.Transform.Test/TransformTest.cs
```diff
@@ -49,33 +68,31 @@ internal static void Test(string workingDirectory, string arguments)
...
-                if (line_args.Length > 0)
+                System.Diagnostics.Process process = new System.Diagnostics.Process();
+                System.Diagnostics.ProcessStartInfo startInfo = new System.Diagnostics.ProcessStartInfo();
+                startInfo.WindowStyle = System.Diagnostics.ProcessWindowStyle.Hidden;
+                startInfo.FileName = "dotnet";
+                startInfo.WorkingDirectory = workingDirectory;
+                // Find TypeCobol.Transform.dll location
+                string currentDirectory = Environment.CurrentDirectory;
+                string configuration = Path.GetFileName(currentDirectory);
+                string pathToExe = Path.Combine(currentDirectory, "..", "..", "..");
```
efr15, line 82: Maye we can rename `pathToExe` to `transformPath`?

efr15 (changes requested):

CLI/test/CLITest.cs
fm-117: done

TypeCobol.Transform.Test/TransformTest.cs
fm-117: done

TypeCobol.LanguageServer/TypeCobolServerHost.cs
fm-117: Fixing the usage message has most certainly no consequences. What matters however are the values used for the parameter. I've changed the level used during tests to match the correct value for `Protocol` both for the server and the robot. When the server (or the robot) is launched by our client we do not use the option and let the server (and the robot) use the default logging level.

efr15 (approved):

Issue #2681, opened 2024-11-28
Ensure return codes are in [0-255] range


fm-117:
Unix/Linux uses a single byte to store processes return codes, therefore return codes are limited to the [0-255] range.
The goal of this issue is to change all return codes and define them in the proper range.

PR #2682, opened 2024-11-28
WI #2681 Change return codes for CLI / LanguageServer / Transform

fm-117:
Note that this PR contains breaking changes !

Clients testing return codes further than a comparison to 0 must be adapted.


efr15 (changes requested):
Comment below should be adapted:
https://github.com/TypeCobolTeam/TypeCobol/blob/b6bd9643604a85e72beeaf4a6a86590334e2b2a5/TypeCobol.Transform/Transform.cs#L63C16-L63C166

fm-117:
> Comment below should be adapted: https://github.com/TypeCobolTeam/TypeCobol/blob/b6bd9643604a85e72beeaf4a6a86590334e2b2a5/TypeCobol.Transform/Transform.cs#L63C16-L63C166

image

efr15 (approved):

PR #2687, opened 2024-12-11
WI #2686 Implement tests outside LSP for GetDataLayout

### LINKED ISSUE #2686: Implement tests outside LSP for GetDataLayout
The idea is to be able to test the notification GetDataLayout outside LSP.

Code should be externalized from `Workspace`. We put it into a `Processor` (by analogy to `IRefactoringProcessor`).

We add a simple test framework to read a COBOL source file, call the `Processor` and compare the actual result to an expected one (store in a file).

efr15:
Fixes #2686

### DIFF
--- a/TypeCobol.LanguageServer.Test/ProcessorTests/DataLayoutProcessorCSVTest.cs
+++ b/TypeCobol.LanguageServer.Test/ProcessorTests/DataLayoutProcessorCSVTest.cs
@@ -1,0 +1,54 @@
+﻿using System.Runtime.CompilerServices;
+using System.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TypeCobol.Compiler;
+using TypeCobol.Test;
+using TypeCobol.Test.Utils;
+
+namespace TypeCobol.LanguageServer.Test.ProcessorTests
+{
+    /// <summary>
+    /// Tests for DataLayout processor with CSV
+    /// </summary>
+    [TestClass]
+    public class DataLayoutProcessorCSVTest
+    {
+        private const string RELATIVE_PATH = "DataLayout";
+        private const string ROOT_PATH = "ProcessorTests";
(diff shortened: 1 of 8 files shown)


TypeCobol.LanguageServer.Test/ProcessorTests/DataLayoutProcessorCSVTest.cs
```diff
@@ -0,0 +1,54 @@
...
+        [TestMethod]
+        public void TestSimple() => DoTestProcessor("simple", "CSV-simple", ExecuteProcessor);
+
+        [TestMethod]
+        public void TestStackedAndNested() => DoTestProcessor("stackedAndNested", "CSV-stackedAndNested", ExecuteProcessor);
+        
+        private (string Header, string[] Rows) ExecuteProcessor(CompilationUnit compilationUnit)
+        {
+            return _processor.GetDataLayoutAsCSV(compilationUnit, ";");
+        }
```
fm-117, line 52: I think this could be factorized a bit more:
- the `DoTestProcessor` method could do its job by requesting only a test name (obtained automatically with the caller member name) and an object encapsulating the prefix to use to get the result file name from the input file and the processing itself
- also the class should be renamed, the goal is to test all variants of DataLayout not only CSV, so the class is `DataLayoutProcessorTest` while the method/object is `ExecuteCSVProcessor`

fm-117 (changes requested):

TypeCobol.LanguageServer.Test/ProcessorTests/DataLayoutProcessorCSVTest.cs
fm-117: Nevermind, we could refactor all this after adding the TREE tests

fm-117 (approved):

Issue #2674, opened 2024-11-15
Make our build multi-platform compatible


fm-117:
Top-level issue to group all adaptations required to make our build process compatible with Linux.
We are currently tied to Windows, the goal is to be able to build (and run) on Linux. MacOS is not a  target.

- [x] Get rid of external scripts (#2675)
- [x] #2679
- [ ] Check uses of `'\'` character in paths
- [ ] Paths are case-sensitive on Linux and case-insensitive on Windows
- [x] Make unit tests work
  - [x] **TypeCobol.LanguageServer.Test.csproj**: server exits with return code 134 (??) and for `RefactoringTests` line break is always considered 2 chars long which is wrong on Linux
  - [x] **TypeCobol.Test.csproj**: Use of backslash to find test files and wrong line breaks in generated documentation
  - [x] **TypeCobol.Transform.Test.csproj**: all tests are ok but we should ensure return codes are in the correct range (0 to 255)
  - [x] **Codegen.Test.csproj**: One file is not found, we have a problem on whitespace generation and one test has a parsing error
  - [x] **TypeCobol.Analysis.Test.csproj**: in dot format, line breaks are replaced by 2 spaces on Windows but only one on Linux
  - [x] **CLI.Test.csproj**: wrong return codes (overflow), and some other problems...
- [x] #2681
- [x] #2684
- [ ] Create documentation for multi-platform development in the TypeCobol project

PR #2691, opened 2024-12-31
WI #2674 Introduce NodePositions record class to improve readability

fm-117:
This is a supplementary PR over #2690 containing the refactoring of node positions in Codegen.

### DIFF
--- a/Codegen/src/Generators/LinearNodeSourceCodeMapper.cs
+++ b/Codegen/src/Generators/LinearNodeSourceCodeMapper.cs
@@ -223,7 +223,7 @@ namespace TypeCobol.Codegen.Generators
             /// <summary>
             /// Precalculated positions
             /// </summary>
-            public Tuple<int, int, int, List<int>, List<int>> Positions
+            public NodePositions Positions
             {
                 get;
                 set;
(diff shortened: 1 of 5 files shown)


Codegen/src/Generators/LinearNodeSourceCodeMapper.cs
```diff
@@ -560,11 +560,8 @@ bool CanInsertBeforeNode(Node beforeNode)
             //Debug.Assert(Nodes[beforeNode.NodeIndex].Positions != null);
             if (!(Nodes[beforeNode.NodeIndex].Positions != null))
                 return false;
-            //Debug.Assert(Nodes[beforeNode.NodeIndex].Positions.Item4.Count > 0);
-            if (!(Nodes[beforeNode.NodeIndex].Positions.Item4.Count > 0))
-                return false;
-            //Debug.Assert(Nodes[beforeNode.NodeIndex].Positions.Item5.Count > 0);
-            if (!(Nodes[beforeNode.NodeIndex].Positions.Item5.Count > 0))
```
efr15, line 567: Considering the code in method `FromToPositions` (line 216 which was removed) the count for the 2 lists corresponding to Item 4 and Item 5 could be different.
So, in theory, the result of the method `CanInsertBeforeNode ` may now differ.

efr15 (changes requested):

Codegen/src/Generators/LinearNodeSourceCodeMapper.cs
fm-117: The code in `FromToPositions` is tricky but in the end the two collections always have the same number of items because the condition `if (srcFirstLine != null)` is always true. Having at least one line in the `lineNumbers` list means we have reached the first line and we have set the `srcFirstLine`.

Issue #2692, opened 2025-01-03
Add packaging metadata


fm-117:
Following  #2674.

While publishing our parser on nuget.org is not a goal at the moment, we would like to improve description of our projects, adding necessary metadata for proper NuGet packaging.

The goals of this issue are:
- ensure required properties for `dotnet pack` are set
- ensure meaningful names and descriptions are provided for resulting nupkgs

PR #2693, opened 2025-01-03
WI #2692 Add basic metadata for NuGet packaging

fm-117:
Added some basic properties for projects eligible to NuGet publishing.
This is incomplete as included/excluded assets are not defined, However it is enough to create:
- `TypeCobol.Analysis`
- `TypeCobol.Parser`
- `TypeCobol.CSCup.Runtime`

packages which is our current target for private publishing.


fm-117:
Also normalized props and csproj files to tabs as it is the default used by VS.

efr15 (approved):

efr15 (approved):

PR #2708, opened 2025-02-21
WI #2707 Avoid full source dump, keep code for future uses

### LINKED ISSUE #2707: Avoid logging full source code
To help detect and fix anomalies happening only in incremental edits scenarios, we added full source code logging along with change tracking and full code elements listing in log traces.

However these traces did not help so far but are extremely heavy.

The goal of this issue is to make log traces lighter, keeping the capability to activate detailed logging if need be.

fm-117:
Fixes #2707

### DIFF
--- a/TypeCobol/Compiler/CompilationUnit.cs
+++ b/TypeCobol/Compiler/CompilationUnit.cs
@@ -1,6 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿using System.Text;
 using TypeCobol.Analysis;
 using TypeCobol.Compiler.CodeElements;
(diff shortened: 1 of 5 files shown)


fm-117:
Ok maybe this is a bit better with the move of the method to the `LoggingSystemExtensions`. We still have to build a version specifically to activate full source code dump, this is not dynamic.
@efr15 tell me what you think about this PR.

efr15 (approved):

efr15:
An idea is to activate full source code dump by an option.
It will require also to be able to set this option client side.

But I don't know if this use case will happen in real life.
Maybe we should keep the code as is and wait for a real situation.

PR #2732, opened 2025-04-08
WI #2731 Add option for mixed debug check, remove option for END PROGRAM

### LINKED ISSUE #2731: New option to control severity of instructions mixing debug and non-debug
The goal of this issue is to turn the diagnostic created when a CodeElement mixes debug and non-debug lines into an optional diagnostic with adjustable severity.

fm-117:
Fixes #2731

Two changes in this PR:
- WI #2731 Add option for checking debug mix in CodeElement
  - a new option is available to control the severity, and possibly disable, the check of CodeElements mixing debug and non-debug lines
- WI #2731 Remove option to check END PROGRAM
  - remove the `diag.cep` option as the performed checks always resolve to an error on IBM compiler

Note that the second change is breaking so I am thonking of creating a v2.8.0 to release this.

### DIFF
--- a/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
+++ b/TypeCobol/Compiler/Diagnostics/CrossChecker.cs
@@ -912,12 +912,9 @@ namespace TypeCobol.Compiler.Diagnostics
         public override bool Visit(End end)
         {
             // Check if PROGRAM END is orphan
-            if (_compilerOptions.CheckEndProgram.IsActive)
+            if (end.CodeElement.Type == CodeElementType.ProgramEnd && !(end.Parent is Program))
             {
-                if (end.CodeElement.Type == CodeElementType.ProgramEnd && !(end.Parent is Program))
-                {
-                    DiagnosticUtils.AddError(end, "Unexpected orphan \"PROGRAM END\".", _compilerOptions.CheckEndProgram.GetMessageCode());
-                }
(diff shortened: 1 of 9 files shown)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1444,48 +1441,41 @@ private void CheckEndNode([CanBeNull]IToken openingToken, Node endNode)
...
         {
-            var checkEndProgramOption = _compilerOptions.CheckEndProgram;
             var lastChild = node.Children.LastOrDefault();
             if (lastChild is End end)
             {
                 // END PROGRAM is present
                 node.SetFlag(Node.Flag.MissingEndProgram, false);
-                if (checkEndProgramOption.IsActive)
+                var programEnd = (ProgramEnd)end.CodeElement;
+                if (programEnd.ProgramName?.Name == null)
```
efr15, line 1450: We could take this opportunity to introduce the local variable `programEndName`. The value `programEnd.ProgramName?.Name` is used 3 times and it would facilitate reading.

efr15 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
fm-117: Yes, done.

efr15 (approved):

PR #2740, opened 2025-04-29
WI #2739 Align DataLayout flags with Node ones

### LINKED ISSUE #2739: InsertVariableDisplay: align DataLayout flags with Node ones
After #2738 some flags have been created on `Node `class. They differ with the existing ones in `DataLayoutNode`.
It is better to align them, especially to make the difference between `isDisplayable `and `ExceedsStandardIndexCapacity`.

efr15:
Fixes #2739 

Some changes in expected result for `miscPgm.cbl`
Nodes with OCCURS > 65535 now have 2 flags set instead of none
The 2 flags are Displayable and ExceedsStandardIndexCapacity for a total value = 10


TypeCobol.LanguageServer/Processor/DataLayoutProcessor.cs
```diff
@@ -255,6 +255,11 @@ internal static DataLayoutNode From(DataDefinition dataDefinition, DataLayoutNod
                     declarationItems.Add($"OCCURS {dataDefinition.MaxOccurencesCount}");
                 }
 
+                if (dataDefinition.IsFlagSet(Node.Flag.ExceedsStandardIndexCapacity))
```
fm-117, line 258: The flag is computed only for DataRedefines and DataDescriptions so maybe move this after the `if (dataDefinition.IsFlagSet(Node.Flag.Displayable))` line 236.
No need to test for the flag for DataRenames or DataCondition for example

efr15: Done

fm-117 (approved):

Issue #2718, opened 2025-03-17
KeyNotFoundException - CodeElementLocator.FindCodeElementAt


efr15:
**Describe the bug**
A `KeyNotFoundException `is thrown by `TypeCobol.LanguageServer.Utilities.CodeElementLocator.FindCodeElementAt` method.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOZERRCP.
       DATA DIVISION.

       WORKING-STORAGE SECTION.

       01 VAR-1 PIC X.
      * Error: VAR 2 instead of VAR-2
       01 VAR 2 PIC X.

       PROCEDURE DIVISION.
           GOBACK
           .
       END PROGRAM TCOZERRCP.
```

**Expected behavior**
No exception.

**Technical**
The local function `WithCorrespondingNode `reads the dictionary `programClassDocument.NodeCodeElementLinkers` with a key which is not present.

For the given source:

- All code elements have been created 
- Only nodes for declarations preceding the error are created (until VAR-1). No nodes are created for PROCEDURE DIVISION, GOBACK etc.

So it's possible to look for an existing code element that is not present in the dictionary `programClassDocument.NodeCodeElementLinkers` because there is no corresponding node.

Example:
The given key '[[ProcedureDivisionHeader]] [8,16:PROCEDURE]<PROCEDURE> --> [26,26+:.]<PeriodSeparator>
' was not present in the dictionary.

Stack trace:
```
at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
    at TypeCobol.LanguageServer.Utilities.CodeElementLocator.<FindCodeElementAt>g__WithCorrespondingNode|0_1(CodeElement codeElement, <>c__DisplayClass0_0&)
    at TypeCobol.LanguageServer.Utilities.CodeElementLocator.FindCodeElementAt(CompilationUnit compilationUnit, Position position) 
    at TypeCobol.LanguageServer.DataLayoutProcessor.GetProgram(CompilationUnit compilationUnit, Position position)
    at TypeCobol.LanguageServer.DataLayoutProcessor.CollectDataLayoutNodesAtPosition(CompilationUnit compilationUnit, Position position, Action`1 convert)
    at TypeCobol.LanguageServer.DataLayoutProcessor.GetDataLayoutAsCSV(CompilationUnit compilationUnit, Position position, String separator)
[...]

PR #2750, opened 2025-05-22
WI #2718 Prevent KeyNotfoundException in FindCodeElementAt method

fm-117:
We should consider that `FindCodeElementAt` may fail so callers have to check the result and decide whether they have enough info to continue or not.

Currently `FindCodeElementAt` is used by DataLayout and InsertVariableDisplay and both require the Node to be valued.


TypeCobol.LanguageServer.Test/ProcessorTests/DataLayout/CSV-Empty.txt
efr15, line 1: Is it possible to add a test for the TREE mode?

efr15 (dismissed):

TypeCobol.LanguageServer.Test/ProcessorTests/DataLayout/CSV-Empty.txt
fm-117: Added test for empty file

efr15 (approved):

PR #2771, opened 2025-06-13
WI #2717 Implement completion after IN

### LINKED ISSUE #2717: Completion after OF or IN is wrong
**Describe the bug**
Requesting suggestions after `IN` or `OF`  keywords (when trying to disambiguate a variable) does not work properly:
- after `IN`, the parser has no custom logic an thus will suggest all variables, including the one we are trying to disambiguate
- after `OF`, the parser always returns the parent of the first variable having the name found before `OF` which is not helpful as the developer tries to find the correct variable among the list of variables having the same name

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOFM117.
       data division.
       working-storage section.
       01 group1.
          05 var1 PIC X.
       01 group2.
          05 var1 PIC X.
       PROCEDURE DIVISION.
           DISPLAY var1 IN 
           DISPLAY var1 OF 
           goback
           .
       END PROGRAM TCOFM117.
```
and/or steps to reproduce the behavior:
1. Try completion after `IN` at line 10
2. Try completion after `OF` at line 11

**Expected behavior**
Completion should suggest `group1` and `group2`.

**Technical**
- find how to factorize completion for both keywords `IN` and `OF` when they are used to disambiguate a variable
- modify `CompletionFactory.GetCompletionForOfParent` method to extend the search to all eligible parents
  - beware of complex scenarios involving multiple chained `OF` / `IN`

**How to test automatically**
LSR test(s)

efr15:
Fixes #2717

### DIFF
--- a/TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
+++ b/TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
@@ -191,7 +191,7 @@ namespace TypeCobol.LanguageServer
             return results;
         }
 
-        public static CompletionItem CreateCompletionItemForSingleVariable(DataDefinition parent, DataDefinition variable, TypeCobolOptions options, bool qualify)
+        public static CompletionItem CreateCompletionItemForSingleVariable(DataDefinition parent, DataDefinition variable, TypeCobolOptions options, bool qualify, TokenType qualificationTokenType = TokenType.OF)
         {
             string name = variable.Name;
             string type = variable.DataType.Name;
@@ -230,7 +230,8 @@ namespace TypeCobol.LanguageServer
                 if (options.IsCobolLanguage)
                 {
                     nameParts.Reverse();
-                    insertText = string.Join(" OF ", nameParts);
+                    var qualificationToken = TokenUtils.GetTokenStringFromTokenType(qualificationTokenType);
+                    insertText = string.Join($" {qualificationToken} ", nameParts);
                 }
                 else
                 {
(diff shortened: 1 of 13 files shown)


TypeCobol.LanguageServer/Completion Factory/CompletionForInOrOf.cs
```diff
@@ -60,11 +60,13 @@ private List<CompletionItem> GetCompletionForOfParent(Node node, Token variableN
         }
 
         private readonly Position _position;
+        private readonly TokenType _completedTokenType; // IN or OF
```
fm-117, line 63: Rename to `_qualificationTokenType` to match where it is used ? The name `_completedTokenType` is a bit puzzling. I understood that it is the token after which completion is requested but it was not immediate (at least for me)

TypeCobol.LanguageServer/Completion Factory/CompletionFactoryHelpers.cs
```diff
@@ -191,7 +191,7 @@ public static List<CompletionItem> CreateCompletionItemsForVariableSetAndDisambi
             return results;
         }
 
-        public static CompletionItem CreateCompletionItemForSingleVariable(DataDefinition parent, DataDefinition variable, TypeCobolOptions options, bool qualify)
+        public static CompletionItem CreateCompletionItemForSingleVariable(DataDefinition parent, DataDefinition variable, TypeCobolOptions options, bool qualify, TokenType qualificationTokenType = TokenType.OF)
```
fm-117, line 194: 3 params revolving around the qualification style. Maybe it could be merged into a `QualificationStyle` enum param: `DoNotQualify, UseTypeCobolQualifiedNameSeparator, UseOF, UseIN` ? See how it goes with the callers. Keep current code if it creates more problems than it solves

fm-117 (changes requested):

TypeCobol.LanguageServer/Completion Factory/CompletionForInOrOf.cs
efr15: Rename to `_lastSignificantTokenType `(because `_qualificationTokenType` is not adequate when field is used for ADDRESS OF)

PR #2786, opened 2025-07-16
WI #2784 Avoid use of non readonly static fields

fm-117:
All cases have been corrected except:
- `FunctionCallResult.callSiteCounter` in _StorageArea.cs_, line 391
- `SecureObjectPool.s_poolUserIdCounter` in _ImmutableList.cs_, line 5727

Both are used to generate unique ids and are legitimate as their only usages are through `Interlocked.Increment(ref int)` method which prevent concurrency issues when accessing from multiple threads.

About performance test: the result file was outdated, apparently since 995dda44ed0af6115f8af49f4cd34befd3b8e30e. Updating it is not strictly part of this issue.

### DIFF
--- a/TypeCobol/Compiler/AntlrUtils/PerfStatsForParserInvocation.cs
+++ b/TypeCobol/Compiler/AntlrUtils/PerfStatsForParserInvocation.cs
@@ -64,6 +64,10 @@ namespace TypeCobol.Compiler.AntlrUtils
             DecisionTimeMs += perfStats.DecisionTimeMs;
             RuleInvocationsCount += perfStats.RuleInvocationsCount;
             TreeBuildingTime += TreeBuildingTime;
+
+            // Adding detailed stats for ANTLR is not defined. DetailedAntlrProfiling is available only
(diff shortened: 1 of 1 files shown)


efr15:
Could you please explain why you remove the static usage of field `AntlrPerformanceProfiler `in `CodeElementsParserStep`?

fm-117:
> Could you please explain why you remove the static usage of field `AntlrPerformanceProfiler `in `CodeElementsParserStep`?

Initially the ANTLR stats object was store on a single static reference which is incorrect because the stats can be computed each time the ANTLR parser is called. Now it is attached to the `PerfsStatsForParserInvocation` object which is more in line with its role. It allows to compare stats between first time parsing and last time parsing for example (event though we don't use this for now).

As a drawback the `PerfsStatsForParserInvocation` object is slightly bigger, holding one more reference. Also, the combination of ANTLR stats is not implemented so the stats won't be available on `TotalParsingTime`.

efr15 (approved):

PR #2790, opened 2025-07-21
WI #2789 Suggest END-xxx keyword according to user filter text

### LINKED ISSUE #2789: Also suggest END-xxx keywords
When using keyword completion, the language server will return only keywords _starting with_ the user filter text. It may be useful to suggest also corresponding `END-xxx` keywords.

For example typing 'P' or 'p' for a keyword should return `PERFORM` and `END-PERFORM`. 

All ending keywords should be suggested after keywords starting with user filter text. Use proposal groups to separate between keywords and ending keywords.

fm-117:
Fixes #2789


efr15 (approved):
We will see how users react to this.
I prefer the IBM way. When the user types the beginning of a keyword they propose the keyword alone and the keyword + its END (and other variants).
Example:
For 'PE', IBM suggests:
PERFORM
PERFORM END-PERFORM
PERFORM UNTIL EXIT

Issue #2232, opened 2022-05-06
InvalidOperationException in SelectAllTokensBetween method


fm-117:
**Describe the bug**
`InvalidOperationException` has been thrown by `SelectAllTokensBetween`. See following stack trace:

```
   à TypeCobol.Compiler.Scanner.TokensLinesIterator.SelectAllTokensBetween(Token startToken, Token stopToken)
   à TypeCobol.Compiler.Preprocessor.PreprocessorStep.CupProcessTokensLinesChanges(CompilationDocument document, ISearchableReadOnlyList`1 documentLines, IList`1 tokensLinesChanges, PrepareDocumentLineForUpdate prepareDocumentLineForUpdate, IDocumentImporter documentImporter, PerfStatsForParserInvocation perfStatsForParserInvocation, List`1& missingCopies)
   à TypeCobol.Compiler.CompilationDocument.RefreshProcessedTokensDocumentSnapshot()
   à TypeCobol.Compiler.FileCompiler.CompileOnce(Nullable`1 exec2Step, Boolean haltOnMissingCopy)
   à TypeCobol.LanguageServer.Workspace.UpdateSourceFile(Uri fileUri, TextChangedEvent textChangedEvent)
   à TypeCobol.LanguageServer.TypeCobolServer.OnDidChangeTextDocument(DidChangeTextDocumentParams parameters)
   à TypeCobol.LanguageServer.VsCodeProtocol.LanguageServer.CallDidChangeTextDocument(NotificationType notificationType, Object parameters)
```

**To Reproduce**
Private source code available, check _TCOJCM04.cbl_ file.

**Expected behavior**
No exception.

fm-117:
Message is `Invalid start or stop token : line or columns number do not define a valid selection interval`.

fm-117:
Dummy fix: instead of throwing the exception set `stopToken` to `startToken` and `stopLineIndex` to `startLineIndex` ?

PR #2816, opened 2025-11-25
WI #2232 Limit probability of InvalidOperationException during preprocessing step

fm-117:
Ok this is not a fix but rather an attempt to limit uses of `SelectAllTokensBetween`.

As our logs show that the method was invoked on tokens that should not participate in a CompilerDirective building, it means that something very wrong happened... However the directive builder should also not create anything in that case so the method should not be called with this new code.


fm-117:
Replaced by #2821

PR #2833, opened 2026-02-06
WI #2814 Check OCCURS cannot be specified for a data which is redefined

### LINKED ISSUE #2814: OCCURS clause cannot be specified for a data which is redefined
**Describe the bug**
An OCCURS clause cannot be specified for a data which is redefined.
However the parser does not create a diagnostic for such cases.

**To Reproduce**

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOCCRED.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
        01 ROOT-GRP.
      * Error: a redefined data cannot be an OCCURS
           05 VAR-OCC PIC X(10) OCCURS 10.
           05 GRP-REDEF REDEFINES VAR-OCC.
              10 VAR1 PIC X(100).
       PROCEDURE DIVISION.
           GOBACK
           .
       END PROGRAM TCOCCRED.
```

**Expected behavior**
A diagnostic should be created for the line corresponding to the VAR-OCC definition.
Here is the IBM error message:
A definition of an object of a "REDEFINES" clause contained an "OCCURS" clause.  The "REDEFINES" clause was discarded.

Finou1404:


TypeCobol.Test/Parser/Programs/Cobol85/OccursRedefines.Nodes.txt
```diff
@@ -0,0 +1,24 @@
+Line 9[35,41] <27, Error, Syntax> - Syntax error : OCCURS cannot be specified for data VAR-OCC which is redefined.
```
fm-117: I think the message should be phrased the other way. The diagnostic is positioned on the `REDEFINES` so the message should tell that it is wrong to redefine an item having an OCCURS clause instead of telling that it is wrong to have an OCCURS clause on a redefined item.

TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
```diff
@@ -159,6 +159,13 @@ public static void OnNode(DataRedefines redefinesNode)
                     redefinedVariable.IsStronglyTyped ? "strongly-typed" : "strictly-typed");
                 DiagnosticUtils.AddError(redefinesNode, message, redefinesSymbolReference, code:MessageCode.SemanticTCErrorInParser);
             }
+
+            // An OCCURS clause cannot be specified for a data which is redefined.
+            if (redefinedVariable.IsTableOccurence)
+            {
+                DiagnosticUtils.AddError(redefinesNode, $"OCCURS cannot be specified for data {redefinesSymbolReference.Name} which is redefined.", redefinesSymbolReference);
+            }
```
fm-117: We could use this PR to do a bit housekeeping: the checks on a `DataRedefines` should be separated between what is related to standard Cobol and what is related to Cobol2002/TypeCobol.

Could you please:
- move the parts related to resolving the REDEFINES target and your new check to `CrossCompleteChecker.Visit(DataRedefines dataRedefines)`
- keep the rest here but add the `redefinedVariable` as an argument of this method

fm-117 (changes requested):

efr15 (dismissed):

TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
fm-117: The `RedefinesChecker` class is also mentioned here

https://github.com/TypeCobolTeam/TypeCobol/blob/9cce04a1906df1e34ff507d79542ec98ee56e008/TypeCobol/Compiler/Symbols/RedefinesSymbol.cs#L31

So don't forget to update that as well.

TypeCobol.Test/Parser/Programs/Cobol85/OccursRedefines.Nodes.txt
Finou1404: Message changed

TypeCobol/Compiler/Diagnostics/Cobol2002Checker.cs
Finou1404: Refactoring done

fm-117 (approved):

PR #2834, opened 2026-02-16
WI #2520 Check variables used in EVALUATE / WHEN

### LINKED ISSUE #2520: Parser does not check conditions
**Describe the bug**
When using a variable in condition instead of a DataCondition (Level-88), the parser does not emit any diagnostic.

**To Reproduce**
```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TCOMFL06.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 var1 PIC X.
          88 cond1 VALUE 'A'.
          88 cond1 VALUE 'B'.
       PROCEDURE DIVISION.
           IF (var1) THEN
              DISPLAY "This is definitely wrong"
           END-IF
           GOBACK
           .
       END PROGRAM TCOMFL06.
```

**Expected behavior**
Diagnostic expected at line 9 signaling an invalid condition.

**Technical**
New check to implement in CrossChecker. List all nodes that should be checked or find another way to factorize condition checks.

**How to test automatically**
Standard program parsing test.

efr15:
Fixes #2520

### DIFF
--- a/TypeCobol.Test/Parser/Programs/Cobol85/EvaluateWhenCompatibility.PGM.txt
+++ b/TypeCobol.Test/Parser/Programs/Cobol85/EvaluateWhenCompatibility.PGM.txt
@@ -1,0 +1,64 @@
+﻿--- Diagnostics ---
+Line 23[17,33] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 25[17,35] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 27[17,29] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 29[17,35] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 35[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 39[17,29] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 41[17,28] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 58[17,26] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 60[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 64[17,30] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 77[17,33] <27, Error, Syntax> - Syntax error : The literal A is compared to another literal A: this is not valid
+Line 77[17,33] <27, Error, Syntax> - Syntax error : The literal B is compared to another literal A: this is not valid
+Line 81[17,29] <27, Error, Syntax> - Syntax error : The literal 1 is compared to another literal A: this is not valid
+Line 81[17,29] <27, Error, Syntax> - Syntax error : The literal 9 is compared to another literal A: this is not valid
+Line 85[17,26] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 87[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 91[17,30] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 95[17,28] <27, Error, Syntax> - Syntax error : The literal A is compared to another literal A: this is not valid
+Line 104[17,33] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 106[17,35] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 108[17,29] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 110[17,35] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 116[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 120[17,29] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 122[17,28] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 133[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 137[17,29] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 139[17,28] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 148[17,45] <27, Error, Syntax> - Syntax error : The object at position 1 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 155[17,24] <27, Error, Syntax> - Syntax error : The literal B is compared to another literal A: this is not valid
+Line 157[17,22] <27, Error, Syntax> - Syntax error : The literal 9 is compared to another literal A: this is not valid
+Line 159[17,34] <27, Error, Syntax> - Syntax error : The literal Z is compared to another literal A: this is not valid
+Line 166[17,33] <27, Error, Syntax> - Syntax error : The literal B is compared to another literal 1: this is not valid
+Line 168[17,36] <27, Error, Syntax> - Syntax error : The literal 2 is compared to another literal 1: this is not valid
+Line 168[17,36] <27, Error, Syntax> - Syntax error : The literal 9 is compared to another literal 1: this is not valid
+Line 175[17,25] <27, Error, Syntax> - Syntax error : The number of WHEN objects is not equal to the number of EVALUATE subjects
+Line 179[17,46] <27, Error, Syntax> - Syntax error : The number of WHEN objects is not equal to the number of EVALUATE subjects
+Line 188[21,35] <30, Error, Semantics> - Semantic error: Symbol NOT-REFERENCED1 is not referenced OffendingSymbol=[21,35:NOT-REFERENCED1]<UserDefinedWord>
+Line 189[22,36] <30, Error, Semantics> - Semantic error: Symbol NOT-REFERENCED2 is not referenced OffendingSymbol=[22,36:NOT-REFERENCED2]<UserDefinedWord>
+
+--- Program ---
+PROGRAM: EVALWHEN common:False initial:False recursive:False
+ author: ? written: ? compiled: ? installation: ? security: ?
+--- Intrinsic:Namespace:Program:Global:Local
+-- DATA --------
+  MAIN-GROUP:Alphanumeric
+  VAR1:Alphanumeric
+  COND1:Level88
+  COND2:Level88
+  VAR2:Alphanumeric
+  VAR3:Alphanumeric
+  VAR4:?
+  VAR-INDEX1:Numeric
+  VAR-INDEX2:Numeric
+  VAR5:Alphabetic
+  VAR6:Numeric
+  VAR7:Numeric
+--- Intrinsic
+-- TYPES -------
+  BOOL:BOOL
+  DATE:DATE
+  CURRENCY:CURRENCY
+  STRING:STRING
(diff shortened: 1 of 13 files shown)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -435,179 +435,210 @@ public override bool Visit(Search search)
...
+                        //else undefined reference
 
+                        return true;
+                    }
+                }
+            }
+            else if (when.Parent?.Parent is Evaluate evaluate) 
+            {
+                SelectionObject[] selectionObjects = when.CodeElement.SelectionObjects;
+                if (selectionObjects.Count() == 0)
```
fm-117: Avoid using LinQ `Count()` extension on arrays, use `Length` property instead

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -435,179 +435,210 @@ public override bool Visit(Search search)
...
+                {
+                    // Nothing in WHEN = nothing to do
                     return true;
                 }
+
+                SelectionSubject[] selectionSubjects = evaluate.CodeElement.SelectionSubjects;
+                if (selectionSubjects.Count() != selectionObjects.Count())
```
fm-117: Same remark

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -435,179 +435,210 @@ public override bool Visit(Search search)
...
+                {
+                    DiagnosticUtils.AddError(when, "The number of WHEN objects is not equal to the number of EVALUATE subjects");
```
fm-117: "The number of WHEN objects does not match the number of EVALUATE subjects"

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -435,179 +435,210 @@ public override bool Visit(Search search)
...
+                }
+                else
+                {
+                    // Check selection subject and object having the same ordinal position
+                    for (int index = 0; index < selectionSubjects.Count(); index++)
```
fm-117: use `Length`

TypeCobol.Test/Parser/Programs/Cobol85/EvaluateWhenCompatibility.PGM.txt
```diff
@@ -0,0 +1,64 @@
+﻿--- Diagnostics ---
+Line 23[17,33] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
```
fm-117: IBM compiler starts counting at 1 in its error message. So either modify the index reported in the message or better: use the names of the mismatching subjects/objects. that would always be clearer for the dev.

TypeCobol.Test/Parser/Programs/Cobol85/EvaluateWhenCompatibility.PGM.txt
```diff
@@ -0,0 +1,64 @@
...
+Line 35[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 39[17,29] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 41[17,28] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 58[17,26] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 60[17,25] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 64[17,30] <27, Error, Syntax> - Syntax error : The object at position 0 in the "WHEN" phrase does not match the type of the corresponding subject in the "EVALUATE" statement
+Line 77[17,33] <27, Error, Syntax> - Syntax error : The literal A is compared to another literal A: this is not valid
```
fm-117: Please add double quotes around literals for improved readability

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+
+    // Specific Checker for Condition
+    static class ConditionChecker
+    {
+        // Structure gathering information about SelectionSubject (from EVALUATE) or SelectionObject (from WHEN)
+        private record SelectionInfo
+        {
+            // DataType computed from condition or expression
+            // Set to Unknown if condition or expression is not valid
+            public DataType DataType { get; init; }
```
fm-117: `init;` can be removed

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+
+            // For numeric and alphanumeric values only
+            public string LiteralValue { get; init; }
```
fm-117: Same here

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
             }
         }
     }
+
+    // Specific Checker for Condition
+    static class ConditionChecker
+    {
+        // Structure gathering information about SelectionSubject (from EVALUATE) or SelectionObject (from WHEN)
+        private record SelectionInfo
```
fm-117: Optional: can be rewritten as `private record SelectionInfo(DataType DataType, string LiteralValue = null)` and removing the properties and constructors

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+                LiteralValue = literalValue;
+            }
+
+            public SelectionInfo(DataType dataType) : this(dataType, null) { }
+
+            // TRUE for numeric and alphanumeric values only
+            public bool IsLiteral => LiteralValue != null;
+
+            /// <summary>
+            /// Determines whether the given object is conflicting with the current one by comparing their type.
```
fm-117: comparing their types

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+            /// If both types are valid, a conflict occurs when one is Boolean and the other one not.
```
fm-117: the other isn't

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+                        DiagnosticUtils.AddError(when, "The objects in the THRU phrase must be of the same data type");
+
+                        return (false, null, null);
+                    }
+                }
+
+                return (true, selectionObjectInfo, selectionObjectInfo2);
+            }
+
+            SelectionInfo GetSelectionInfo(Node node, BooleanValueOrExpression booleanComparisonVariable, VariableOrExpression alphanumericComparisonVariable)
```
fm-117: I would rather have this code as a private static method in `SelectionInfo` type along with two public `Create` methods, one accepting an `Evaluate` and a `SelectionSubject`, the other a `When` and a `SelectionObject`.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+
+                                    // Not valid (variable not referenced or mixed types)
+                                    return new SelectionInfo(DataType.Unknown);
+                                }
+
+                            default:
+                                // RelationCondition (for instance VAR1 = "A")
+                                return new SelectionInfo(DataType.Boolean);
+                        }
+                    }
```
fm-117: Why bother exploring the expression ? We could consider the DataType as boolean and let the other checkers decide whether the expression is correctly written. I'd like to avoid recursively checking every expression twice.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,294 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+                    if (selectionObjectInfo2 != null && selectionObjectInfo2.IsLiteral)
+                    {
+                        DiagnosticUtils.AddError(when, $"The literal {selectionObjectInfo2.LiteralValue} is compared to another literal {selectionSubjectInfo.LiteralValue}: this is not valid");
+                    }
+                }
+            }
+
+            (bool IsValid, SelectionInfo SelectionObjectInfo, SelectionInfo SelectionObjectInfo2) CheckWhen()
+            {
+                if (!CheckConditionalExpression(booleanComparisonVariable?.Expression, when, true))
```
fm-117: The `IsValid` and the new boolean return type of `CheckConditionalExpression` are used to avoid multiple diagnostics on same node. Is it required ? If it can be removed this would simplify the code.

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
fm-117: Coud you test without the `case LogicalOperation` ? I agrre that we have to keep the first case so as to retrieve the actual DataType of the variable.

efr15: The case LogicalOperation is needed to handle valid syntax like `WHEN NOT VAR-ALPHA`.
In this case we have a `BooleanComparisonVariable `= `LogicalOperation `= {`Operator = NOT, LeftOperand = null, RightOperand = VAR-ALPHA`} and the result DataType should not be boolean.
You can see some examples of such syntax in:
https://github.com/TypeCobolTeam/TypeCobol/blob/9cce04a1906df1e34ff507d79542ec98ee56e008/TypeCobol.Test/ThirdParty/Nist/NC225A.cbl#L449-L450

Finou1404 (dismissed):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol.Test/Parser/Programs/Cobol85/EvaluateWhenCompatibility.PGM.txt
efr15: Done

TypeCobol.Test/Parser/Programs/Cobol85/EvaluateWhenCompatibility.PGM.txt
efr15: Done (IBM style)

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done. `Create `for `When `requires an additional parameter to distinguish  between `from `and `thru `expression of a range

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Code removed

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Code removed

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Good! Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Both are removed. No multiple diagnostics because not valid SelectionInfos are ignored in the type comparison.
There was a mix between null and Unknown for not valid SelectionInfo => now I always use Unknown.

Code is simpler but performance could be affected.
However, I think that in the nominal case, expressions should be simple (a variable or a literal), so this shouldn't be significant.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -605,25 +605,25 @@ usedKey.Parent is DataDefinition parentUsedKey &&
...
                 else
                 {
                     // Check selection subject and object having the same ordinal position
-                    for (int index = 0; index < selectionSubjects.Count(); index++)
+                    for (int index = 0; index < selectionSubjects.Length; index++)
                     {
                         var selectionSubject = selectionSubjects[index];
                         var selectionObject = selectionObjects[index];
-                        ConditionChecker.CheckWhenStatement(evaluate, when, selectionSubject, selectionObject, index);
+                        ConditionChecker.CheckWhenStatement(evaluate, when, selectionSubject, selectionObject, index + 1);
```
fm-117, line 627: Please add a comment explaining the `+ 1` on index.

fm-117 (changes requested):
I think we should simplify the grammar using a single definition for `WHEN` code elements whether they appear in `EVALUATE` or `SEARCH`.
  - this should avoid odd cases where alphanumeric comparison variables/expressions are wrapped as `BooleanComparisonVariable`

We should also improve the checks performed on `ConditionalExpression` as we'll have multiple visits

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1827,4 +1814,298 @@ private static DataDefinition GetDataDefinitionType(DataDefinition symbol, Symbo
...
+                }
+
+                // Variable
+                return GetSelectionInfoFromStorageArea(alphanumericComparisonVariable.StorageArea);
+
+                SelectionInfo GetSelectionInfoFromStorageArea(StorageArea storageArea)
+                {
+                    DataDefinition dataDefinition = node.GetDataDefinitionFromStorageAreaDictionary(storageArea, true);
+
+                    return new SelectionInfo(dataDefinition?.DataType ?? DataType.Unknown);
```
fm-117: `var dataType = dataDefinition?.DataType`;
return dataType != null ? new SelectionInfo(dataType) : Unknown;

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

fm-117 (dismissed):

fm-117 (approved):

PR #2835, opened 2026-02-25
WI #2520 Check USAGE compatibility between EVALUATE and WHEN

### LINKED ISSUE #2520: Parser does not check conditions

efr15:
Fixes #2520

There is an issue with USAGE OBJECT REFERENCE.  The diagnostic below is wrong:
`Line 23[11,30] <27, Error, Syntax> - Syntax error : A group item cannot be empty. Add children, picture or valid usage (INDEX, COMP-1, COMP-2, POINTER, POINTER-32, PROCEDURE-POINTER or FUNCTION-POINTER). OffendingSymbol=[11,30:VAR-OBJECT-REFERENCE]
`
OBJECT REFERENCE should be part of the "stand alone" usages (see new issue created #2836)


TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1821,10 +1821,11 @@ static class ConditionChecker
     {
         // Structure gathering information about SelectionSubject (from EVALUATE) or SelectionObject (from WHEN):
         // DataType is computed from condition or expression and set to Unknown if not valid
+        // Usage is retrieved from the data definition (if relevant) or set to None
```
fm-117, line 1824: set to None or `null` ? The signature uses a nullable type but apparently it can be removed.

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1840,6 +1841,49 @@ public bool HasConflictingType(SelectionInfo other)
...
+            /// <returns>true if conflict</returns>
+            public bool HasConflictingUsage(SelectionInfo other)
+            {
+                var usage = Usage ?? DataUsage.None;
+                var otherUsage = other?.Usage ?? DataUsage.None;
+
+                // None and UTF-8 (which is not managed, see #2504) are ignored
+                if (otherUsage is DataUsage.None or DataUsage.UTF8) return false;
+
+                return usage switch
```
fm-117: Maybe testing equality of both usages before the switch will simplify a bit, removing the identity cases from the switch ?

fm-117 (changes requested):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: nullable removed

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
```diff
@@ -1821,10 +1821,11 @@ static class ConditionChecker
     {
         // Structure gathering information about SelectionSubject (from EVALUATE) or SelectionObject (from WHEN):
         // DataType is computed from condition or expression and set to Unknown if not valid
+        // Usage is retrieved from the data definition (if relevant) or set to None
         // LiteralValue is only relevant for numeric and alphanumeric values
-        private record SelectionInfo(DataType DataType, string LiteralValue = null)
+        private record SelectionInfo(DataType DataType, DataUsage Usage, string LiteralValue = null)
```
fm-117: Optional: as most of the usages have the value `None` for Usage property, you can add a default value. Use the named parameter syntax for the cases where you have to pass a value for the literal without an Usage.

fm-117 (approved):

TypeCobol/Compiler/Diagnostics/CrossChecker.cs
efr15: Done

fm-117 (approved):

fm-117 (approved):

Finou1404 (approved):

PR #2841, opened 2026-03-11
Fix parser hang on unsupported SQL statements when EnableSqlParsing is true

allieseb:
## Summary

- Fix infinite hang in `CompileOnce()` when `EnableSqlParsing = true` and the COBOL source contains SQL DML statements not supported by the ANTLR grammar (INSERT, UPDATE, DELETE, DECLARE CURSOR, OPEN, FETCH, CLOSE) or SELECT with WHERE/INTO/JOIN clauses.
- The Scanner now checks the first keyword of each `EXEC SQL` block against the list of statements with complete grammar rules before enabling SQL token parsing. Unsupported statements fall back to `ExecStatementText` mode.
- `IsSqlStatementStartingKeyword()` is extended with DML tokens so ANTLR error recovery stops at statement boundaries.

## Problem

A COBOL program with standard DB2 SQL (`synth_cobol_sql.cbl`, 204 lines) caused `CompileOnce()` to never return. The file contains:

| SQL Statement | Grammar Support | Behavior Before Fix |
|---|---|---|
| `CONNECT TO` / `CONNECT RESET` | Full rule | OK |
| `COMMIT` | Full rule | OK |
| `SELECT col1, col2 INTO :var FROM table WHERE ...` | Partial (only `SELECT table.*`) | **Hangs** |
| `INSERT INTO table VALUES (...)` | No rule | **Hangs** |
| `UPDATE table SET col = val WHERE ...` | No rule | **Hangs** |
| `DELETE FROM table WHERE ...` | No rule | **Hangs** |
| `DECLARE CURSOR FOR SELECT ...` | No rule | **Hangs** |
| `OPEN` / `FETCH` / `CLOSE` cursor | No rule | **Hangs** |

When the ANTLR parser encounters SQL tokens with no matching grammar rule, error recovery tries all ~100 `codeElement` alternatives for each token, causing exponential backtracking.

### `Scanner.cs`
- Added `IsSupportedSqlStatement(startIndex, endIndex)` method that peeks at the first SQL keyword after `EXEC SQL`
- Modified `ScanSqlCodeOrExecStatementTextOrExecSqlInclude()`: only sets `InsideSql = true` if the SQL keyword is in the supported list (COMMIT, ROLLBACK, CONNECT, TRUNCATE, SAVEPOINT, DROP, LOCK, RELEASE, WHENEVER, SET, GET, ALTER, EXECUTE)
- Unsupported SQL falls through to `ExecStatementText` — same behavior as `EnableSqlParsing = false` for those blocks

### `SqlStatementElement.cs`
- Added `SQL_INSERT`, `SQL_UPDATE`, `SQL_DELETE`, `SQL_DECLARE`, `SQL_OPEN`, `SQL_FETCH`, `SQL_CLOSE` to `IsSqlStatementStartingKeyword()` so error recovery stops consuming at DML statement boundaries

## Test plan

- [x] Compilation of `synth_cobol_sql.cbl` (204 lines, 14 EXEC SQL blocks with all DML types): **30s+ timeout → < 1 second**
- [x] Supported SQL statements (COMMIT, CONNECT, ROLLBACK, etc.) still parsed as SQL tokens
- [x] No regression on existing COBOL test files without SQL

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/TypeCobol/Compiler/Scanner/Scanner.cs
+++ b/TypeCobol/Compiler/Scanner/Scanner.cs
@@ -2001,8 +2001,15 @@ namespace TypeCobol.Compiler.Scanner
 
             if (tokensLine.ScanState.AfterExecSql)
             {
-                // Expect SQL code
-                tokensLine.ScanState.InsideSql = true;
+                // Only enable SQL token parsing for statements that have grammar rules.
+                // Unsupported statements (INSERT, UPDATE, DELETE, DECLARE CURSOR, OPEN,
+                // FETCH, CLOSE, SELECT with INTO/WHERE/JOIN) fall back to ExecStatementText
+                // to avoid ANTLR parser hangs during error recovery.
+                if (!compilerOptions.EnableSqlParsing || IsSupportedSqlStatement(startIndex, endIndex))
+                {
+                    tokensLine.ScanState.InsideSql = true;
+                }
+                // else: unsupported SQL → InsideSql stays false → ExecStatementText fallback
             }
 
             if (tokensLine.ScanState.InsideSql && compilerOptions.EnableSqlParsing)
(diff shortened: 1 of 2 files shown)


fm-117:
Replaced by #2842

PR #2848, opened 2026-04-21
WI #2847 Protect scanner against empty tokens

### LINKED ISSUE #2847: Protect against empty tokens
In some cases (reported in SQL scanner but to be confirmed in Cobol scanner), the scanner may be stuck in an infinte loop creating empty tokens and not advancing its index.

While this should not happen, it is safer to protect against empty tokens.

Example code reproducing the problem:

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. UnsupportedBySqlScanner.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 WS-DATE-FIN        PIC X(10).
       PROCEDURE DIVISION.
           EXEC SQL
               SELECT * FROM PROJECTS
      * KO: operator '<' is not supported by SQL scanner
               WHERE DATE_FIN < :WS-DATE-FIN
           END-EXEC
       END PROGRAM UnsupportedBySqlScanner.
```

fm-117:
Closes #2847

### DIFF
--- a/TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/UnsupportedByScanner.rdz.cbl
+++ b/TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/UnsupportedByScanner.rdz.cbl
@@ -6,7 +6,7 @@
        PROCEDURE DIVISION.
            EXEC SQL
                SELECT * FROM PROJECTS
-      * KO: operator '<' is not supported by SQL scanner
-               WHERE DATE_FIN < :WS-DATE-FIN
+      * KO: operator '$' is not supported by SQL scanner
+               WHERE DATE_FIN $ :WS-DATE-FIN
            END-EXEC
        END PROGRAM UnsupportedBySqlScanner.
--- a/TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/UnsupportedByScanner.rdz.cbl
+++ b/TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/UnsupportedByScanner.rdz.cbl
@@ -1,0 +1,12 @@
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. UnsupportedBySqlScanner.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 WS-DATE-FIN        PIC X(10).
+       PROCEDURE DIVISION.
+           EXEC SQL
+               SELECT * FROM PROJECTS
+      * KO: operator '<' is not supported by SQL scanner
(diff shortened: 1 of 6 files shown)


TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/UnsupportedByScanner.rdz.cbl
```diff
@@ -0,0 +1,12 @@
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. UnsupportedBySqlScanner.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 WS-DATE-FIN        PIC X(10).
+       PROCEDURE DIVISION.
+           EXEC SQL
+               SELECT * FROM PROJECTS
+      * KO: operator '<' is not supported by SQL scanner
```
efr15: Maybe we could change `<` to `$` so the test will remain valid even if `<` is implemented in SQL scanner.

TypeCobol/Compiler/Scanner/AbstractScanner.cs
```diff
@@ -42,7 +42,20 @@ protected AbstractScanner(string line, int startIndex, int lastIndex, TokensLine
...
+        public Token? GetNextToken()
+        {
+            var token = GetTokenStartingFrom(currentIndex);
+
+            // Safety check: always avoid creating empty tokens otherwise the scanner will run on an infinite loop.
+            // The index MUST progress, so force scan an invalid token of length 1 instead of an empty one.
+            if (token?.Length == 0)
+            {
+                token = ScanOneChar(currentIndex, TokenType.InvalidToken);
+                tokensLine.AddDiagnostic(MessageCode.SyntaxErrorInParser, token, $"Found unexpected char '{token.Text}', this syntax is not supported by this parser.");
```
efr15: If we report this error on the client side, it would be interesting to have the code line as well to be able to investigate and reproduce the problem.

efr15 (changes requested):

TypeCobol.Test/Parser/Programs/Cobol85/ExecSql/UnsupportedByScanner.rdz.cbl
fm-117: Done

TypeCobol/Compiler/Scanner/AbstractScanner.cs
fm-117: Extended error message + dedicated diagnostic code

efr15 (approved):

PR #2861, opened 2026-07-21
WI #2849 Add MaxOccurs in DataLayout request (CSV mode only)

### LINKED ISSUE #2849: Add MaxOccurs in DataLayout request (CSV mode only)
In DataLayout request with CSV mode the infomation MaxOccurs is currently not available straight forward.
It must be retrieved in the declaration (example : DATA OCCURS 10).
It could be helpful to have it directly in each line of the CSV result.
The MaxOccurs should be omitted in the TREE mode.

**Technical**
Enhancement to be done in DataLayoutProcessor.

**How to test automatically**
Adapt existing test results.

Use Processor tests mechanism: add a new test specific to CSV only.
Its COBOL code should contain:

- an OCCURS with UNBOUNDED clause (to illustrate that, in this case, the parser considers MaxOccurs to be equal to 1)
- some nested OCCURS

efr15:
Fixes #2849


TypeCobol.LanguageServer/Processor/DataLayoutProcessor.cs
```diff
@@ -278,6 +279,7 @@ internal static DataLayoutNode From(DataDefinition dataDefinition, DataLayoutNod
                 bool isNamed = !string.IsNullOrEmpty(dataDefinition.Name);
                 var name = isNamed ? dataDefinition.Name : FILLER;
                 int occursDimension = parent.OccursDimension + (incrementDimension ? 1 : 0);
+                long maxOccurs = dataDefinition.IsTableOccurence ? dataDefinition.MaxOccurencesCount : 0;
```
fm-117, line 283: Could you add a comment explaining that the default value is different between the DataLayout result (which uses 0 for non-OCCURS data)  and DataDefinition (which uses 1) ?

fm-117 (changes requested):

TypeCobol.LanguageServer/Processor/DataLayoutProcessor.cs
efr15: Done

fm-117 (approved):

